PR #5, opened 2026-08-18
feat(replay): randomize synthetic request lengths

dreamtalen:
## Summary

- add `random_range_ratio` and `random_seed` to standalone synthetic replay through the CLI, Sweeper workload schema, and engine runner
- independently sample ISL and OSL from `[int(ratio * configured_length), configured_length]`, materializing the complete ISL vector before OSL
- preserve fixed-length behavior by default with `random_range_ratio=1.0`
- reject invalid ratios, non-`u64` seeds, trace workloads, and randomized multi-turn workloads

InferenceX uses `--random-range-ratio 0.8` for the predictor-comparison workloads. Replaying those requests at fixed lengths phase-locks closed-loop completions and creates synchronized prefill batches, which inflates simulated TTFT. In the original [Dynamo experiment](https://github.com/ai-dynamo/dynamo/pull/13284), randomized lengths reduced MiniMax-M2.5 1k/1k TTFT MAPE from 277.53% to 29.07%, and point-weighted TTFT MAPE across 28 operating points from 126.48% to 24.05%.

This follow-up targets the post-refactor AISimulate boundary directly: request lengths are materialized in the standalone Python runner before crossing the JSON/PyO3 execution seam. It adds no Dynamo or KVBM dependency.

## Validation

- `ruff format --check src tests`
- `ruff check src tests`
- `git diff --check origin/main...HEAD`
- `PYTHONPATH=src /Volumes/workplace/dynamo/dynamo/bin/python -m pytest -q tests/test_runner.py tests/test_replay_cli.py tests/sweeper/test_config.py` — 107 passed


src/aisimulate/sweeper/config.py
```diff
@@ -334,6 +340,29 @@ def _validate_workload(self) -> Workload:
             v = getattr(self, name)
             if v is not None and v <= 0:
                 raise ValueError(f"{name} must be positive, got {v}")
+        if (
```
jasonqinzhou, line 359: [P2] Validate the derived lower bounds

The schema accepts ratios in `(0, 1]` without checking whether `int(isl * ratio)` or `int(osl * ratio)` becomes zero. For example, `isl=1`, `osl=1`, and `random_range_ratio=0.8` passes `Workload` validation but fails later in `_sample_synthetic_lengths`, so a Sweeper config can survive setup and then fail every evaluation. Please validate both derived bounds here, or clamp them to one if that matches the intended contract.

src/aisimulate/sweeper/config.py
```diff
@@ -190,6 +190,8 @@ class Workload(BaseModel):
     num_request_ratio: float | None = (
         None  # request count multiplier for concrete concurrency or request_rate
     )
+    random_range_ratio: float = 1.0
+    random_seed: int = 0
```
jasonqinzhou, line 194: [P2] Prevent Pydantic type coercion

These non-strict fields normalize booleans, integral floats, and numeric strings before the `mode="after"` validator runs. As a result, `random_seed=True`, `1.0`, and `"7"` are accepted as integers, while `random_range_ratio=True` or `"0.8"` is also accepted. The existing `isinstance(self.random_seed, bool)` check therefore cannot enforce the documented `u64` contract. Please use strict field types or `mode="before"` validators for these inputs.

jasonqinzhou (commented):
Two validation gaps to address before merge.

jasonqinzhou (approved):
Approved at exact head `dd50d6eda46e` after addressing both validation threads. Local validation: 107 focused Python tests, 290 broader source Python tests, 455 Rust workspace tests, Ruff, formatting, and diff checks.

PR #11, opened 2026-08-18
feat(generator): lower Sweeper candidates into deployment requests

jasonqinzhou:
## Summary

- lower a public AISimulate Sweeper Candidate plus its workload into the existing typed GeneratorRequest
- preserve the evaluated backend version, topology, worker counts, engine and KV limits, model facts, concurrency, and supported Dynamo Router, Planner, and KVBM settings
- fail closed on GPU-count drift or unmapped adapters, document the ownership boundary, and cover actual deployment artifact rendering

## Scope

This is the first implementation slice of [AIC-1647](https://linear.app/nvidia/issue/AIC-1647/sweeper-build-the-unified-aisimulate-configuration-generator). It establishes the deterministic Sweeper-to-generator bridge and reuses the existing generator. CLI result-envelope/export commands and top-N/Pareto UX remain follow-up work while the existing AIC CLI stays supported.

## Testing

- full generator suite: 295 passed, 4 skipped
- cross-package suite: 206 passed
- focused bridge and engine-limit suite after final changes: 18 passed
- Ruff 0.14.1 format and check: passed


tedzhouhk (changes requested):
we need to discuss which config generator to use for DGD generation.

tedzhouhk (approved):

jasonqinzhou:
> we need to discuss which config generator to use for DGD generation.

Discussed and we go with legacy config generator first. Will have better one later to collaborate with DGDRv2.

PR #12, opened 2026-08-19
feat(cli): add strict aggregate SLA filtering [AIC-1646]

jasonqinzhou:
## Summary
- build on the unified CLI from merged PR #29 and expose strict SLA through the public `aisimulate recommend` configuration
- add `optimization.strict_sla` and lower it to the Sweeper `goal.strict_sla` gate
- reuse the existing `ttft_ms`, `itl_ms`, and `e2e_ms` bounds
- filter violations before sampler observation, scalar ranking, or Pareto dominance
- add an AIConfigurator-to-AISimulate migration guide with executable CLI-to-CLI examples
- consolidate the public CLI design and migration guide under `docs/cli/`

## Semantics
Default replay goodput remains request-level: slow requests contribute no tokens to goodput without rejecting the candidate. With `optimization.strict_sla: true`, configured bounds additionally gate aggregate candidate means. Bounds are inclusive, and missing or non-finite metrics fail closed.

Strict filtering may use TTFT or ITL independently for non-goodput targets. Goodput targets still require a complete per-request SLA.

## Stack
- base: `main` (#29 merged)
- follow-up: #40 adds `request_latency_ms`

## Validation
- `ruff check` on changed public CLI, Sweeper, and test files
- `pytest -q tests/sweeper tests/test_cli_config.py tests/test_unified_cli.py tests/test_unified_traffic_runtime.py tests/test_stack_discovery.py` (295 passed, 3 skipped)

Linear: AIC-1646


python/aisimulate/src/aisimulate/config/cli.py
```diff
@@ -66,10 +69,18 @@ def _validate_cross_component(self) -> CoreRecommendationConfig:
             and "disaggregated" in modes
         ):
             raise ValueError(f"{source.format} requires aggregated engine mode")
-        if self.optimization.target in {"goodput", "goodput_per_gpu"}:
-            sla = self.evaluation.sla
-            if sla is None or (sla.e2e_ms is None and (sla.ttft_ms is None or sla.itl_ms is None)):
-                raise ValueError(f"optimization target {self.optimization.target!r} requires a complete evaluation.sla")
+        sla = self.evaluation.sla
+        if self.optimization.strict_sla:
```
tedzhouhk: Could we make request-level Replay use the same optional-bound semantics here? An unset TTFT or ITL should mean that metric is unbounded (`None`, conceptually +infinity), while every configured bound is enforced. Today this accepts ITL-only/TTFT-only for recommendation, but the partial `goal.sla` is copied into `ReplaySpec` and native `SlaThresholds::validate` still requires the pair; generated prediction validation also rejects the same partial SLA, so the documented form produces no recommendations. Please update the native Replay SLA definition/validation and the prediction schema together, and add engine/Dynamo plus generated-YAML round-trip tests for TTFT-only and ITL-only inputs.

python/aisimulate/src/aisimulate/sweeper/score.py
```diff
@@ -99,14 +105,53 @@ def score_report(report: dict[str, float], target: OptimizationTarget) -> float:
...
+        ("ttft", "mean_ttft_ms", sla.ttft_ms),
+        ("tpot", "mean_tpot_ms", sla.itl_ms),
+        ("e2e", "mean_e2e_latency_ms", sla.e2e_ms),
+    )
+    violations: list[str] = []
+    for label, metric, bound in checks:
+        if bound is None:
+            continue
+        try:
+            value = float(report[metric])
```
tedzhouhk, line 179: Please fail closed when the report has no qualifying latency samples. Native Replay serializes an empty latency distribution as finite `0.0` means, so a candidate with `completed_requests: 0` satisfies every positive strict SLA here. For minimized TTFT/E2E it can even rank ahead of candidates that served requests. At minimum, reject zero completions/sample counts and retain that count in `Candidate.metrics` so the final `analyze_candidates()` pass applies the same rule.

python/aisimulate/src/aisimulate/sweeper/config.py
```diff
@@ -70,12 +75,15 @@ class SLATarget(BaseModel):
     @model_validator(mode="after")
```
tedzhouhk, line 75: This changes the direct Sweeper SDK to accept a partial token SLA even when `strict_sla` is omitted/false, while native Replay still rejects it. If the intended contract is that each request-level SLA field is independently optional, please make that change in Replay too and add a non-strict SDK/replay regression test. Otherwise this validator needs to preserve the previous TTFT+ITL pairing outside strict mode. The SDK, unified CLI, and native runner should not accept different SLA shapes.

docs/cli/migrate-from-aiconfigurator.md
```diff
@@ -0,0 +1,102 @@
...
+  --osl 128 \
+  --ttft 800 \
+  --tpot 30 \
+  --strict-sla
+```
+
+Save this AISimulate configuration as `recommendation.yaml`:
+
+```yaml
+traffic:
```
tedzhouhk, line 56: This is not behaviorally equivalent to the legacy command above: `aiconfigurator cli default` has no offered-load input and compares aggregated/disaggregated configurations under the eight-GPU capacity, while this YAML injects fixed 4 RPS, forces aggregated mode, and optimizes throughput. Could we either use a legacy invocation with equivalent load/mode/objective semantics, or label this translation as illustrative rather than a matching migration?

tedzhouhk (commented):
Agent review: four inline findings on the strict-SLA contract and migration semantics.

jasonqinzhou:
@tedzhouhk Thanks for the review — I fixed all four findings in fa524d3:

- aligned independent TTFT/ITL bounds across prediction, Sweeper, native Replay, and the engine/Dynamo contracts
- restored the TTFT+ITL pairing requirement for the non-strict direct Sweeper SDK
- added native latency sample counts and fail-closed handling for zero completions/samples in strict filtering and latency ranking, retaining the counts in Candidate.metrics
- relabeled the migration example as illustrative and documented the load/mode differences

Coverage now includes generated recommendation YAML -> prediction and engine-boundary tests for TTFT-only and ITL-only, plus an optional ai-dynamo integration contract for both fields.

Local validation: 170 focused Python tests passed; 5 optional ai-dynamo packaging checks skipped because ai-dynamo is not installed locally; 26 Rust report tests passed; Ruff, cargo fmt, and diff checks passed.

I left the review threads open for your verification.

python/aisimulate/src/aisimulate/config/cli.py
tedzhouhk: Thanks — the strict partial-bound path, native Replay validation, and prediction round-trip are fixed. One semantic mismatch remains relative to the intended request-level contract: native Replay and `predict` now accept TTFT-only/ITL-only, but non-strict `recommend`, direct Sweeper, and goodput targets still require the TTFT+ITL pair (or E2E). Since an unset request-level field should simply be unbounded, could we allow any configured SLA bound in those paths too? `strict_sla` should only control the additional aggregate hard filter, not the request-level SLA shape. We can defer the full cross-repo engine/Dynamo E2E coverage from this PR.

tedzhouhk (approved):

dreamtalen (approved):

PR #15, opened 2026-08-19
feat(sweeper): define canonical result semantics [AIC-1471]

jasonqinzhou:
## Summary

- Add schema-versioned `SweepResult` v1.0 as the shared exhaustive and optimizer-guided result envelope.
- Preserve the complete candidate-attempt ledger with feasible, infeasible, unsupported, timed-out, and failed statuses plus stable reason categories and invariant-checked counts.
- Expose scalar top-N and Pareto views by stable candidate ID while supporting `all`, `feasible`, and `views` retention policies.
- Carry run, model, hardware, backend/version, topology, workload, objective/SLA, performance-data, power, per-operation, and runner provenance.
- Provide strict lossless JSON round-trip and a documented one-row-per-candidate CSV analysis view.

## Unified CLI integration

- Integrate the current unified `aisimulate` CLI from `main` and remove the superseded `python -m aisimulate.sweeper` entry point.
- Make `run_recommendation()` return the canonical `SweepResult` while preserving the existing selected-candidate behavior.
- Write `recommendation.json` beside generated recommendation YAML files, including a durable empty result when no feasible candidate exists.
- Preserve each candidate's `prediction_config` through the result ledger, JSON round-trip, CSV export, and conversion back to the SDK `Candidate` type.
- Keep overwrite cleanup and human-readable CLI output aligned with the new result artifact.

## Search and provenance behavior

- Preserve the unified optimizer's global trial budget, algorithm/seed behavior, round-robin branches, timeout handling, prediction-config factories, parallel-projection failures, and typed adapter rejections.
- Record parallel projection and adapter constraints as stable structured reason categories.
- Infer AIConfigurator performance-data provenance only when an active role actually uses the AIC timing source; fixed or polynomial timing no longer claims an AIC performance database.

## Review follow-up

- Rebuild CLI selected views after concrete adapter canonicalization and deduplication, preserving a one-to-one candidate-ID to YAML artifact mapping.
- Preserve normalized replay metrics on every completed replay, including later-infeasible rows.
- Count same-batch coalesced suggestions explicitly as cache hits, source candidate provenance from concrete `ReplaySpec` and deployment metadata, and classify zero-sample latency as typed `no_samples` infeasibility.
- Update the authoritative CLI output tree, overwrite, empty-result, partial-failure, and durable-ledger contract.

## Compatibility and consumers

- Make `Sweeper.run()` the sole canonical SDK contract returning `SweepResult`.
- Require callers that need the scalar top-N or Pareto selection to use `result.selected_candidates`; there is no competing list-returning execution API.
- Document the field-by-field mapping from legacy `ColumnsAgg`, `ColumnsDisagg`, EPD, AFD, `best_config_topn.csv`, `pareto.csv`, and per-operation source artifacts.
- Document scalar, Pareto, empty, mixed-mode, and failure examples plus approved migration exceptions.

## Scope boundaries

Deployment artifact generation remains owned by the generator component. This PR does not touch the generator bridge files in PR #11.

The repository currently has an optimizer-guided executor, so this PR wires that path and the public unified CLI to `SweepResult`. It also exposes `SearchStrategy.EXHAUSTIVE` for the standalone exhaustive producer tracked by AIC-1697. Generator and QA adoption remain with their owning parity issues.

## Base / validation

- Built as signed-off commits directly on current `main`; PR #12 and old main merge history are not replayed.
- Review follow-up focused result, search, and unified CLI regression suite: 42 passed.
- Review follow-up full Sweeper and unified CLI suite: 279 passed with 5 expected cross-repository integration skips.
- Changed-file `ruff check`, `ruff format --check`, and `git diff --check`: passed.
- Exact `main...head`: 19 intended files, zero commits behind; obsolete `docs/cli-design.md` was not resurrected.
- Current repository CI runs the full matrix only on canonical pushes; DCO and codeowners are the checks attached to this fork head. Branch policy still requires a code-owner approval.

## Tracking

- AIC-1471: https://linear.app/nvidia/issue/AIC-1471/sweeper-define-legacy-and-refactored-sweeper-result-semantics
- Parent parity program AIC-1481: https://linear.app/nvidia/issue/AIC-1481/sweeper-close-legacy-sweeper-parity-gaps


python/aisimulate/src/aisimulate/sweeper/search.py
```diff
@@ -776,14 +934,43 @@ def run(
         *,
         on_round: Callable[[int, list[Candidate]], None] | None = None,
     ) -> list[Candidate]:
-        """Run the sweep and return feasible candidates sorted best-first.
+        """Compatibility view returning the full ranked/frontier candidate list.
```
tedzhouhk: We do not need to preserve the old list-returning Sweeper API. Please remove this compatibility adapter and make the canonical `SweepResult` path the single public entry point (ideally `run()` itself). Callers that only need selected candidates can explicitly use `result.selected_candidates`. Keeping both `run()` and `run_result()` creates two competing result semantics and preserves a compatibility commitment the project has not made.

python/aisimulate/src/aisimulate/sweeper/search.py
```diff
@@ -1222,4 +1507,40 @@ def _run_branch_round(state: _BranchSearchState) -> None:
                 remaining = len(failure_reasons) - len(displayed)
                 suffix = f" | +{remaining} more distinct reason(s)" if remaining else ""
                 tqdm.write(f"Sweeper failure reason(s): {' | '.join(displayed)}{suffix}")
-        return result
+        selected_ids = [record_id_by_candidate_object[id(candidate)] for candidate in selected_candidates]
```
tedzhouhk, line 1541: [P1] These selected views are built from internal candidates before the CLI canonicalizes adapter config and deduplicates concrete prediction YAML. Two internal candidates with the same `prediction_config` can therefore appear as two selected IDs in `recommendation.json` while the CLI emits one YAML and one stdout row. Keep the full ledger, but build the public selected view after concrete-config canonicalization/dedup, or add an explicit candidate-ID to artifact mapping so the canonical result and emitted recommendations cannot disagree.

python/aisimulate/src/aisimulate/sweeper/search.py
```diff
@@ -1006,11 +1198,67 @@ def _eval_batch(todo: list[tuple[Suggestion, _PreparedCandidate]]):
...
+                        prediction_config=deepcopy(
+                            candidate.prediction_config if candidate is not None else prediction_config
+                        ),
+                        used_gpus=(
+                            candidate.used_gpus
+                            if candidate is not None
+                            else (int(snapshot["used_gpus"]) if snapshot.get("used_gpus") is not None else None)
+                        ),
+                        score=candidate.score if candidate is not None else None,
+                        metrics=(deepcopy(candidate.metrics) if candidate is not None else {}),
```
tedzhouhk: [P1] This drops all replay metrics whenever a completed replay is later classified infeasible. For example, strict-SLA rejection retains `report_metrics`, but the ledger row gets `metrics={}`; the values are used only to extract power into provenance. A lossless ledger must retain the normalized report metrics for every candidate that completed replay, regardless of feasibility, with `status` carrying the rejection decision. Otherwise users cannot verify or reanalyze the rejection.

python/aisimulate/src/aisimulate/sweeper/search.py
```diff
@@ -1114,32 +1377,37 @@ def _run_branch_round(state: _BranchSearchState) -> None:
...
-                            _record(outcome, None)
+                            _record(
+                                outcome,
+                                None,
+                                candidate_config=_suggestion_snapshot(suggestion, config),
+                                reason=reason,
+                                reason_category=build_result.reason_category,
+                                runner_metadata=build_result.runner_metadata,
+                            )
                             for _duplicate in duplicates:
```
tedzhouhk: [P1] Same-batch duplicate suggestions that share a failed/infeasible result consume `max_trials` and each receive `observe_infeasible`, but `retain_record=False` removes every duplicate from both the ledger and counts without recording a cache/coalesced hit. Two identical failures in one batch produce one failed row, while the same trials in separate batches produce two rows, so accounting depends on batching/parallelism. Either record every optimizer attempt or define and count every coalesced duplicate explicitly.

python/aisimulate/src/aisimulate/sweeper/result.py
```diff
@@ -0,0 +1,508 @@
...
+    return CandidateProvenance(
+        model=config.search_space.model_name,
+        hardware=config.search_space.hardware_sku,
+        backend=(str(candidate_config["backend"]) if candidate_config.get("backend") is not None else None),
+        backend_version=(
+            str(candidate_config["backend_version"]) if candidate_config.get("backend_version") is not None else None
+        ),
+        performance_data=performance_data,
+        topology=topology_fields,
+        workload=config.workload.model_dump(mode="json"),
```
tedzhouhk: [P2] Candidate provenance must come from the concrete `ReplaySpec`, not the run-level search domain plus heuristics. With load choices `[8, 16]`, a candidate executed at concurrency 16 currently records the workload default/domain value 8 here. The same function also guesses performance-data identity instead of consuming `BackendDeploymentSpec.performance_model_metadata`, losing role/model/NextN details. Pass the materialized workload, effective concurrency, and canonical deployment/perf-model metadata into provenance; keep the input domain only in run provenance.

python/aisimulate/src/aisimulate/sweeper/search.py
```diff
@@ -722,7 +861,26 @@ def _score_prepared(
         observe_metrics = {"objective": candidate.score}  # single metric, pre-signed higher-is-better
     if prepared.prediction_config is not None:
         candidate = candidate.model_copy(update={"prediction_config": deepcopy(prepared.prediction_config)})
-    return candidate, observe_metrics, "feasible", ""
+    non_finite_objectives = [name for name, value in (candidate.objectives or {}).items() if not math.isfinite(value)]
```
tedzhouhk, line 902: [P2] This is an independent sweep-semantic change, not result serialization. Current main keeps a missing/zero-sample latency candidate with score `-inf` and calls `sampler.observe`, so it ranks last. This PR removes it from the feasible set and calls `observe_infeasible`, changing Bayesian feedback and subsequent suggestions. The change may be desirable, but it should be explicitly justified and tested as a scoring decision (preferably a typed zero-sample infeasible category), or split from this output-contract PR.

docs/sweeper/results.md
```diff
@@ -2,54 +2,229 @@
...
+> Within a schema version, fields keep their meaning and units; incompatible changes require a new
+> `schema_version` and an explicit converter.
 
-Sweeper materializes every optimizer suggestion in the main process. It unrolls the backend
-selection, asks each configured provider for its concrete adapter configuration and runtime hooks,
-then constructs a `ReplaySpec`.
+`SweepResult` is the canonical result envelope for both exhaustive and optimizer-guided execution.
+The complete candidate ledger is the source of truth. Scalar top-N and Pareto fronts are views that
+refer to ledger rows by stable `candidate_id`; they do not discard the other evaluated candidates.
+The unified `aisimulate recommend` command writes this envelope as `recommendation.json` beside
```
tedzhouhk, line 16: [P2] Please update the authoritative `docs/cli/design.md` output contract in this PR. It still explicitly says run metadata, trial ledgers, status/error records, recommendation indexes, and partial-failure payloads are deferred, and its output tree omits `recommendation.json`. Since this PR intentionally implements that deferred contract, the CLI design, overwrite behavior, empty-result semantics, and output-directory tree must be updated together.

tedzhouhk (commented):
Agentic review: the canonical-result direction is reasonable, and normal finite-candidate sweep behavior is mostly preserved. The inline findings below need to be addressed so the ledger and selected views are actually authoritative and lossless.

jasonqinzhou:
@tedzhouhk Thank you for the detailed review. I have pushed follow-up fixes through `435cb33` addressing all seven findings:

- `Sweeper.run()` is now the sole public execution contract returning `SweepResult`.
- CLI selected views are rebuilt after concrete adapter canonicalization and deduplication, preserving a one-to-one candidate-ID-to-YAML mapping.
- Completed-replay metrics are retained for infeasible rows, and same-batch coalesced suggestions are counted explicitly as cache hits.
- Candidate provenance now comes from the concrete `ReplaySpec` and deployment performance-model metadata.
- Missing/zero-sample latency is explicitly modeled and tested as typed `no_samples` infeasibility.
- The authoritative CLI output, overwrite, empty-result, and partial-failure contract is updated.

Validation: 279 passed with 5 expected cross-repository integration skips; the focused review-regression suite passed 42 tests; changed-file Ruff, format, and diff checks passed. Could you please re-review when you have a chance?

tedzhouhk (commented):
Follow-up after checking 435cb33: most review findings are fixed. One sweep-semantic issue remains: zero or missing latency samples now call sampler.observe_infeasible instead of the previous sampler.observe with a worst-ranked score. Typing this as no_samples and testing it makes the result contract clearer, but it still changes Bayesian feedback and can change later suggestions. Since this PR is intended to change result and output semantics only, please preserve the previous sampler feedback path while recording the candidate as typed no_samples in SweepResult, or split the search-behavior change into a separate PR.

Cross-repo follow-up: making Sweeper.run return SweepResult is the desired API, but Dynamo currently has one integration-test caller that still treats the return value as a list. That caller needs to use result.selected_candidates before Dynamo adopts the AISimulate release.

tedzhouhk (approved):

PR #16, opened 2026-08-19
feat(sweeper): add constrained minimum-GPU optimization [AIC-1776]

jasonqinzhou:
## Summary

- add optimizer-native `min_gpus` selection over directly evaluated complete candidates
- require an explicit `min_goodput_rps` hard constraint and an SLA before minimum-GPU selection
- reject missing, non-finite, or below-threshold replay goodput request-rate data before ranking
- minimize each feasible candidate's actual `used_gpus`; do not extrapolate replicas or assume linear scaling
- record rejected candidates through the canonical `SweepResult` outcome and reason contract
- document the legacy AIConfigurator request-rate migration and the remaining concurrency limitation

## Review feedback addressed

The original post-processing API computed `ceil(target / capacity) * candidate.used_gpus`. That API,
its exports, tests, and documentation have been removed. Each candidate is now replayed at one fixed
synthetic offered request rate. Replay must report at least the configured SLA-good request rate,
and only then does the optimizer rank the complete candidate by its actual GPU count.

The public schema currently limits `min_gpus` to a concrete synthetic request-rate workload. It
requires `evaluation.sla` and `optimization.constraints.min_goodput_rps`, and rejects a goodput
requirement above the offered request rate. `min_gpus` is deliberately unavailable as a Pareto
objective because its value comes from candidate topology rather than a replay metric.

Candidates that miss the goodput constraint are recorded as infeasible with the
`sla_constraint` reason category. Missing or non-finite goodput metrics are recorded as runner
contract failures. This keeps constrained minimum-GPU evaluation inside the canonical result ledger.

## Migration

`docs/cli/migrate-from-aiconfigurator.md` contains a runnable translation of legacy
`aiconfigurator cli recommend --target-request-rate ...`. It explains why the offered rate should
be above the required goodput rate for a finite replay window. The legacy `--target-concurrency`
case is explicitly marked as having no exact AISimulate equivalent yet rather than claiming false
parity.

## Base / dependencies

- Refreshed against `main` at `d63cc0e64afc3e921c71bf6edcd318fcdbab41f5`.
- No stacked PR history remains; the current `main` result contract is used directly.
- No post-processing sizing API or linear fleet-size extrapolation is introduced.

## Validation

- full repository test suite: 410 passed, 5 expected cross-repository integration skips
- focused result/config/scoring/search suite: 93 passed
- complete `tests/sweeper` suite: 276 passed, 5 expected cross-repository integration skips
- complete unified-CLI engine E2E matrix: 18 passed
- CLI config and unified-CLI suite: 45 passed
- constrained engine E2E fixture verifies the one-GPU feasible recommendation and minimum-goodput evidence
- changed-Python `ruff check` and `ruff format --check`: passed
- `git diff --check`: passed

## Linear

- AIC-1776


tedzhouhk (changes requested):
Requesting changes because “minimum GPU” is not yet defined precisely enough to establish a public sizing API.

A valid minimum-GPU objective must be framed as a constrained optimization problem:

```text
minimize total GPU count
subject to:
  a fixed workload/traffic shape,
  served load or goodput >= the target,
  the configured SLA,
  and a feasible complete topology.
```

The implementation here instead reads a candidate raw `request_throughput_rps`, assumes the entire evaluated deployment can be replicated with perfectly linear capacity, and computes `ceil(target / capacity) * candidate.used_gpus`. It does not re-simulate the scaled deployment, does not require target goodput or a served-load fraction, and does not prove that Router, P/D balance, queueing, KV capacity, Planner behavior, or network effects remain valid after replication. Calling that result the minimum GPU requirement is therefore stronger than the data supports; it is a linear fleet-size extrapolation.

The normal efficiency-selection use cases are already covered by the existing optimization targets:

- `throughput_per_gpu` for raw GPU efficiency.
- `goodput_per_gpu` for SLA-compliant GPU efficiency.

If we later need an absolute capacity-planning objective, it should be designed inside the optimizer with explicit hard constraints such as `min_goodput_rps` or `min_served_load_fraction`, evaluated against the target traffic and SLA. The optimizer can then minimize `used_gpus` among complete candidates that actually satisfy those constraints.

I suggest closing this PR rather than publishing the current post-processing API. We can open a separate design for constrained minimum-GPU optimization if the existing per-GPU objectives plus capacity constraints are insufficient.

tedzhouhk:
Thanks for reworking this. The new implementation addresses the modeling concern in my previous review: it evaluates complete candidates directly, applies an explicit SLA-good request-rate constraint, and minimizes actual `used_gpus` without assuming linear replication.

I still think there is a more fundamental product/API question to resolve: **what concrete user or workflow needs `min_gpus + min_goodput_rps`?**

The normal recommendation use cases appear to already be covered:

- `goodput`: maximize SLA-compliant throughput within the configured GPU budget.
- `goodput_per_gpu`: maximize SLA-compliant GPU efficiency.

The new objective is useful only for a distinct inverse capacity-planning request: “I have an externally defined requirement of at least X SLA-good requests/sec; find the smallest evaluated topology that satisfies it.” Before publishing two new public schema concepts, please identify the actual caller or user workflow that supplies this absolute `min_goodput_rps`, and explain why the existing objectives plus candidate reporting/selection are insufficient.

Without such a requirement, this risks adding a mathematically valid optimization problem that users do not actually have. In that case I would prefer not to add `min_gpus` and `min_goodput_rps` to the public API. Making “minimum GPU” well-defined is necessary, but it does not by itself establish that this should be a supported product objective.

PR #24, opened 2026-08-20
feat(replay): add typed agentic graph execution

PeaBrane:
## Summary

Add the typed Agentic Mooncake v2 graph, validated dense execution plan, dependency-aware `WorkloadDriver`, closed-loop lane lifecycle, exact trajectory accounting, and aggregated/offline-P/D execution semantics to AISimulate.

This is the AISimulate half of the cross-repository agentic replay refactor originally developed in ai-dynamo/dynamo#13501. Companion: ai-dynamo/dynamo#13586, which owns Weka and request-trace ingestion, CLI/PyO3 adapters, and Dynamo documentation. Related design issue: ai-dynamo/dynamo#13477.

### Why

Completion-only `wait_for` dependencies could not faithfully represent long-running agentic workflows: an overlapping subagent should begin from its parent's dispatch, a causal continuation should wait for successful completion, and a failed play must stop new descendants while allowing emitted work to settle before recycling its lane.

This change provides one typed graph and lifecycle shared by:

- aggregated offline replay;
- aggregated online mock replay;
- disaggregated offline P/D replay.

The hot path compiles string IDs into dense node/play indices, visits each dependency edge once, and keeps trajectory summaries independent of detailed per-request capture.

### Release ordering

The AISimulate PR can merge without breaking Dynamo because Dynamo is pinned to the already-published dev1 artifacts. AISimulate main is not consumed automatically.

But the Dynamo feature PR cannot merge until:

1. AISimulate changes merge.
2. New artifacts are published, e.g. Rust `0.1.0-dev.2` and Python `0.1.0.dev2` (the release tooling may choose the repository's current `0.12.0` line instead).
3. The Dynamo PR bumps both pins and adds its adapters in the same commit/PR.

Without that bump, Dynamo compiles against dev1, where the new graph and `WorkloadDriver` APIs do not exist. A temporary local path override can support development, but not merge or CI.

## ComputeLab evidence

The combined AISimulate + Dynamo candidate was validated on SC-01 CPU nodes against a frozen public-Weka semantic oracle.

- Oracle: 9,757 requests, 256 plays, 399 dispatch-triggered overlap anchors, and 213 post-completion anchors.
- Executable single-model cohort: 3,971/3,971 requests and 118/118 complete trajectories in every lifecycle row.
- Direct Weka and materialized Agentic Mooncake v2 produced byte-identical request ledgers and semantic reports.
- Zero fatal/error/panic matches.

| Engine semantics | Topology | Candidate/baseline median | Upper 95% bound |
|---|---:|---:|---:|
| vLLM | aggregated | 1.031060 | 1.034936 |
| vLLM | offline P/D, source-first | 1.046210 | 1.047729 |
| SGLang | aggregated | 1.033697 | 1.038065 |
| SGLang | offline P/D, destination-first | 1.025898 | 1.028347 |

Each performance row used five alternating warmups and a persisted balanced 60-pair schedule. Every upper 95% bound remained below the predeclared 1.05 gate; lag/order correlations remained within 0.30.

Resource checks: peak RSS stayed at 36,864 KiB; replay binary size grew 3.39% and `.text` 4.09%. Preparation time was measured separately: 103.741 ms for the legacy fixture, 383.185 ms for materialized v2, and 3,039.982 ms for direct Weka ingestion in the companion Dynamo layer.

## Intentional semantic changes

- Dispatch-triggered children can overlap their parent.
- Completion edges fire only after successful logical completion.
- Failure skips undispatched descendants and delays lane reuse until emitted work is quiescent.
- P/D completion means final decode completion, not prefill or handoff completion.
- Typed Agentic Mooncake v2 is a clean schema transition; legacy/headerless files are rejected.

These differences were treated as semantic corrections, not forced back into parity with the old completion-only scheduler.

## Unsupported modes

- online P/D replay;
- planner/scaling agentic replay;
- `replay_concurrency` with agentic lanes;
- standalone AISimulate Weka ingestion (owned by the companion Dynamo adapter);
- Hugging Face loading or any Dynamo/AIPerf runtime integration.

## Validation

- Published PR head: `ca98807b0c43f7e15a41463cfe0949c81a7c4270`.
- Companion Dynamo adapter head validated through a temporary path override: `8d623a5a3e6e00eaf06d789178e8305ef6ff450e`.
- `cargo test -p aisimulate-core` (426 unit, 5 artifact, 20 engine-integration, 1 facade, 2 firewall, 9 runtime-contract tests)
- `cargo test --workspace`
- `cargo clippy -p aisimulate-core --all-targets -- -D warnings`
- Ruff check and format-check for the changed Python replay CLI/config files
- DCO signed commit

The 14 extracted Rust files are byte-identical to the tested implementation from ai-dynamo/dynamo#13501; the two Python files preserve subsequent AISimulate-main CLI additions before formatting.


PeaBrane:
/ok to test [ca98807](https://github.com/ai-dynamo/aisimulate/pull/24/commits/ca98807b0c43f7e15a41463cfe0949c81a7c4270)

tedzhouhk (approved):

python/aisimulate/src/aisimulate/replay/cli.py
```diff
@@ -17,6 +17,7 @@
     "mooncake",
     "mooncake-delta",
     "agentic_mooncake",
+    "weka",
```
jasonqinzhou, line 20: Could we keep `weka` out of the standalone CLI choices, or reject it explicitly in `main()` before constructing the engine runner? `python -m aisimulate.replay` routes through `EngineReplayRunnerFactory`, whose traffic materializer rejects every non-`mooncake` format. As written, argparse accepts `--trace-format weka` and `--agentic-lanes`, but that command necessarily fails later. Since standalone Weka ingestion is explicitly unsupported, please make that boundary visible at parsing and add a CLI regression test. Dynamo can still use the broader shared parser through a separate choice set.

jasonqinzhou (approved):
Reviewed at the exact head. The typed graph and lifecycle implementation is well structured, and the focused validation is strong. Approving with one non-blocking standalone CLI boundary suggestion inline.

python/aisimulate/src/aisimulate/replay/cli.py
PeaBrane: Fixed in e9f1e96. The standalone entrypoint now rejects `--trace-format=weka` before constructing `EngineReplayRunnerFactory`, while the shared parser remains broad enough for Dynamo. Added a CLI regression proving the runner is not entered.

PR #29, opened 2026-08-21
feat(cli): unify prediction and recommendation inputs

tedzhouhk:
## Summary

- add the single `aisimulate predict` / `aisimulate recommend` CLI with strict YAML schemas, overrides, stack discovery, and config-adapter ABI
- keep AISimulate config engine-only while optional top-level sections remain adapter-owned
- route synthetic, session, and every documented trace format through the shared traffic/runtime path
- preserve native Dynamo agent launch, blocking/background, parent sequence, last-child join, and missing-duration timing semantics
- implement exact default/custom/off preset semantics, including mixed disaggregated role presets and feasibility filtering
- preserve linear/log range scale without eagerly expanding large integer log domains
- add conditional adapter search spaces so inactive knobs are not sampled, while preserving existing Sweeper Python APIs
- enforce the global optimizer trial budget with stable branch round-robin scheduling and branch-identity seeds
- publish the consolidated `aisimulate` wheel and `aisimulate-core` crate together as the planned 0.12.0 transition release

## Validation

- root Python suite with Dynamo entry-point metadata: 349 passed
- Rust workspace: 935 passed, 1 ignored; subsidiary crate suites: 5, 20, 1, 2, and 9 passed
- native Dynamo agentic regressions cover blocking/background launch, parent sequence, last-finishing-child join, timestamp fallback, and absent `total_time_ms`
- a real native Dynamo agentic tool trace completed 4 requests and 1 trajectory through `aisimulate predict`
- pinned backend-version and strict preset-leaf regressions passed
- real engine-only mixed preset recommendation and generated YAML round-trips passed
- release wheel built and installed as 0.12.0; packaging and artifact contracts passed
- Ruff, Cargo fmt, DCO, and `git diff --check` passed

The follow-up Dynamo integration PR is release-gated so Dynamo can pin the published 0.12.0 artifacts.


scripts/build_release_artifacts.py
```diff
@@ -16,13 +16,13 @@
...
 ROOT = Path(__file__).resolve().parents[1]
-VERSION = "0.12.0"
+VERSION = "0.13.0"
 
 EXPECTED_PYTHON_PROJECTS = {
     ROOT / "python" / "aisimulate" / "pyproject.toml": "aisimulate",
     ROOT / "python" / "aisimulate-core" / "pyproject.toml": "aisimulate-core",
 }
-EXPECTED_CRATE = ROOT / "crates" / "aisimulate-core" / "Cargo.toml"
+EXPECTED_CRATE = ROOT / "crates" / "core" / "Cargo.toml"
```
jasonqinzhou, line 24: **[P1] Align the published crate with the artifact contract** This changes the release artifact to crates/core, while docs/artifact-contract.md and docs/core-api.md still define the public aisimulate-core crate as the native estimator from crates/aisimulate-core. That estimator is now publish = false, and its README still instructs users to depend on 0.13.0 with the embed-python feature, which the Replay crate does not expose. Please choose or rename the crate that owns the public package name, then align the manifests, docs, and an API smoke test; as written, 0.13.0 publishes a different API from the one it advertises.

crates/core/src/replay/loadgen/dynamo.rs
```diff
@@ -0,0 +1,456 @@
...
+            continue;
+        };
+        if let Some(parent_indices) = by_session.get(parent_session)
+            && let Some(parent) = parent_indices
+                .iter()
+                .copied()
+                .filter(|index| entries[*index].start_ms <= entries[first].start_ms)
+                .max_by_key(|index| entries[*index].start_ms)
+        {
+            previous[first] = Some(parent);
```
jasonqinzhou: **[P1] Preserve agent launch and join causality** This representation permits only one predecessor and adds only a dependency from the first child request to the parent spawn request. parse_record also drops tool terminal records, so consumer_request_id and execution_mode never reach lowering. Blocking and background traces can require a parent consumer to wait on the last-finishing child while retaining the parent sequence edge. The canonical Dynamo lowering implements those tool-aware launch/join semantics: https://github.com/ai-dynamo/dynamo/blob/1a732cfb8fd83ea857e20a50a17df2babf74e261/lib/data-gen/src/request_trace/agentic.rs#L98-L248. Please port that behavior and its launch/join/background regressions before claiming native Dynamo trace compatibility.

crates/core/src/replay/loadgen/dynamo.rs
```diff
@@ -0,0 +1,456 @@
...
+            );
+            ensure!(
+                request_ids.insert(request.request_id.clone()),
+                "duplicate request_id {:?}",
+                request.request_id
+            );
+            let start_ms = request
+                .request_received_ms
+                .unwrap_or(record.event_time_unix_ms) as i64;
+            let duration_ms = request.total_time_ms.unwrap_or(0.0);
```
jasonqinzhou: **[P1] Use event time when total_time_ms is absent** The request-trace schema allows total_time_ms to be absent. Dynamo canonical request_times uses event_time_unix_ms as the end in that case: https://github.com/ai-dynamo/dynamo/blob/1a732cfb8fd83ea857e20a50a17df2babf74e261/lib/data-gen/src/request_trace/load.rs#L453-L471. Defaulting duration to zero makes end equal start. For agentic dependencies this inflates the recorded gap by the previous request runtime, then replay waits for simulated completion and adds that runtime again. Please derive the fallback duration from event_time_unix_ms minus request_received_ms and add the missing-duration regression.

python/aisimulate/src/aisimulate/recommend.py
```diff
@@ -0,0 +1,670 @@
...
+    context = engine.get("context_length", "max")
+    if context != "max" and (
+        not isinstance(context, int) or isinstance(context, bool) or context <= 0
+    ):
+        raise ValueError("engine.context_length must be 'max' or a positive integer")
+
+    workers = engine.get("workers")
+    if not isinstance(workers, dict):
+        raise ValueError("engine.workers is required")
+    search_space: dict[str, Any] = {
```
jasonqinzhou, line 100: **[P1] Carry a fixed backend_version into every candidate** engine.backend_version is accepted and documented as fixed when set, but this lowering omits it from SearchSpace. Candidate preparation then unconditionally resolves the latest database version and recommendation output emits that replacement. A request pinned to 0.18.0 can therefore be evaluated and emitted with a different backend version. Please add a pinned version field and honor it during resolution and materialization.

python/aisimulate/src/aisimulate/recommend.py
```diff
@@ -0,0 +1,670 @@
...
+    if not isinstance(value, dict):
+        raise ValueError(f"{path} entries must be mappings")
+    missing = set(_PARALLEL_KEYS) - set(value)
+    unknown = set(value) - set(_PARALLEL_KEYS)
+    if missing or unknown:
+        raise ValueError(
+            f"{path} entry must cover exactly {_PARALLEL_KEYS}; "
+            f"missing={sorted(missing)}, unknown={sorted(unknown)}"
+        )
+    return {key: int(value[key]) for key in _PARALLEL_KEYS}
```
jasonqinzhou: **[P2] Reject invalid preset leaves rather than coercing them** int() silently changes values such as 1.9 to 1, true to 1, and string 2 to integer 2, so the evaluated candidate differs from the authored config despite the strict-schema contract. Please require a non-boolean positive integer for every preset leaf and reject anything else.

Cargo.toml
```diff
@@ -6,17 +6,18 @@ members = [
...
+# is published to crates.io; the estimator ships inside the Python wheel.
 exclude = [
     "crates/aisimulate-core",
     "crates/tests/public-api",
 ]
 resolver = "3"
 
 [workspace.package]
-version = "0.12.0"
+version = "0.13.0"
```
jasonqinzhou: **[P1] Keep this release on 0.12.0** The active release project explicitly defines this migration as AIC 0.12.0 / Dynamo 1.5.0 and requires exact three-artifact RC qualification at 0.12.0: https://linear.app/nvidia/project/dynamo-150-and-aic-0120-unified-aisimulate-repository-8dbcf44081ff. AIC has published 0.11.0, while AISimulate 0.12.0 has not shipped, so advancing these artifacts to 0.13.0 skips the planned transition release. Please keep the workspace, both wheels, the public crate, matching dependencies, CI assertions, docs, tests, and lockfiles at 0.12.0 unless there is an approved release-plan decision moving this PR to the post-0.12 cycle.

scripts/build_release_artifacts.py
tedzhouhk: Resolved by the single-crate consolidation in fcc7ad4 (#35). `crates/core` is now the only publishable `aisimulate-core` crate, the old estimator manifest is gone, `embed-python` is exposed by the published crate, both artifact/core API docs point to `crates/core`, and the external public-API smoke crate exercises that contract. The consolidated release is now consistently versioned 0.12.0 in 573c0f1.

crates/core/src/replay/loadgen/dynamo.rs
tedzhouhk: Fixed in 1c37d32. Request timing now uses `event_time_unix_ms` as the end when `total_time_ms` is absent, matching Dynamo `request_times`, while retaining start-plus-duration when duration is present. The missing-duration agentic regression verifies a 10 ms dependency gap instead of double-counting the predecessor runtime.

jasonqinzhou (approved):

python/aisimulate/pyproject.toml
```diff
@@ -98,7 +98,8 @@ build-backend = "maturin"
 [project.scripts]
 # AISimulate owns and ships the application, but preserves the established
 # AIConfigurator command identity. The distribution name is not a second
-# top-level application command.
+# top-level application command for AIC workflows.
+aisimulate = "aisimulate.main:main"
```
jasonqinzhou, line 102: **[P1] Keep installed FPM provenance compatible with the new command** Adding this console entry point makes pip add `../../../bin/aisimulate` (and the corresponding Windows launcher paths) to the installed distribution `RECORD`. `collector.fpm_forward.planner._is_installer_console_script` currently exempts only the legacy `aiconfigurator` launchers, so the installed-wheel FPM smoke test deterministically fails with `installed 'aisimulate' RECORD has unsafe path '../../../bin/aisimulate'`. Please extend the narrow launcher allowlist for both command names and add regression coverage; do not broadly permit arbitrary parent-traversal paths.

jasonqinzhou: Fixed in 7839cd9. The provenance filter now narrowly accepts the `aisimulate` POSIX and Windows launchers alongside `aiconfigurator`; regression coverage exercises all six known launcher paths and still rejects `../../../bin/untrusted`. Validation passed: 17 focused provenance tests, all 100 FPM planner tests, the exact installed-wheel FPM E2E, the 346-test root suite, and the Rust workspace suite on current `main`.

jasonqinzhou (approved):
Re-audited exact head 1a7b0e4 after integrating current main and fixing the installed FPM provenance regression. All review threads are resolved; all nine checks pass, including application-tests and the release-artifact contract.

PR #39, opened 2026-08-25
fix(fpm): admit nearby prefill KV sites

simone-chen:
## Summary

- add an optional raw-distance fallback to the scattered-site log2 admission gate
- enable it only for the FPM prefill KV site axis with an absolute limit of 32 tokens
- preserve the normal log2 distance gate on batch and leave decode/GEMM behavior unchanged
- document the rule and cover both admission and batch-gate preservation

## Motivation

A prefill query at batch 1 and KV 1 could not use collected KV sites 0 and 16. Their absolute gaps are small, but the log2 metric makes the 1-to-16 distance equal to 4, beyond the existing limit of 2.

## Validation

- cargo fmt --check
- cargo test -p aisimulate-core --lib: 928 passed, 1 ignored
- MiniMax M2.7 H200 FPM replay: coverage improved from 22/27 to 26/27; TP4 and TEP4 at concurrency 128 and 256 now succeed
- the remaining DEP4 concurrency-256 failure is an unrelated decode-domain miss and is unchanged


crates/core/src/perfmodel/perf_database/perf_interp.rs
```diff
@@ -910,6 +918,7 @@ impl SiteIndex {
             .iter()
             .map(|&p| coords[p].max(1e-12).log2())
             .collect();
+        let q_site: Vec<f64> = site_axes.iter().map(|&p| coords[p]).collect();
```
jasonqinzhou: [P2] This now allocates a `Vec<f64>` on every unknown scattered-site resolution, including GEMM where `site_axis_abs_distance_fallback` is `None`. The surrounding code identifies this resolver as an engine-step hot path. Please read the fallback coordinate directly from `coords[site_axes[axis]]` and construct `q_site` only inside the frontier-waiver branch, so the FPM exception adds no allocation to disabled callers.

jasonqinzhou: Fixed in 4ed6c6333747a6d8f53c9cd94224bfd7ad459280. The fallback now reads `coords[site_axes[axis]]` directly, while `q_site` is constructed only in the frontier-waiver branch. Focused validation passed: 37 shared-resolver tests, both new FPM admission tests, `cargo fmt --check`, and `git diff --check`.

jasonqinzhou (approved):
Approved at exact head 4ed6c6333747a6d8f53c9cd94224bfd7ad459280. The FPM raw-distance admission fallback is narrowly scoped, the disabled-path allocation finding is fixed, the review thread is resolved, focused local validation passed, and all nine exact-head checks are green.

PR #40, opened 2026-08-25
docs(cli): map AIC request latency to E2E SLA [AIC-1646]

jasonqinzhou:
## Summary
- document the legacy `--request-latency` migration in `docs/cli/migrate-from-aiconfigurator.md`
- map it to the existing `evaluation.sla.e2e_ms` contract plus `optimization.strict_sla: true`
- explain the fixed-output synthetic equivalence between mean E2E latency and mean TTFT plus TPOT
- record that aggregate-only E2E filtering has no public AISimulate mapping today

## Why
For fixed-output synthetic traffic:

```text
mean_e2e_latency_ms = mean_ttft_ms + mean_tpot_ms * (output_tokens - 1)
```

The existing `e2e_ms` field therefore expresses the legacy numerical bound. Reusing it avoids a duplicate request-latency alias, preserves the shared prediction/recommendation SLA schema, and lets `strict_sla: true` apply the same threshold as an aggregate candidate gate.

## Scope
- documentation only: one migration-guide file differs from `main`
- no new schema field or scoring/search behavior
- aggregate-only constraints remain out of scope

## Validation
- parsed both YAML examples in the migration guide successfully
- `git diff --check origin/main`: passed
- exact `main...head`: two signed-off commits, one intended documentation file, zero commits behind

Linear: AIC-1646


tedzhouhk (changes requested):
Suggestion: close this PR and use the existing `evaluation.sla.e2e_ms` contract instead.

Under the fixed-output-length synthetic workload required by this PR,

```text
mean_ttft_ms + mean_tpot_ms * (osl - 1) == mean_e2e_latency_ms
```

so `request_latency_ms` does not introduce a new numerical constraint. The existing configuration already covers the normal workflow:

```yaml
evaluation:
  sla:
    e2e_ms: 12000
optimization:
  strict_sla: true
```

This applies E2E at request level for goodput and applies the same bound to the aggregate mean as a hard candidate filter. Adding `request_latency_ms` creates a second, overlapping name for E2E latency with conflicting semantics, and because it is recommendation-only it also conflicts with the unified requirement that generated recommendation YAML can be passed directly to `predict`.

The only distinct behavior here is an aggregate-only E2E bound that does not participate in request-level goodput. Unless that use case is explicitly required, it does not justify another public SLA field. If we do need it later, it should be modeled explicitly as an aggregate constraint such as `optimization.constraints.mean_e2e_latency_ms`, rather than as a second SLA alias.

For legacy migration, I suggest mapping `--request-latency` to `evaluation.sla.e2e_ms` plus `optimization.strict_sla: true`.

tedzhouhk (approved):
The duplicate request-latency SLA field and implementation have been removed. Keeping this as a documentation-only migration mapping to the existing  plus  is clear and avoids overlapping public contracts.

PR #43, opened 2026-08-25
fix(cli): stabilize unified engine workflows

tedzhouhk:
## Summary

- keep KV-capacity estimation independent from fixed/polynomial timing
- carry optional performance-model identity in runner-neutral deployment metadata without enabling AIC runtime timing
- deduplicate final concrete recommendation configs while preserving optimizer trial accounting and rank order
- add CPU-only predict/recommend E2E coverage for traffic formats, aggregated/disaggregated execution, presets, ranges, and recommendation round trips

## Validation

- `maturin develop --uv`
- `python -m pytest -q -c pytest.ini tests` (369 passed)
- engine unified CLI E2E matrix (16 passed)
- Ruff checks on all changed Python files

Dynamo consumes the additive performance-model metadata in ai-dynamo/dynamo#13665.


tests/e2e/test_unified_cli_engine.py
```diff
@@ -0,0 +1,125 @@
...
+pytestmark = [
+    pytest.mark.integration,
+    pytest.mark.pre_merge,
+    pytest.mark.gpu_0,
+]
+
+_REPO_ROOT = Path(__file__).resolve().parents[2]
+_CONFIG_ROOT = Path("tests/e2e/configs/unified_cli")
+_PREDICT_CASES = tuple(sorted((_REPO_ROOT / _CONFIG_ROOT / "predict/engine").glob("*.yaml")))
+_RECOMMEND_CASES = tuple(sorted((_REPO_ROOT / _CONFIG_ROOT / "recommend/engine").glob("*.yaml")))
```
jasonqinzhou, line 48: Could we pin the expected matrix cardinality here (currently 10 predict and 6 recommend cases), or list the cases explicitly? With unrestricted globs, accidentally deleting or renaming a YAML silently reduces collection, and an empty parameter set is skipped, so CI can remain green while this E2E coverage disappears.

jasonqinzhou (approved):

PR #47, opened 2026-08-26
feat(replay): add engine replay parity skill

dreamtalen:
## Summary

- add an AI Simulate engine-only replay parity skill scoped to round-robin semantic regression detection
- add a pinned Mooncake 5,000-row provenance recipe and qualified vLLM, SGLang, and TRT-LLM golden points
- add an isolated Rust qualification runner, cross-checkout builder, frozen configs, assertions, and negative/determinism tests

## Validation

- independent subagent review: no PR-blocking findings
- skill quick validation
- runner rustfmt and release Clippy with warnings denied
- 6 runner integration tests
- 20 aisimulate-core engine tests and 5 artifact tests
- two fresh processes per golden row produced byte-identical canonical output


python/aisimulate/.agents/skills/aisimulate-replay-parity/scripts/replay-parity-runner/src/main.rs
```diff
@@ -0,0 +1,484 @@
...
+                ReplayTerminalStatus::Failed => "failed",
+            })
+            .or_default() += 1;
+    }
+
+    let pressure = report.runtime_evidence.pressure.as_ref();
+    let kv_ingest = report.runtime_evidence.kv_ingest.as_ref();
+    json!({
+        "row_name": config.row_name,
+        "source_revision": config.source_revision,
```
jasonqinzhou: [P1] Bind source_revision to the compiled checkout

The runner treats source_revision from editable config JSON as authoritative provenance, while build_runner.py only prints git rev-parse HEAD in a separate build record and never binds that SHA into the binary. A candidate-built runner can therefore accept a baseline source_revision and emit a successful report mislabeled as the baseline; I reproduced that at head 38d5c42 with the committed fb7b0d5 config. Please embed the checkout revision or require a verifiable build manifest, and reject any config/revision mismatch.

python/aisimulate/.agents/skills/aisimulate-replay-parity/scripts/build_runner.py
```diff
@@ -0,0 +1,143 @@
...
+        "checkout_dirty": checkout_dirty,
+        "checkout_revision": revision,
+        "manifest": str(manifest),
+        "runner_source_sha256": _runner_source_sha256(),
+    }
+    if not args.prepare_only:
+        subprocess.run(
+            [
+                args.cargo,
+                f"+{args.toolchain}",
```
jasonqinzhou: [P2] Support Cargo installations without rustup shorthand

This command always inserts +<toolchain> after the cargo executable. That syntax is provided by the rustup proxy, not Cargo itself; with a valid Homebrew Cargo 1.96 installation the documented builder fails with no such command: +1.96.0, and --cargo cannot opt out because the prefix is unconditional. Please make the toolchain selector optional or invoke rustup run <toolchain> cargo explicitly, and cover both command shapes in a builder test.

python/aisimulate/.agents/skills/aisimulate-replay-parity/scripts/replay-parity-runner/src/main.rs
dreamtalen: Fixed in d39116b. source_revision is removed from editable configs and is now rejected as an unknown field. build_runner.py embeds git rev-parse HEAD into the binary through AISIMULATE_SOURCE_REVISION, and the runner reports only that compile-time value (or null for an explicitly unbound direct build). I also added a negative test for user-supplied source_revision and verified a runner built from the candidate checkout reports the candidate SHA while preserving the semantic canonical digest.

python/aisimulate/.agents/skills/aisimulate-replay-parity/scripts/build_runner.py
dreamtalen: Fixed in d39116b. --toolchain is now optional: without it the builder invokes the selected --cargo executable directly; with it the builder uses explicit rustup run TOOLCHAIN CARGO rather than Cargo +toolchain shorthand. Added Python unit tests covering both direct Cargo and rustup command shapes, and validated the explicit rustup build end to end.

jasonqinzhou (approved):
Reviewed exact head d39116b. The two raised findings are addressed, their focused validation passed, and the conversations are resolved. Approving with the current-main canonical-digest integration risk explicitly accepted.

PR #49, opened 2026-08-26
feat(docs): add E2E accuracy overview [AIC-1836]

jasonqinzhou:
## Summary

- add a public-safe E2E Accuracy Overview sourced from the existing accuracy-evidence pipeline
- mirror the compact deployed Overview layout, including AISimulate-first summary cards and side-by-side AISimulate/AIC columns
- publish a deterministic aggregate with exact-cell coverage, TTFT/TPOT accuracy, curve-shape error, failure counts, and reproducible provenance
- fail closed on mismatched release URLs and unknown hardware without explicit multi-node scope
- add client-side URL/schema guards, focused tests, ownership routing, and links from the AISimulate documentation landing page and repository README
- intentionally exclude the exploratory E2E Gym and internal-only navigation from the public surface

## Published snapshot

- source: SemiAnalysis InferenceX `db-dump/2026-08-24`
- scope: 759 single-node measured points; 737 multi-node rows excluded
- AISimulate 0.12.0: 754 successful predictions, 5 explicit failures, 99.34% evidence coverage
- AISimulate accuracy: 30.09% TPOT MAPE, 76.06% TTFT MAPE
- AISimulate curve shape: 17.94% TPOT, 31.40% TTFT
- AIC baseline: 30.85% TPOT MAPE, 51.47% TTFT MAPE

## Evidence boundaries

- numbers apply only to the exact measured cells in this snapshot; they are not universal support or deployment certification
- client-observed TTFT and TPOT remain separate, and no pass/fail threshold is imposed
- raw measured latencies, internal links, and internal run identifiers are excluded from the public bundle
- a live unauthenticated public URL remains owned by [PR #64](https://github.com/ai-dynamo/aisimulate/pull/64) / OPS-8319; this PR lands the versioned page and data, not the deployment

## Validation

- `pytest -q tests/test_e2e_accuracy_overview.py` (12 passed)
- static-page branding tests (3 passed)
- CODEOWNERS policy tests (99 passed) and strict ownership coverage (3,619/3,619)
- `ruff check` and `ruff format --check`
- JavaScript syntax and JSON structure checks
- byte-for-byte deterministic regeneration of `summary.json`
- `git diff --check origin/main...HEAD`

## Tracking

- AIC-1836
- OPS-8319 for live public hosting

PR #50, opened 2026-08-26
docs: rebrand static pages for AISimulate [AIC-1836]

jasonqinzhou:
## Summary

- rebrand the static-docs landing page and repository links for AISimulate
- label the existing compatibility page as the **Legacy AIC Support Matrix**
- direct users to the FPE Support Matrix for current strict-native forward-pass estimator coverage while keeping the evidence boundary explicit
- remove release-branch discovery, selection, caching, and hardcoded fallbacks because AISimulate currently has no release branches; the legacy page reads `main` only
- hide the outdated Architecture Universe and remove its remaining README entrypoint
- update DeepWiki, support-matrix source, and raw-data README links for AISimulate
- add focused regression coverage for branding, navigation, the main-only data path, and README entrypoints

## Validation

- `uv run pytest -q -c python/aisimulate/pytest.ini python/aisimulate/tests/unit/test_static_docs_branding.py` (3 passed)
- `ruff check` and `ruff format --check`
- strict CODEOWNERS coverage: 100% (3,594/3,594 files); policy suite: 99 passed
- local browser render loaded 7,367 PASS and 2,955 FAIL rows from the checked-in AISimulate `main` snapshot
- visual verification of the legacy notice, FPE call-to-action, and main-only branch label

## Visibility and dependency boundary

The legacy matrix now reads only the snapshot checked into AISimulate `main`; it does not use a public AIConfigurator compatibility mirror or pinned AIC commit. The page and its data remain private/local while AISimulate is private, and the same AISimulate URL becomes anonymously readable when the repository and hosting become public.

This PR versions the rebranded static pages but does not create a new AISimulate Pages deployment. The FPE call-to-action targets `docs/fpe-support-matrix/`, supplied by draft PR #41, so that route remains a separate dependency.

## Stack

This is the foundational static-pages PR. PR #49 is stacked on this branch so its diff stays focused on the E2E Accuracy Overview.


python/aisimulate/docs/support-matrix/index.html
```diff
@@ -50,12 +50,11 @@
 const { useState, useEffect, useCallback, useMemo, useRef } = React;
 
 // ── Constants ───────────────────────────────────────────────────────
-const REPO = 'ai-dynamo/aiconfigurator';
+const REPO = 'ai-dynamo/aisimulate';
```
jasonqinzhou, line 53: [P1] This changes every non-local matrix fetch to the internal AISimulate repository. An unauthenticated request to the exact index currently returns 404, and the AISimulate Pages path is not published, so a public deployment cannot load the matrix. Also, AISimulate has no release/* refs while the selector still falls back to release/0.8.0 and release/0.7.0, making its advertised release choices fail. Please publish/bundle a public snapshot for main and remove or replace the stale release fallback.

python/aisimulate/docs/index.html
```diff
@@ -3,7 +3,7 @@
 <head>
 <meta charset="UTF-8">
 <meta name="viewport" content="width=device-width, initial-scale=1.0">
-<title>AI Configurator</title>
+<title>AISimulate</title>
```
jasonqinzhou, line 6: [P2] The landing page is rebranded and hides Architecture Universe, but python/aisimulate/README.md still promotes the outdated Universe and sends support-matrix users to the old AIC Pages site; its raw-data link also targets the obsolete aic-core path. Please update or remove those remaining public entrypoints as part of this rebrand.

python/aisimulate/docs/support-matrix/index.html
jasonqinzhou: Fixed in 221d2f9. Public main data is pinned to the byte-identical AIC compatibility snapshot, anonymous browser loading is verified, local preview uses the checked-in AISimulate path, and all fallback release refs were verified across split and legacy layouts.

python/aisimulate/docs/index.html
jasonqinzhou: Fixed in 221d2f9. The outdated Universe entrypoint and old AIC Pages link are removed; DeepWiki, support-matrix source, and raw-data links now resolve within AISimulate.

python/aisimulate/docs/support-matrix/index.html
jasonqinzhou: Follow-up in 088c829: per product direction, the public AIConfigurator compatibility mirror and pinned AIC commit are gone. The matrix now reads only the snapshot in AISimulate `main`, remains private/local while the repository is private, and becomes publicly readable from the same AISimulate source when the repository and hosting go public. The release selector also remains removed until AISimulate has real release refs.

simone-chen (approved):

PR #56, opened 2026-08-28
perf(moe-a2a): retain interpolation grids

jthomson04:
## Summary

- Retain one exact-SMS `AxisCurve` and one lazy `PreparedGrid` for each MoE A2A slice.
- Build one lazy combined-dtype grid map per DeepEP-HT dispatch slice and reuse its surfaces for dispatch and combine queries.
- Remove the unused stateless Grid query path now that all internal callers use prepared owners.
- Avoid per-query shape-string allocation while preserving dtype fallback, exact-SMS exceptions, ragged intersections, errors, and interpolation results.
- Preserve the parent result when DeepEP-HT dispatch and combine have no shared combined surface.

## Validation

- focused `moe_a2a` release tests: 16 passed
- focused `perf_interp` release tests: 39 passed
- `cargo test -p aisimulate-core --release`: 1,097 passed; 1 ignored

### End-to-end forward-pass prediction A/B

Measured the warm host time for the public full-model static prediction call,
`InferenceSession.run_static_latency_only`. The workload was a DeepSeek-R1
large-EP prefill prediction on H200 SXM with SGLang 0.5.14 and the SILICON
database: batch 1, ISL 1024, OSL 2, TP1, attention DP8, MoE TP1/EP8,
DeepEP-HT, and SMS 18. SMS 18 selects the two-dimensional A2A interpolation
path. The timed boundary excludes process start and model/database setup.

Each arm used its own release runtime, ran on CPU 0 with `PYTHONHASHSEED=0`,
and completed 20 warmups plus 10,000 measured predictions with no errors.

| Metric | Base `dd1e819` | Head `1c2b1cf` | Change |
| --- | ---: | ---: | ---: |
| Full-model prefill prediction host time, median | 28.502 us | 6.312 us | 4.52x faster; 77.9% lower |
| Full-model prefill prediction host time, mean | 28.577 us | 6.349 us | 4.50x faster; 77.8% lower |

Output parity was exact. Both arms predicted `177.74980685943262 ms`
(`f64` bits `406637fe6af4728c`). The per-operation breakdown included
`context_moe_dispatch` and `context_moe_combine`, and both used the `silicon`
source.

- Case SHA-256: `b4537b2475b686e3aa645eccde3db53ebf0a1aba3364b73218a1cc0abbfe5bab`
- Benchmark SHA-256: `8e1264d2f69e74b3ab45c74eb227767317a284aa8822a0e944b87935d0af5a7a`
- Exact functional parity SHA-256 after removing PID: `9e7124e29a542c7d49304862b5d06b5aa9b931eb502c590d23301597f56d4c06`
- Base AISimulate source: `dd1e819e4b5b304f457f387d950412357eeff7ae`
- Head AISimulate source: `1c2b1cf7f5aeeedd4d5f8f3ff78d0bf14c6da2b2`
- Base runtime SHA-256 / build ID: `e79ca358aad9bf7cfbfdc8b5672c03152d601faafee1cea470b6dc5d1c3ed6be` / `dde959872a11e2c3a2d6688d5ea3767408116e62`
- Head runtime SHA-256 / build ID: `c68dedab8ce25e45f1f1e51f06957410a5bcf54a1fe526395ddf2fcedc2371e2` / `e1b0c168feedba49c806f01c52c4441305bb2ebc`

This is one serial process per arm on one host. It is direct evidence for this
fixed synthetic workload, not a host-to-host or run-to-run variance estimate.
The host prediction time is not engine latency.

The benchmark harness and result artifacts remain uncommitted.


tianhaox (commented):
Nice speedup. Verified locally: crate builds, the 84 tests in the touched modules pass, error strings are byte-identical to base, and the new `OnceLock` caches are equivalent to the per-call rebuild they replace. Two non-blocking suggestions inline, both about silent-fallback paths that the current tests can't catch.

crates/core/src/perfmodel/perf_database/moe_a2a.rs
```diff
@@ -378,29 +451,28 @@ impl MoeA2aTable {
...
+                comm_backend,
+                other_phase,
+                &self.data_root,
+            )?;
+            if let Some(other) = other_shapes.get(&shape) {
+                let (dispatch, combine_grids, combine_dtype) = if phase == "dispatch" {
+                    (slice, other_phase_grids, other_dtype)
+                } else {
+                    (other, phase_grids, used_dtype)
                 };
```
tianhaox: This tuple swap is correct today, but it's a trap: `combined` hangs off the *dispatch*-dtype slice while the map is keyed by the *combine* dtype. If someone later swaps `used_dtype`/`other_dtype` in either arm, it compiles and the suite still passes, because every deepep_ht test uses the same dtype name for both phases. It only misfires when the two phases resolve to different names (e.g. dispatch collected as `fp8_block`, combine only as `fp8`), and then `combined.get(combine_dtype)` misses and line 472 silently returns the un-apportioned `latency` with no error.

Two cheap things worth doing in this PR:
1. Add a test where dispatch and combine resolve to different dtype names, so the tuple order is pinned.
2. Consider making the `combined` miss an error (or at least a `debug_assert!`) rather than a silent fallback, since a wrong number here looks plausible.

Longer term, `combined` is a dispatch-only field living on the generic per-shape slice (dead on nvlink / deepep_ll / trtllm / combine slices, invariant enforced only by the comment at line 109). Hoisting it to load time as a per-backend `BTreeMap<(MoeA2aShapeKey, dispatch_dtype, combine_dtype), OnceLock<PreparedGrid>>` would let `SmsGrid` be just `{curves, grid}` and remove the direction normalisation in `query` entirely. All dtypes are known at load, so this is straightforward. Happy to leave that for a follow-up.

crates/core/src/perfmodel/perf_database/moe_a2a.rs
```diff
@@ -246,35 +347,38 @@ impl MoeA2aTable {
...
-            sms,
         };
-        Ok(grids
-            .by_keys
-            .range(key_at(0)..=key_at(u32::MAX))
-            .next()
-            .is_some())
+        Ok(shapes
+            .get(&shape)
+            .is_some_and(|slice| !slice.curves.is_empty()))
```
tianhaox: This is an `Option`-flavoured copy of the same `fp8_block -> fp8 -> sole default` chain that `resolve_dtype` (line 165) implements. Now that the PR introduces `resolve_dtype`, the two can drift: an alias added only there makes `has_shape` return `false` for a shape `query()` happily resolves, and the sole caller (`operators/moe_a2a.rs` ~L182) then sets `use_node1_fallback` and silently returns a node=1 donor estimate instead of the collected multi-node data.

Suggest factoring one `Option`-returning helper, e.g. `select_dtype(phase_grids, requested) -> Option<(&str, &ShapeGrids)>`, and having `resolve_dtype = select_dtype(..).ok_or_else(err)` while `has_shape` uses it directly.

Also, the trailing `.is_some_and(|slice| !slice.curves.is_empty())` can't be false: every `SmsGrid` is built in `load_moe_a2a_grids` from an entry that had at least one sms curve inserted. `Ok(shapes.contains_key(&shape))` says the same thing.

jthomson04:
/ok to test [c93eb7a](https://github.com/ai-dynamo/aisimulate/pull/56/commits/c93eb7a2832c9cf8d6b018ffab503f20a98bae6d)

tianhaox (approved):

PR #57, opened 2026-08-28
feat: simulate native vLLM host offload

dreamtalen:
## Summary

This PR adds native vLLM CPU KV-cache offload (G1↔G2) to AISimulate and exposes it through `aisimulate predict` and `recommend`.

It implements the initial vLLM phase of [ai-dynamo/dynamo#13794](https://github.com/ai-dynamo/dynamo/issues/13794):

- `HostTier` provides reusable G2 capacity, residency, LRU, pins, and D2H/H2D transfer timing.
- `VllmHostOffloadAdapter` retains vLLM-specific lookup, store, load, and touch policy.
- The vLLM G1 manager owns destination reservations, source-reuse dependencies, and scheduler fences.
- Replay advances host transfers through virtual-time deadlines; `HostTier` does not own the clock.

The implemented profile matches vLLM's default `OffloadingConnector` behavior: prompt-only, missing-only asynchronous write-through, asynchronous load, and LRU eviction. It does not restore KVBM or model G3/G4.

## Public configuration

`workers.<role>.kv_cache.bytes_per_token` describes the physical KV-cache geometry for that worker role and resolves independently from its TP/PP/MoE shape. `engine.kv_transfer.bytes_per_token` remains a separate PD-link payload control; its `auto` default resolves from the prefill/source role. Both fields may be set when physical cache and transfer representations differ.

```yaml
engine:
  mode: aggregated
  model: MiniMaxAI/MiniMax-M2.5
  hardware: b200_sxm
  backend: vllm
  backend_version: "0.24.0"
  context_length: 131072
  workers:
    aggregated:
      parallelism: {replicas: 1, tensor: 4, pipeline: 1, attention_data: 1, moe_tensor: 4, moe_expert: 1}
      scheduler: {max_batched_tokens: 8192, max_sequences: 256}
      kv_cache:
        block_size: 32
        prefix_caching: true
        bytes_per_token: auto
        capacity: {type: fixed, blocks: 54453}
        host_offload:
          num_host_blocks: 108906
          d2h_bandwidth_gbps: 32.0
          h2d_bandwidth_gbps: 32.0
      timing: {type: default}
```

```bash
aisimulate predict --stack engine --config host-offload-prediction.yaml
```

`num_host_blocks` is independent of G1 capacity. D2H and H2D remain separate knobs even though both defaults are 32 GB/s. Recommendation initially accepts only a fixed host-offload descriptor; these fields are not search dimensions.

## Limitations

This PR implements only the vLLM profile. Unsupported combinations fail closed rather than silently ignoring host offload:

- **Aggregated workers only.** Disaggregated replay has separate prefill/decode cache ownership and a PD KV handoff; the ordering and dependency relationship between that handoff and each role's host tier are not modeled yet.
- **Prefix caching must be enabled.** The native profile identifies reusable prompt blocks through vLLM's prefix-cache hashes and lookup path; without prefix caching there is no validated block-identity/reuse contract for this adapter.
- **Attention DP must be 1.** Per-DP-rank G2 ownership, request placement, capacity/bandwidth accounting, and rank-local artifact identity are not modeled yet.
- **Native speculative decoding/MTP is unsupported.** MTP recomputes the final cached prefix block and may release rejected speculative reservations; those rules have not yet been composed with H2D reservations and D2H source fences.

SGLang/TRT-LLM profiles and optional vLLM frequency/CPU-lookup filters are follow-up work described by the DEP. Because only one vLLM host-offload profile exists today, `backend=vllm` selects it implicitly rather than introducing a separate profile-version field.

## Validation

- Full Rust workspace tests, Python prediction/recommendation/runner tests, Clippy with `-D warnings`, rustfmt, Ruff, and `git diff --check` passed.
- Tests cover host-tier capacity/LRU/pins, FIFO transfer timing, scheduler boundaries, H2D reservations, D2H dependency fencing, preemption/cancellation, reports/artifacts, CLI lowering, legacy schema compatibility, and unsupported-scope rejection.
- G1-only replay parity was rerun after rebasing onto current `main`; no regression was found when host offload is absent.

### Real B200 parity

The timing result was repeated with two independent cold-start G1/G2 pairs using MiniMax-M2.5, vLLM 0.25.1, TP4 on 4×B200, BF16 KV, block size 32, and 5,000 fixed-schedule Mooncake requests. Each GPU column shows the two-run range; MAPE is the mean absolute percentage error across those runs.

#### G1 only

Configuration: 54,453 G1 blocks; host offload disabled.

| Metric | GPU result | AISimulate / Mocker | MAPE |
|---|---:|---:|---:|
| Prefix-cache reuse | 34.4783%-34.5045% | 34.3443% | 0.43% |
| Mean TTFT | 398.988-402.875 s | 358.619 s | 10.55% |
| Mean TPOT / ITL | 134.378-135.466 ms | 125.979 ms | 6.63% |

#### G1 + G2

Configuration: 54,453 G1 blocks and 108,906 G2 blocks. AISimulate uses the default 32/32 GB/s D2H/H2D bandwidth; the first GPU pair measured connector-effective bandwidth of 33.447/45.425 GB/s.

| Metric | GPU result | AISimulate / Mocker | MAPE |
|---|---:|---:|---:|
| G1 cache reuse | 34.5388%-34.6790% | 34.3260% | 0.82% |
| G2 cache reuse | 7.5385%-7.6956% | 7.8177% | 2.65% |
| Combined cache reuse | 42.2174%-42.2344% | 42.1438% | 0.19% |
| Mean TTFT | 385.248-392.471 s | 332.021 s | 14.61% |
| Mean TPOT / ITL | 132.327-134.714 ms | 119.181 ms | 10.73% |

Both cold-start pairs showed a positive G2 benefit: cache reuse improved by 7.71-7.76 percentage points, TTFT by 2.58%-3.44%, and TPOT by 0.56%-1.53%.

## Suggested review order

1. `crates/core/src/engine/host_offload/tier.rs` — Generic G2 capacity, LRU, pins, and transfer timing.
2. `crates/core/src/engine/scheduler/vllm/host_offload.rs` — vLLM-specific host-offload policy.
3. `crates/core/src/engine/kv_manager/vllm_backend.rs` — G1 reservations and source-dependency fencing.
4. `crates/core/src/engine/scheduler/vllm/core.rs` — vLLM scheduler integration.
5. `crates/core/src/replay/agg.rs` — Host-transfer virtual-time integration.
6. `python/aisimulate/src/aisimulate/config/engine.py` and `compiler.py` — Public schema, validation, and lowering.
7. `tests/test_host_offload_cli.py` and `crates/core/tests/engine.rs` — CLI and engine integration coverage.


tedzhouhk (approved):

python/aisimulate/src/aisimulate/sweeper/deploy.py
```diff
@@ -97,20 +97,24 @@ def _engine_args_payload(sample: dict[str, Any], role: str, *, backend_version:
...
-                    tp_size=tp,
-                    pp_size=int(sample[f"{prefix}pp"]),
-                    moe_tp_size=moe_tp,
-                    moe_ep_size=moe_ep,
-                )
-                if configured_bytes == "auto"
-                else int(configured_bytes)
+    host_offload = sample.get(f"{role}_native_host_offload")
+    legacy_geometry = sample.get("kv_transfer_bytes_per_token")
+    if sample.get("kv_transfer_enabled", False) or host_offload is not None:
```
tedzhouhk: [P1] Preserve legacy `SearchSpace` configs here. `kv_transfer_enabled` is new and defaults to `False`, so a pre-existing config that only sets `kv_transfer_bytes_per_token=333` now skips this block while still lowering `kv_transfer_bandwidth`; both roles lose `kv_bytes_per_token` and Rust silently disables transfer timing. Treat a non-`None` legacy geometry as enabling transfer (or migrate it in a validator), and add a regression test for a serialized pre-PR `SearchSpace` without the new flag.

python/aisimulate/src/aisimulate/config/engine.py
```diff
@@ -280,10 +294,66 @@ def _validate_roles(self) -> EngineRecommendationConfig:
...
+    ]
+
+
+def _validate_kv_geometry_authority(workers, transfer: KvTransferConfig | None) -> None:
+    if transfer is None or transfer.bytes_per_token is None or "bytes_per_token" not in transfer.model_fields_set:
+        return
+    duplicate_roles = [
+        role
+        for role in ("aggregated", "prefill", "decode")
+        if (worker := getattr(workers, role)) is not None and "bytes_per_token" in worker.kv_cache.model_fields_set
```
tedzhouhk: [P2] This fields-set check makes an accepted legacy config fail a normal round-trip. Input with only `engine.kv_transfer.bytes_per_token` passes because the worker's default `auto` is not in `model_fields_set`; `model_dump()` materializes that default, and re-validating then sees both locations as explicit and rejects the same config. Normalize the alias before defaults/serialization, or otherwise ensure dumped output cannot become self-conflicting, and add `model_dump()` → `model_validate()` tests for prediction and recommendation.

tests/test_host_offload_cli.py
```diff
@@ -0,0 +1,399 @@
...
+
+from __future__ import annotations
+
+import json
+
+import pytest
+import yaml
+from pydantic import ValidationError
+
+from aisimulate import main as cli
```
tedzhouhk, line 19: [P2] Keep source-only collection working. `tests/conftest.py` ignores the existing runtime suites when the standalone `aisimulate` package/native extension is unavailable, but this new file imports it unconditionally and is not in `collect_ignore`. `pytest --collect-only tests/test_host_offload_cli.py` currently fails with `ModuleNotFoundError`. Add this file to the same guard or move it under an already guarded suite.

crates/core/src/engine/scheduler/vllm/core.rs
```diff
@@ -931,6 +1122,12 @@ impl VllmCore {
         self.state.is_empty()
     }
 
+    pub(crate) fn is_ready(&self) -> bool {
+        !self.state.running_members.is_empty()
+            || !self.state.waiting_members.is_empty()
+            || !self.state.connector_waiting_members.is_empty()
```
tedzhouhk: [P2] Requests that are only waiting for an H2D/store deadline should not make the worker runnable. If every request is deferred in `connector_waiting`, each unrelated coordinator timestamp starts an effect-free pass that scans, removes, and requeues all waiters below, producing an `O(events × waiters)` hot path. Track deadline-only waiters separately, or make `is_ready()` false until `next_internal_deadline_ms()` wakes them.

tedzhouhk (commented):
Agentic review: inline comments for findings 2, 3, 4, and 6.

python/aisimulate/src/aisimulate/config/engine.py
dreamtalen: As we discussed, `engine.kv_transfer.bytes_per_token` and `workers.<role>.kv_cache.bytes_per_token` are now independent rather than aliases, since the transfer payload geometry may differ from the physical cache geometry for that role. This also removes the fields-set compatibility plumbing, and both prediction and recommendation configs now round-trip cleanly.

python/aisimulate/src/aisimulate/sweeper/deploy.py
dreamtalen: Good catch. I removed `kv_transfer_enabled`; a non-`None` legacy `kv_transfer_bytes_per_token` now directly enables and lowers the PD transfer geometry. Added regression coverage for a serialized pre-PR `SearchSpace`.

tests/test_host_offload_cli.py
dreamtalen: Fixed. I added this test to the source-only `collect_ignore` guard and added a module-level `importorskip`. It now skips cleanly when the standalone runtime is unavailable.

crates/core/src/engine/scheduler/vllm/core.rs
dreamtalen: Fixed. Deadline-only connector waiters now retain FIFO membership in a separate blocked set without making the worker runnable. They are woken only by host-transfer progress or load cancellation, with coverage confirming we do not perform effect-free selections before wakeup.

PR #60, opened 2026-08-31
feat(replay): add policy-neutral telemetry sampling

sachalmalick:
Replay now exposes policy-neutral, virtual-time telemetry through a dedicated observer instead of extending the Planner scaling snapshot. Sampling is opt-in, independent of Planner ticks, and observational with respect to replay accounting, caps, and deadlock detection.

### How This Was Implemented

- Added typed baseline, periodic, and final telemetry snapshots for arriving/completed traffic, queue state, live rank gauges, role counters, and topology.
- Added a dedicated `ReplayTelemetryObserver` and configurable simulated-time cadence without changing the `ReplayScalingPolicy` contract.
- Kept the disabled path allocation-free and bounded enabled-state memory to live ranks plus one retired-rank interval accumulator per role.
- Preserved interval counters across worker retirement without retaining per-sample history in the replay core.
- Made telemetry heartbeats cap-neutral and deadlock-neutral, including zero-duration final flushes for observations at `t=0`.

<details>
<summary>Walkthrough</summary>

- `replay/telemetry` defines the public snapshot and observer contract.
- Aggregate and disaggregated replay schedule telemetry after settled work and before scaling decisions at the same timestamp.
- Engine telemetry separates live gauges from interval counters and folds retired ranks into bounded accumulators.
- The built-in JSON runner fails fast when telemetry is requested because that runtime does not expose the observer yet.

</details>

### Validation

- `cargo test -p aisimulate-core` — 954 passed, 1 ignored; integration and doc suites passed.
- `pytest -q` — 406 passed, 5 skipped.
- `cargo clippy -p aisimulate-core --all-targets` — passed with pre-existing perf-model warnings only.
- Ruff, `cargo fmt --all -- --check`, and `git diff --check` passed.

CLOSES: LLM-192


dreamtalen (approved):
LGTM
Non-blocking follow-up: could you verify there is no perf regression when telemetry sampling is disabled, and evaluate the overhead when it is enabled?
Skill `python/aisimulate/.agents/skills/aisimulate-replay-parity` might be helpful as reference.

sachalmalick:
Verified this on the exact merge pair (`8033829…` → `5fd279b…`) using the replay-parity skill's pinned 5,000-request Mooncake slice across all five supported engine/topology rows.

Method: release builds with the same lockfile/core features, five warmups per arm, then 60 balanced adjacent pairs per row (30 in each order), with every sample in a fresh process. The timer wraps only the prepared topology runtime's `run()` loop, excluding trace preparation, collector finalization, detailed capture, and serialization. I used three binaries so features/link layout cannot confound the ratios: plain parent vs plain candidate for the disabled gate, then the same telemetry-capable candidate binary in disabled/no-op modes for enabled overhead.

Disabled telemetry passes the predeclared 1.05 median-ratio gate on every row:

| Row | Median overhead | One-sided 95% upper bound |
|---|---:|---:|
| vLLM aggregated | +0.78% | +1.17% |
| vLLM disaggregated | +0.38% | +0.84% |
| SGLang aggregated | +3.04% | +3.39% |
| SGLang disaggregated | +0.67% | +1.26% |
| TRT-LLM aggregated | +1.38% | +2.01% |

All five disabled rows passed the predeclared lag-1/order/trend diagnostics, so this rules out a population-median slowdown above 5% on this host/workload at one-sided 95% confidence (rather than claiming literally zero regression).

Enabled sampling with a counting no-op observer was effectively flat at a 5 s virtual cadence: the geometric mean of row medians was +0.036%. Row medians were -0.37% to +0.54%; clean one-sided upper bounds were +0.20% to +1.35%. The primary TRT-LLM row had a near-zero estimate but narrowly breached the lag-1 diagnostic (0.328 vs <0.30), so I retained it as inconclusive and ran an independent second-seed confirmation: +0.26% median, +0.44% upper bound, diagnostics clean. vLLM aggregated cadence sensitivity was +0.06% median at 1 s (upper +1.16%) and +0.16% at 60 s (upper +0.74%). Sample counts were stable and cadence-plausible (vLLM aggregated: 590/119/11 at 1 s/5 s/60 s; disabled always 0).

Whole-process peak RSS showed no consistent enabled signal: the 5 s geometric mean of row median RSS ratios was +0.065% (descriptive only; macOS `ru_maxrss` is coarse and not loop-local). The comparable plain candidate grew +0.993% in file size and +1.084% in Mach-O `__TEXT,__text`, both below the 5% artifact gate.

Qualification: parent/candidate canonical outputs were byte-identical across all five rows. Enabled detailed reports had identical per-request/runtime evidence; only wall-time-derived values and ~1e-12 worker-second/GPU-hour floating accumulation changed because telemetry heartbeats split the same virtual-time integral.

Two limitations/notes:

- This is intrinsic AISim sampling + callback dispatch only; downstream JSONL serialization/persistence is intentionally excluded.
- The exact parent did not compile because this PR also carried an unrelated three-line DSV4 return-type fix. I backported that exact patch to the parent and recorded its checksum so those unrelated lines were equal across arms. The host was an M4 Pro/macOS 26.5.2 without hard CPU affinity; all pair-order/trend diagnostics are reported accordingly.

PR #18, opened 2026-08-19
feat(sweeper): safely expand explicit parallel domains [AIC-1773]

jasonqinzhou:
**Review blocker:** PP/CP currently use AIC analytical timing and resource accounting. Backend validation of replay latency/throughput, including PP scheduling effects, is not established. The human review requests qualified replay support before enabling these search capabilities. A decision is pending on explicitly experimental gating versus completing that qualification; this PR is not yet ready for human re-review or merge.

Explicit PP/CP and custom parallelism values were restricted by the built-in topology menu. This change expands aggregated, prefill, and decode domains with bounded preparation, while preserving the existing Bayesian/random optimizer and trial limits.

- Add per-role GPU, TP, PP, attention-DP, MoE-TP, MoE-EP, CP, and worker choices; total-candidate GPU constraints; and independent prefill/decode worker ceilings. Existing defaults remain PP=CP=1.
- Refuse oversized integer/shape/worker/P-D products before traversal or allocation, and cap retained pools without truncation. Share the budget across public range/preset lowering and topology preparation, reject non-advancing float ranges, and record deterministic preparation/pruning counts in `SweepResult.provenance.search_domain`.
- Propagate PP/CP through public recommendation/prediction, candidate identity, AIC timing/KV estimation, and native GPU accounting. CP inputs are limited to SGLang aggregated/prefill models with AIC CP modeling and TP=attention-DP=1; unsupported encoder/AFD, host-offload, and custom KV-geometry combinations are rejected.
- Preserve equal-GPU PP alternatives and one-GPU MoE behavior. Reject invalid pinned/custom dimensions, conflicting GPU domains, and inactive role limits. Enforce PP/CP capability checks for direct replay and native entrypoints, including AFD companion rejection.
- Add migration-guide mappings and runnable analytical PP/CP and independent P/D worker examples. Synchronize the CLI design and core API references, including CP defaults, GPU formulas, fallback rejection, and unqualified replay fidelity.

Scope: this is the bounded explicit-domain delivery of [AIC-1773](https://linear.app/nvidia/issue/AIC-1773/sweeper-restore-complete-parallel-and-execution-search-dimensions). Actual batch/context operating points, broader automatic PP/CP defaults, and exhaustive execution-search qualification remain tracked there. Scheduler capacities retain their existing semantics. Preparation guards do not replace process-memory admission or evaluation supervision (#169/#170).

[...]

PR #63, opened 2026-08-31
feat(sweeper): add EPD vision search [AIC-1774]

jasonqinzhou:
## Summary

- add a fixed canonical image workload for synthetic and Mooncake-trace replay, including one- and multi-image visual context tokens
- search capability-gated `E+agg` and `E+P+D` topologies with independently resolved encoder system/data identity, TP, batch size, worker count, timing, memory, and rate degradation
- carry the concrete encoder pool through replay, exact GPU accounting, objectives, result provenance, and fail-closed deployment artifact handling
- require explicit positive encoder power and coverage evidence instead of coercing missing data to zero
- preserve approved legacy encoder rate/TTFT semantics with differential coverage against the imported AIC implementation

## Scope boundary

- this PR provides the fixed image profile needed by AIC-1774; generic per-request image count/shape distributions remain owned by AIC-1765
- colocated/native vision execution remains fail-closed; image workloads must explicitly enable EPD
- EPD deployment artifact generation remains unsupported until the generator consumes the encoder pool contract

## Stack and dependencies

- stacked on integration commit `e364708`, which combines the current AIC-1772 heterogeneous execution work ([#22](https://github.com/ai-dynamo/aisimulate/pull/22)) and AIC-1773 parallel-dimension work ([#18](https://github.com/ai-dynamo/aisimulate/pull/18))
- the PR targets `main`, so the draft diff temporarily includes those unmerged dependency commits
- current `main` has replaced several optimizer, result-schema, runner, and support-matrix contracts since this stack was created; the branch remains conflicting and must be re-integrated as a stack rather than resolved by choosing either side wholesale
- AIC-1765 remains the dependency for the general canonical per-request multimodal workload

## Validation

- exact CI-scope Ruff check: passed
- Sweeper suite at the original EPD head: `369 passed, 3 skipped`
- repository root suite at the original EPD head: `441 passed, 3 skipped`
- packaged `Qwen/Qwen3-VL-8B-Instruct` / H200 / vLLM encoder catalog smoke: 48 concrete candidates resolved
- focused fail-closed encoder-power suite at `b1bffc3d`: `10 passed`

The stacked checkout's native one-round engine smoke reaches replay but is currently blocked by the pre-existing AIC timing-schema mismatch for `cp_size`; the AIC-1774 changed-code suites and catalog/replay contract coverage pass independently.

Linear: AIC-1774


python/aisimulate/src/aisimulate/sweeper/epd.py
```diff
@@ -0,0 +1,294 @@
...
+                    backend_key=backend_key,
+                    estimator=estimator,
+                    tp=int(row["tp"]),
+                    batch_size=int(row["bs"]),
+                    num_workers=int(num_workers),
+                    latency_ms=float(row["encoder_latency"]),
+                    throughput_rps_per_worker=float(row["seq/s"]),
+                    memory_gib_per_worker=float(row["memory"]),
+                    rate_degradation=search_space.encoder_rate_degradation,
+                    power_w_per_worker=float(row.get("power_w", 0.0) or 0.0),
```
jasonqinzhou: Missing encoder power is silently converted to 0 W / 0% coverage here, so downstream energy and provenance can look like measured zero rather than unavailable data. Please fail catalog resolution when either power field is absent, non-finite, or non-positive; EPD results should not be produced from an unproven zero-power sentinel.

jasonqinzhou: Fixed in b1bffc3d. Catalog resolution now requires explicit, positive, finite power_w and power_coverage (coverage <= 1), and the EncoderWorkerSpec contract rejects zero/missing power sentinels. Added missing/zero-power regression cases; the focused EPD suite passes 10 tests.

PR #67, opened 2026-08-31
feat(config): add server knob taxonomy and backend enums

simone-chen:
## Summary

- add one normalized server-knob group taxonomy for vLLM, SGLang, and TRT-LLM
- identify kernel knobs that the current AISimulate replay API cannot represent
- add public string enums for attention and MoE backend selectors
- normalize YAML/string inputs and reject unsupported backend values
- document the taxonomy and enum values

## Testing

- `ruff check` and `ruff format --check` on changed Python files
- 26 focused enum, compatibility, and kernel-contract tests pass
- broader modified-file run: 336 passed; 2 existing gated Hugging Face cases failed because `RedHatAI/DeepSeek-V4-Flash-NVFP4-FP8` returned HTTP 401


python/aisimulate/src/aiconfigurator_core/sdk/__init__.py
```diff
@@ -19,7 +19,9 @@
 from typing import TYPE_CHECKING, Any
 
 __all__ = [
+    "AttentionBackend",
```
jasonqinzhou, line 22: The new exports also need to be added to the repository's exact facade contract in `tests/cross_package/test_core_public_api.py`. Building and installing this head's wheel, then running that test file, gives `1 failed, 5 passed`: `test_sdk_facade_exports_the_canonical_objects` rejects `AttentionBackend` and `MoEBackend` as extra `sdk.__all__` members. Please extend `EXPECTED_FACADE` and add object-identity assertions for both enums so the intended public API is covered and the suite remains green.

jasonqinzhou (changes requested):
**REQUEST_CHANGES · 6.8/10 · high confidence**

## What this PR does

This PR adds a normalized dashboard taxonomy for server configuration knobs and introduces public string enums that normalize and validate attention and MoE performance-model backend selectors across Task and ModelConfig.

The backend normalization is thoughtfully centralized and backed by focused Task and ModelConfig tests, while the taxonomy clearly distinguishes normalized presentation keys from backend-specific flags.

## Why this score

Codex's exact-wheel evidence supports a 6.8/10 because the repository's public-SDK facade contract now fails on the two intentional exports, which would ordinarily require changes. The canonical recommendation is BLOCKED, however, because the helper did not observe the required claude-fable-5 model, so the independent dual-review contract could not be completed and no comment is selected for posting.

python/aisimulate/src/aiconfigurator_core/sdk/__init__.py
jasonqinzhou: Fixed in 40674b1c by extending EXPECTED_FACADE and adding canonical object-identity assertions for AttentionBackend and MoEBackend. Verified against the built and installed wheel: 344 tests passed across test_core_public_api.py and the complete modified ModelConfig/Task suites; pinned Ruff check and format also pass.

jasonqinzhou (approved):
Approved at exact head 40674b1c17987e47d4ff3ebd8d43558c9e330187 after the facade-contract fix, resolution of the only review thread, local installed-wheel validation (344 passed), and the full vetted 14-job CI matrix passing on amd64 and arm64. The head also merges cleanly with current main.

PR #68, opened 2026-09-01
feat(replay): model host-scoped G2 for attention DP

Aphoh:
## Summary

- add explicit cache-domain IDs to replay engine and worker topology
- share one host-pinned G2 tier, capacity budget, and H2D/D2H transfer lanes across attention-DP ranks in the same domain while keeping G1 rank-local
- retain the single-threaded `Rc`/`RefCell` representation and borrowed host-offload observations; make the scheduler-rank ownership boundary movable without adding locks
- emit ordered, tier-aware host store and eviction events and expose them to Dynamo's replay router
- support native host offload in disaggregated attention-DP replay
- add validated Agentic Mooncake v2 dependency graphs and an AgentX-style snapshot/warmup/profile/drain lifecycle for aggregated and disaggregated replay
- normalize agentic start times from graph roots, so dependency-gated zero placeholders cannot suppress timeline normalization

This is the AISimulate prerequisite for ai-dynamo/dynamo#14087 and ai-dynamo/dynamo#14100.

## Validation

- `cargo test --offline -p aisimulate-core --lib` (990 passed, 1 ignored)
- `cargo check -p aisimulate-core`
- `cargo check -p aisimulate-core --features python`
- `cargo check -p aisimulate-core --features extension-module`
- editable AISimulate wheel build
- focused host-offload topology and disaggregated attention-DP tests
- dependent Dynamo integration: `cargo check --locked --offline -p dynamo-mocker --lib`
- dependent Dynamo integration: all 202 `dynamo-mocker` library tests passed with `replay-bench` (193 sandbox-compatible tests plus 9 local-socket tests outside the sandbox)
- a 48-lane GLM-5.2/GB200 disaggregated context-aware AIC replay completed with stable conversation session IDs and exact phase-local attention-DP pinning

## Dynamo follow-up

Dynamo draft ai-dynamo/dynamo#14100 pins commit `e34f11cc44329f5452cb2189ec059297b2464d4c` and wires host-tier events plus agentic session affinity into offline KV placement. AISimulate still needs a public, compatible crate release before the Dynamo change can pass public CI without this internal Git dependency.


tianhaox (commented):
Reviewed the full diff and verified the branch locally: `cargo test -p aisimulate-core` passes 1028/1028, including the new `host_offload_topology` tests. The core mechanics check out — I specifically traced the shared-domain wakeup path (`completion_generation` cursor + lazy `is_due` re-evaluation in `GeneralizedMockerEngine::process_internal_work`) and found no lost-wakeup or livelock hazard; stashed completions are covered by `has_pending_work()`, and the `expect` on KV byte geometry in `build_with_cache_domains` is unreachable since both `EngineFactory` constructors validate first. Nice work overall.

Two things I'd like resolved before merge:

**1. Cross-rank eviction event attribution needs an explicit contract.**
In a shared domain, rank A's store can evict a block that rank B stored earlier. `HostKvEventTransactions::stage_store` publishes the `Removed` (HostPinned) event on rank A's stream, while the original `Stored` went out on rank B's stream — and `KvEvent` carries only `dp_rank` + `tier`, not `cache_domain_id`. A downstream indexer keyed by `(dp_rank, tier)` would leak stale HostPinned entries and overestimate host hits. Since ai-dynamo/dynamo#14100 consumes these events and this contract becomes hard to change once a public crate release ships, could you either (a) include `cache_domain_id` in host-tier events, or (b) confirm the Dynamo side merges HostPinned streams per domain and document on `KvEventTier::HostPinned` that `dp_rank` identifies the publisher only, with block ownership scoped to the cache domain?

**2. Test coverage for the scenario above.**
The new tests cover shared capacity/bandwidth and deterministic reuse, but not the event-stream shape when one rank evicts another rank's block. A two-rank same-domain test asserting the resulting Stored/Removed sequence would pin down whichever contract you choose in (1).

Non-blocking notes (fine as follow-ups or PR-description updates):
- `HostCacheRankHandle::set_observer` mutates the domain-global observer from a rank-scoped handle (last writer wins). Current replay wiring sets the same `Arc` on every rank so it's fine today, but a comment or debug_assert would help API users.
- `route_decode` now passes real request hashes for **all** disaggregated replays, not just host-offload ones — custom decode placement adapters will observe hashes where they previously got `None`. Worth calling out explicitly as a behavior change in the PR description.
- `KvEventPublisher`'s `Rc<Cell>` makes vLLM engines `!Send` even without host offload, and serialized `KvEvent` payloads now include the `tier` field. Both belong in the changelog for the public crate release this PR already calls for.

dreamtalen:
Thanks @Aphoh for adding this. I'd like to clarify the intended simulation scope of this shared g2 model.
In my previous DEP https://github.com/ai-dynamo/dynamo/issues/13794 and implementation https://github.com/ai-dynamo/aisimulate/pull/57, G2 refers to the framework-native CPU host memory tier, and in vllm the cpu tier is scoped to each DP engine instance rather than shared by all ranks on the same host. Are you targeting a specific vllm deployment where dp ranks can share the cpu cache through some external connector?

That said, I think the shared g2 is very useful to simulate for sure, for example some workloads in agentx testing used mooncake shared g2 solution. Would you be open to make this PR a more generic shared g2 solution? Where the user can configure the g2 is host scoped shared or cluster shared easily?

PR #69, opened 2026-09-01
fix(fpm): add GLM-5.2/MiniMax-M2.7 profiles and fix extrapolation

simone-chen:
## Summary

- Add bundled B200/vLLM 0.25.1 FPM data for GLM-5.2 DEP8 and TEP8.
- Add bundled H200/vLLM 0.25.1 FPM data for MiniMax-M2.7.
- Add a baseline-only decode left-boundary hold for mixed-step marginal composition.
- Preserve explicit `forward_model` selection in native AIC replay.
- Query FPM decode with exact scheduler totals and align inferred KV capacity to the collected decode domain.
- Document the new built-in FPM coverage and baseline policy.

The GLM source contains separate DEP8 and TEP8 shards. This PR combines them into one atomic parquet/metadata pair because both share the same system, backend, and version path. The sidecar preserves each source digest and row count.

## Mixed-step baseline extrapolation

Mixed steps subtract a synthetic decode pass baseline. That baseline can fall below the minimum KV collected for the padded decode batch curves, even though the real decode workload is covered.

The new policy is intentionally narrow:

- Exact decode batches use the measured latency at that batch curve minimum KV.
- Off-lattice batches hold each padded bracket curve at its own minimum KV, then interpolate the two floor latencies along the batch axis.
- Ordinary decode queries remain strict and never extrapolate below collected curve coverage.
- The collector and parquet schema are unchanged.

This removes false mixed-step coverage misses without weakening normal decode coverage checks.

## S1 low-KV replay recovery

The reported 8k1k TP4 failures queried `(batch=1, prefill=8066, KV=1)` and `(batch=1, prefill=7938, KV=2)`. Both points are inside the collected token domain and are bracketed by the same-batch `(prefill=4096/8192, KV=0/16)` curves.

The failed dashboard results were produced at commit `9db8609`, before the low-KV site-admission fix in `#39`. Current interpolation admits nearby raw-KV sites while keeping the normal batch-distance gate. No additional token-axis extrapolation is introduced here.

Current native replay also rejected the recipe's explicit `forward_model="fpm"`. This PR restores that optional field and passes it to `compile_engine`; omission still selects the default op-level model.

| Concurrency | Before | After | Completed requests | Mean TTFT | Mean TPOT |
|---:|---|---|---:|---:|---:|
| 128 | Replay failed at `(1, 8066, 1)` | Success | 1,280/1,280 | 1,908.72 ms | 60.78 ms |
| 256 | Replay failed at `(1, 7938, 2)` | Success | 2,560/2,560 | 3,820.57 ms | 132.56 ms |

Direct real-database predictions for the former failure coordinates are 198.125 ms and 194.966 ms, respectively.

## S1 decode-ceiling recovery

DEP4 concurrency 256 previously failed for 8k1k and 1k8k at `total_kv_read_tokens=546048`, two tokens above the collected decode ceiling of 546046. The legacy adapter reconstructed this coordinate from a rounded mean context, hiding the larger capacity mismatch: AIC inferred 551728 runtime KV tokens while the FPM database covers 546046.

The fix keeps decode non-extrapolating:

- FPM replay now sends the exact scheduler past-KV total instead of reconstructing it from a rounded mean. Op-level replay keeps its existing coordinate path.
- When `num_gpu_blocks` is inferred, replay caps it to the FPM decode ceiling on a whole-block boundary: `floor(546046 / 16) = 34127` blocks, or 546032 tokens.
- Explicit user capacity remains authoritative and is not changed.
- The Python-to-Rust execution descriptor preserves whether capacity was explicit or inferred.

| Workload | Topology | Concurrency | Before | After | Completed requests | Mean TTFT | Mean TPOT |
|---|---|---:|---|---|---:|---:|---:|
| 8k1k | DEP4 | 256 | Decode coverage failure | Success | 2,560/2,560 | 6,258.22 ms | 81.66 ms |
| 1k8k | DEP4 | 256 | Decode coverage failure | Success | 2,560/2,560 | 2,056.98 ms | 43.18 ms |

The complete MiniMax-M2.7 S1 replay grid improves from 79/81 to 81/81 successful points.

## Validation

- Validated schemas, checksums, row counts, identities, latency values, and duplicate keys for both pairs.
- Exercised exact-row GLM DEP8/TEP8 and MiniMax prefill/decode queries through the Rust FPM loader.
- Added operator and engine regressions for per-curve floor holding, off-lattice batch interpolation, strict ordinary decode behavior, native FPM timing selection, exact decode totals, capacity-origin propagation, and decode-domain capacity alignment.
- Ran `cargo test -p aisimulate-core --features python`: 993 passed, 1 ignored.
- Ran `pytest tests/test_runner.py -q`: 38 passed.
- Ran `pytest python/aisimulate/tests/unit/sdk/test_fpm_forward.py -p no:timeout -q`: 17 passed.
- Built the wheel, verified all four data files are packaged, and replayed both DEP4 concurrency-256 workloads plus the full 81-point S1 grid.
- Ran dashboard importer tests: 14 passed.
- Ran dashboard web tests: 59 passed; TypeScript typecheck passed.
- Ran strict CODEOWNERS validation.

Strict same-commit replay against the MiniMax-M2.7 server logs:

| Workload | Before coverage | Before MAPE | After coverage | After MAPE |
|---|---:|---:|---:|---:|
| 1k1k TEP4 mixed | 306/731 | 6.78% | 731/731 | 8.60% |
| 1k1k TP4 mixed | 330/731 | 4.57% | 731/731 | 6.68% |
| 1k1k DEP4 mixed | 72/240 | 11.25% | 240/240 | 18.92% |
| 8k1k TP4 mixed | 373/5,232 | 3.23% | 5,554/5,554 | 5.08% |

The before MAPE is calculated only over the previously successful subset. The after MAPE includes the newly covered, harder samples, so the two values do not describe identical sample sets. For 8k1k, recompute-aware reconstruction also changes the observed mixed-step denominator from 5,232 to 5,554.

- FPM perf-database prediction errors: 3,477 -> 0.
- S0, prefill, and pure-decode predictions are unchanged.

The recompute-aware parser restores the previously unavailable 8k steps. It reports 333 direct recompute inferences and 2,392 downstream reconstructed mixed steps separately from exact reconstruction.


crates/core/src/perfmodel/operators/fpm_forward.rs
```diff
@@ -177,8 +194,52 @@ impl FpmForwardOp {
...
+        let batch = batch_size as f64;
+        let latency = if let Some((lo_row, hi_row)) = Self::decode_bracket_rows(cell, batch) {
+            let lo_value = row_floor(lo_row)?;
+            if hi_row == lo_row {
+                lo_value
+            } else {
+                let hi_value = row_floor(hi_row)?;
+                let weight = (batch - lo_row as f64) / (hi_row as f64 - lo_row as f64);
+                lo_value + (hi_value - lo_value) * weight
+            }
```
liyuanzhe1991, line 249: `query_pass_baseline` always blends **both** bracket rows' floors, but the decode query it pairs with (`decode_bracket`, L423-426) degrades to a **single** row when only one curve covers the KV coordinate — so `query(B, KV) − query_pass_baseline(B)` can mix two different curve models and the shared-pass cost no longer cancels.

Concrete case on this PR's own test lattice (`pass_baseline_holds_each_curve_floor_then_interpolates_batch`), mixed step with `num_decode_requests=15, sum_decode_kv_tokens=20`:

- decode query `(15, 20)`: row 16's curve `[32, 64]` doesn't cover kv=20 → degrades to row 9 alone → `5.0 + 1.0·(20−18)/(64−18) ≈ 5.043`
- baseline `(15)`: blends both floors regardless → `5.0 + (9.0−5.0)·(15−9)/(16−9) ≈ 8.429`
- marginal = `(5.043 − 8.429).max(0)` = **0** — the decode riders are priced as free. The coverage-consistent answer (baseline degraded to row 9's floor = 5.0, same rule as the query) is ≈ **0.043**.

In the mirror band (kv covered only by the higher row) the marginal roughly doubles instead of clamping to 0. Since adjacent batch curves' collected KV ranges are ragged on every real table, these mismatch bands are always present.

Suggested fix: pass the KV coordinate into `query_pass_baseline` and apply the same covered-row selection as `decode_bracket`, so both sides of the subtraction are always computed on the same rows.

jasonqinzhou, line 249: The decode query drops to the single bracket row that covers `total_kv_read_tokens`, but this baseline always interpolates both rows' floor latencies without seeing that KV coordinate. Mixed composition can then subtract a baseline from a different curve mixture: on the new test lattice at `(batch=15, kv=20)`, decode uses row 9 only (about 5.043 ms) while the baseline blends rows 9 and 16 (about 8.429 ms), so the marginal decode cost is clamped to zero. The shipped cells have the same ragged one-sided coverage bands. Please pass the decode KV total into `query_pass_baseline`, apply the same covered-row selection as `decode_bracket`, and add one-sided-coverage tests for both mixed-step call paths.

jasonqinzhou (changes requested):
**REQUEST_CHANGES · 6.6/10 · low confidence**

## What this PR does

This PR bundles exact-version GLM-5.2/B200 and MiniMax-M2.7/H200 forward-pass datasets, adds a mixed-step decode-baseline floor policy, preserves native replay's explicit forward-model selection, and changes FPM replay to use exact scheduler KV totals while capping inferred capacity to the collected decode domain.

The change is unusually careful about data provenance and runtime intent: the new parquet pairs have atomic sidecars with source digests and row counts, explicit versus inferred KV capacity is preserved across the Python-to-Rust descriptor, and focused regressions cover exact totals, capacity alignment, op-level compatibility, and the baseline-only floor hold.

## Why this score

The completed Codex pass scores the exact-head implementation 6.6 and requests changes for one confirmed P1 modeling defect: off-lattice decode can use one KV-covering bracket row while the new mixed-pass baseline blends both row floors, allowing real marginal decode work to clamp to zero. That defect is already covered by an unresolved same-head GitHub thread, so no duplicate comment is selected. The canonical recommendation is BLOCKED because the required helper requested claude-fable-5 but observed only claude-sonnet-4-6 and claude-haiku-4-5; the skill does not permit treating those fallback models as Fable or submitting a non-compliant dual-review packet.

crates/core/src/perfmodel/operators/fpm_forward.rs
Arsene12358: Confirmed and fixed in 6ec39c6.

Reproduced your case on this PR's lattice exactly: `query_totals(15, 20)` degrades to row 9 and returns `5.0 + (20−18)/(64−18) ≈ 5.043`, while the old `query_pass_baseline(15)` returned `8.428571…`, so the marginal clamped to 0. The mirror band behaves as you described — the query keeps only the upper row and the blend inflates the marginal instead.

It is reachable on the bundled tables, not only synthetically. Sweeping every bracket of the six MiniMax-M2.7 (h200) and two GLM-5.2 (b200) decode cells: adjacent capture-rung curves have ragged ceilings, so every bracket has a band covered by exactly one row, and those bands sit directly under the decode ceiling — which is where the capacity alignment in this PR deliberately parks near-capacity mixed steps. Sampling those bands, the marginal decode term shifts by a median 0.1 ms, p90 1.8 ms, and up to 10.8 ms (−75%) on the GLM-5.2 DEP8 cell's (513, 1024) bracket. Measured through the operator on the real MiniMax-M2.7 DEP4 cell:

| batch | total_kv | query | baseline before | baseline after | marginal before | marginal after |
|---:|---:|---:|---:|---:|---:|---:|
| 20 | 546,000 | 47.869 | 23.517 | 23.603 | 24.352 | 24.266 |
| 108 | 545,000 | 51.746 | 31.650 | 31.563 | 20.096 | 20.183 |
| 300 | 545,000 | 65.896 | 44.733 | 44.494 | 21.162 | 21.402 |
| 420 | 545,000 | 84.437 | 53.759 | 54.290 | 30.678 | 30.147 |
| 20 | 400,000 (both rows cover) | 51.539 | 23.517 | 23.517 | 28.022 | 28.022 |

The fix is your suggestion: `query_pass_baseline` now takes `total_kv` and selects its bracket rows with the same rule as `decode_bracket`. I pulled that rule out into a shared `FpmForwardOp::row_covers` so there is one coverage predicate rather than two copies that can drift again. Both call sites pass the coordinate their own decode query used — `sum_decode_kv_tokens` for the telemetry rank, `bs * (isl + osl//2 + 1)` for the static mixed step. When both rows cover, the blend is unchanged.

Two regressions, both failing on the pre-fix code with exactly the numbers above:

- `pass_baseline_drops_the_same_uncovered_row_as_the_query` (operator) — a lattice ragged in both directions (row 9 spans [18, 64], row 16 spans [32, 96]); asserts the baseline degrades to the covered row's floor in each band, that the marginal is that curve's own rise above its floor, and that the both-covered blend is untouched. Pre-fix: `baseline 8.428571428571429 must be row 9's floor`.
- `fpm_mixed_baseline_follows_the_query_off_a_ragged_bracket_row` (engine) — the same shape through `mixed_step_latency`, proving the mixed-step call site threads the query's own coordinate and not just the batch. Pre-fix: `got 26.47767857142857, want 25.90625`.

One boundary this deliberately does not close, for the record: on an EXACT collected batch whose own curve does not cover the coordinate, the query falls through to the resolver's cross-site transfer while the baseline still uses that row's floor. That path predates this PR, and matching it would mean replicating the resolver's site selection inside the baseline. Happy to file it separately if you want it closed.

Suites: `cargo test --workspace` 1028 passed; `cargo test -p aisimulate-core --features python` 1039 passed / 1 ignored; `pytest -q -c pytest.ini tests` 425 passed / 5 skipped; `python/aisimulate` `-m 'unit or build'` 4115 passed / 13 skipped; `cargo fmt --check` clean.

simone-chen: Fixed in 6ec39c6. Follow-up fa9559c adds the requested one-sided regression for the second mixed-step path (`ForwardPassMetrics` telemetry); the existing engine regression covers `mixed_step_latency`. Both now pass the exact decode KV coordinate into the shared baseline row selector. `cargo test --workspace` passes: 1,028 passed, 1 ignored.

simone-chen: Fixed in 6ec39c6. Follow-up fa9559c adds the requested one-sided regression for the second mixed-step path (`ForwardPassMetrics` telemetry); the existing engine regression covers `mixed_step_latency`. Both now pass the exact decode KV coordinate into the shared baseline row selector. `cargo test --workspace` passes: 1,028 passed, 1 ignored.

jasonqinzhou (approved):
Approved exact head 4a68f90 after verifying the KV-coverage-consistent baseline fix on both mixed-step paths, integration with current main, the full Python-enabled core suite, focused Python runner/FPM tests, and bundled parquet integrity. All review threads are resolved.

PR #71, opened 2026-09-01
perf(aic): add forward prediction benchmark gate

jthomson04:
## Summary

This PR adds the benchmark harness that a later advisory CI check will run. It does not add or enable the CI workflow yet.

The harness compares forward-pass prediction time between a base revision and a head revision. Each revision runs its own worker from its own source checkout and matching installed wheel. This keeps the comparison valid if internal SDK interfaces change in the pull request. The only shared interface is the versioned worker JSON protocol, currently version 1.

The worker accepts either one case for local debugging or an ordered list of cases for CI. CI sends the complete matrix as one batch per revision and round. A default full comparison uses 12 worker processes: one base/head availability pair plus five base/head measurement pairs.

Inside each worker, cases are grouped by model and database mode. Each group resets prediction state and creates one model, database view, session, and engine, then measures all nine distinct points. The current matrix therefore performs 48 group setups instead of 432 case setups during a full comparison.

## Benchmark matrix

The fixed matrix has 36 cases:

- Models: Qwen3-32B and Qwen3-235B-A22B
- Database modes: SILICON and EMPIRICAL
- Phases: four context points and five generation points from the existing prediction-regression grid
- System and backend: B200 SXM with vLLM 0.24.0

The benchmark calls `InferenceSession.run_static_latency_only` through the Rust engine-step path. It measures prediction CPU time only. It does not measure model execution, end-to-end simulation, session construction, database loading, or engine construction.

## Cold and warm definitions

Before each model/database-mode group, the worker:

1. Calls the public database eviction interface and treats it as the complete prediction-cache reset contract.
2. Releases the prior group objects and creates a new model, database view, session, and engine.
3. Primes each phase through the public prediction interface with one unmeasured `(batch_size=2, ISL=2048)` query. This point is supported by all current groups and is outside the measured matrix.
4. Times the first prediction for each distinct target point as `cold`.
5. Repeats that target ten times without recording the calls, then records 100 calls and reports their median as `warm`.

Thus, `cold` means a previously unseen target query against a steady-state session. It does not mean an empty session and does not include process startup or first-use index construction. `warm` means repeated use of the same target query. Output redirection is outside the timed calls.

The harness does not inspect or manage individual SDK caches and does not call private engine-cache helpers.

## Process and ordering model

A default full comparison starts 12 workers:

- One base/head batch pair for availability
- Five base/head batch pairs for measurement

The controller alternates which revision runs first. It also reverses case order on alternating rounds so one point does not always occupy the same position. Missing, duplicate, malformed, reordered, or unexpected batch results are blocking run errors.

## Comparison policy

- Run five paired rounds and alternate which revision runs first.
- Mark a round as over threshold only when head is both more than 10% slower and more than 2 us slower than base.
- Confirm a regression when at least 80% of the selected rounds exceed the threshold. The default is four of five. Smoke mode is one of one.
- Treat `DATA_MISS` on both revisions as skipped.
- Treat base `DATA_MISS` with a working head as skipped because there is no base timing.
- Treat a working base that becomes `DATA_MISS`, malformed worker output, worker failure, protocol mismatch, or incomplete case coverage as an invalid comparison.

Confirmed regressions and invalid comparisons make the harness return a failure result. Matched data misses, gained coverage, improvements, and unstable results do not. The future workflow will publish this result as an advisory check and will not add it to branch protection.

Raw results are written atomically after the paired availability pass and each completed paired round, so a cancelled run keeps the work that finished.

The harness writes:

- Full raw responses in `raw_results.json`
- Machine-readable comparisons in `comparison.json` and `comparison.csv`
- A compact human-readable summary in `summary.md`, with blocking results visible and noisy/full-matrix details collapsed
- Short blocking annotations in `annotations.txt`

## Scope

This PR intentionally does not:

- Add the GitHub Actions workflow. That is the follow-up PR after protocol version 1 is on `main`.
- Benchmark the full end-to-end simulation path. That interface is still changing.
- Change the benchmark models, modes, measured points, sample count, rounds, thresholds, CPU pinning, or thread limits.
- Add persistent workers, worker restart logic, NDJSON sessions, cache sentinels, CI sharding, or cache-specific orchestration.
- Change or reuse the standalone `benchmark_engine_step.py`; that tool remains identical to upstream.

## Validation

Final integration head: `e80e4a03d9dcb6226f7bc69fca11dfddfc8894c1`, including `main` at `a4c77310d7609745863ac8867c207f850beac5c3`.

- Fresh focused benchmark suite: 65 passed.
- CI workflow-contract suite: 29 passed.
- Rebuilt native package: all 36 matrix cases returned `OK` with finite predicted values.
- Scoped Ruff, formatting, and `git diff --check`: passed.
- All 10 review findings verified fixed and their threads resolved.
- [Exact-head Fast CI](https://github.com/ai-dynamo/aisimulate/actions/runs/34636304303): passed.
- [Exact-head Full CI](https://github.com/ai-dynamo/aisimulate/actions/runs/34636308762): passed (31 successful jobs; the main-only staging job skipped as expected).

### Earlier author validation

The following measurements were recorded before the final integration update:

- Ruff check: passed
- Ruff format check: passed
- Focused unit suite: 40 passed
- Optimized `maturin build --release`: passed
- The off-matrix priming point works for both phases across all four current model/mode groups
- Full 36-case availability and one-round comparison: all 144 worker results were `OK`, with four worker processes
- Full 36-case five-round comparison: passed in 25.01 seconds with 12 worker processes, no confirmed regressions, no invalid comparisons, and no skipped comparisons
- All 36 grouped predicted values exactly matched the prior per-case-isolated implementation
- Single-case and grouped-batch modes returned the same status and exact predicted value for the sampled SILICON case
- Deliberate-regression comparison: failed as expected in 25.06 seconds with the same cold and warm regressions in 5/5 rounds
- Standalone engine-step benchmark: unchanged from upstream

The inherited `--gpu-memory-utilization` golden-test failure is also present on upstream `main` and is not changed in this PR.


python/aisimulate/tools/forward_perf_gate/worker.py
```diff
@@ -0,0 +1,321 @@
...
+        moe_ep_size=case["moe_ep_size"],
+    )
+
+
+def _record_error(response: dict, exc: Exception) -> None:
+    if isinstance(exc, PerfDataNotAvailableError):
+        response.update(
+            {
+                "status": "DATA_MISS",
+                "error": {"type": type(exc).__name__, "message": str(exc).splitlines()[0][:500]},
```
tianhaox: **Blocking-ish:** `str(exc).splitlines()[0]` raises `IndexError` when the exception message is empty (`MemoryError()`, a bare `assert` such as `aiconfigurator_core/sdk/models/qwen35.py:184`, `StopIteration`). Because this runs inside the `except` handlers, the `IndexError` escapes `_run_case_group`, the worker exits 2, and the controller records `WORKER_ERROR` for the whole batch and aborts the run.

Same pattern at line 199. Suggest `(str(exc).splitlines() or [""])[0][:500]` (or `str(exc).split("\n", 1)[0]`). The existing test only raises `RuntimeError("case failed")`, so it never hits this.

python/aisimulate/tools/forward_perf_gate/worker.py
```diff
@@ -0,0 +1,321 @@
...
+        tp_size=case["tp_size"],
+        pp_size=case["pp_size"],
+        attention_dp_size=case["attention_dp_size"],
+        moe_tp_size=case["moe_tp_size"],
+        moe_ep_size=case["moe_ep_size"],
+    )
+
+
+def _record_error(response: dict, exc: Exception) -> None:
+    if isinstance(exc, PerfDataNotAvailableError):
```
tianhaox: Only `PerfDataNotAvailableError` is treated as `DATA_MISS`. The FFI in `crates/core/src/perfmodel/py.rs:106-116` deliberately maps coverage gaps to three sibling SDK types as well: `EmpiricalNotImplementedError`, `MissingSystemFlopsError`, `SolNotImplementedError`. Half the matrix runs `database_mode="EMPIRICAL"`, where a missing util basis surfaces as `EmpiricalNotImplementedError` (e.g. `operators/gemm.rs` converts a missing lookup into it in that mode).

As written, such a case becomes `INVALID`, `compare.pair_disposition` returns INVALID instead of SKIP, and `run.py` aborts at prewarm so the other 35 cases never run. Suggest treating the sibling classes as `DATA_MISS` too (or using `perf_database.has_perf_data_not_available_cause` plus the sibling types).

python/aisimulate/tools/forward_perf_gate/measurement.py
```diff
@@ -0,0 +1,203 @@
...
+            database = perf_database.get_database(case.system_name, case.backend_name, case.backend_version)
+        else:
+            database = perf_database.get_database_view(
+                case.system_name,
+                case.backend_name,
+                case.backend_version,
+                database_mode=database_mode,
+                shared_layer=shared_layer,
+            )
+        if database is None:
```
tianhaox, line 67: `get_database` / `get_database_view` return `None` (they don't raise) when no `(system, backend, version)` directory matches. Turning that into a plain `RuntimeError` means a base revision that lacks the `b200_sxm/vllm/0.24.0` data becomes `INVALID`, and the pair is INVALID → whole run aborts. The intended path per the PR description is "base DATA_MISS, head has no timing baseline → SKIP". Suggest raising `PerfDataNotAvailableError` here (or having the worker classify this branch as `DATA_MISS`).

python/aisimulate/tools/forward_perf_gate/worker.py
```diff
@@ -0,0 +1,321 @@
...
+        ensure_rust_library_present()
+        session_setup_ms, session, group_runtime_config = measure_session_setup_ms(
+            representative,
+            suppress_loader_output=True,
+            database_mode=cases[0]["database_mode"],
+            shared_layer=False,
+        )
+        phases = list(dict.fromkeys(case["phase"] for case in cases))
+        with redirect_output(True):
+            for phase in phases:
```
tianhaox: The priming call sits inside the group-level `try`, so a failure of the off-matrix `(bs=2, isl=2048)` point stamps every case in the model×mode group with that error. If it's a `DATA_MISS` on both revisions, a whole 9-point slice silently drops into the collapsed "Skipped" section, and the recorded reason names the priming shape rather than any target point.

Suggest wrapping the priming call separately: on failure, either fall back to priming with the first real case in the group, or record a distinct `PRIMING_FAILED` reason so it's visible in the summary.

python/aisimulate/tools/forward_perf_gate/worker.py
```diff
@@ -0,0 +1,321 @@
...
+                phase_case = next(case for case in cases if case["phase"] == phase)
+                phase_call(
+                    session,
+                    priming_runtime_config(group_runtime_config, phase=phase),
+                    phase=phase,
+                    stride=phase_case["stride"],
+                )()
+    except Exception as exc:
```
tianhaox: Suggestion: `except Exception` does not catch PyO3's `PanicException` (it derives from `BaseException` since pyo3 0.20). A Rust panic in `run_static_latency_only` for one case kills the whole worker process → `WORKER_ERROR` → round aborted → blocking. Same at line 271. If per-case isolation is the goal, consider catching `BaseException` here (re-raising `KeyboardInterrupt`/`SystemExit`), or explicitly `pyo3_runtime.PanicException`.

python/aisimulate/tools/forward_perf_gate/compare.py
```diff
@@ -0,0 +1,477 @@
...
+                "ratio": ratio,
+                "exceeds": exceeds,
+            }
+        )
+
+    result["rounds"] = round_results
+    if skipped_reasons and not invalid_reasons and not round_results and len(skipped_reasons) == len(rounds):
+        unique_reasons = list(dict.fromkeys(skipped_reasons))
+        if len(unique_reasons) > 1:
+            invalid_reasons.append("response status changed between measured rounds")
```
tianhaox, line 194: Nit: when all rounds are SKIP but with more than one distinct reason, this appends "response status changed between measured rounds" and then falls through to line 188, which appends the identical string again. Classification is still INVALID_COMPARISON, but the JSON/CSV/markdown show the reason twice. An `elif` or early `return` fixes it.

tianhaox (commented):
Overall the design looks right to me: revision-local workers, controller from base, grouped setup, atomic checkpoints. I traced the SDK calls (`get_database_view`, `unload_database`, `run_static_latency_only`, `RuntimeConfig.replace`) and they match. The comments below are mostly about cases where a single-point failure gets amplified into a run-wide `INVALID_RUN`, which defeats the per-case isolation the README describes. The first four are the ones I'd fix before landing; the rest are suggestions.

python/aisimulate/tools/forward_perf_gate/run.py
```diff
@@ -0,0 +1,401 @@
...
+    if args.rounds <= 0 or args.warmup < 0 or args.iterations <= 0:
+        print("error: invalid round, warmup, or iteration count", file=sys.stderr)
+        return 2
+
+    cpu = min(os.sched_getaffinity(0)) if hasattr(os, "sched_getaffinity") else 0
+    matrix_cases = case_matrix.expand_cases()
+    selected_cases = matrix_cases[:1] if args.smoke else matrix_cases
+    rounds, warmup, iterations = args.rounds, args.warmup, args.iterations
+    if args.smoke:
+        rounds, warmup, iterations = 1, 1, 3
```
tianhaox: Suggestion: `--smoke` silently overrides `--rounds/--warmup/--iterations` after they've been validated, and `raw["configuration"]` records the overridden values, so an operator passing `--smoke --rounds 3` gets 1 round with no hint. Either make them mutually exclusive in argparse, or use `default=None` and only apply the smoke defaults when the user didn't pass explicit values.

python/aisimulate/tools/forward_perf_gate/compare.py
```diff
@@ -0,0 +1,481 @@
...
+    if base_status == "OK" and head_status == "OK":
+        return "COMPARE", None
+    if base_status == "DATA_MISS" and head_status == "DATA_MISS":
+        details = list(dict.fromkeys(filter(None, (_error_text(base), _error_text(head)))))
+        suffix = f": {'; '.join(details)}" if details else ""
+        return "SKIP", f"DATA_MISS on base and head{suffix}"
+    if base_status == "DATA_MISS" and head_status == "OK":
+        detail = _error_text(base)
+        suffix = f": {detail}" if detail else ""
+        return "SKIP", f"base DATA_MISS{suffix}; head has no timing baseline"
```
jasonqinzhou, line 71: Could this skip path distinguish gained coverage from coverage that disappears after availability was established? `run.py` records a successful `OK/OK` prewarm pair in `raw["prewarm"]`, but `compare_raw()` never consults it. I reproduced prewarm `OK/OK` followed by all five measured rounds returning base `DATA_MISS` and head `OK`; both metrics were classified `SKIPPED` and the overall result had `blocking=false`. That contradicts the documented contract that a working base which stops working is an invalid, blocking comparison. Please carry the prewarm disposition into each case comparison (or otherwise preserve the established baseline state) and add this transition as a regression test.

python/aisimulate/tools/forward_perf_gate/measurement.py
```diff
@@ -0,0 +1,204 @@
...
+    runtime_config: config.RuntimeConfig,
+    *,
+    phase: str,
+) -> config.RuntimeConfig:
+    """Return one supported off-matrix query that initializes a prediction phase."""
+    return replace(
+        runtime_config,
+        batch_size=2,
+        isl=2048,
+        osl={"context": grid.CTX_OSL, "generation": grid.GEN_OSL}[phase],
```
jasonqinzhou: Could we decouple the priming OSL from `prediction_regression_gate.grid` here? The controller sends the same target cases to both revision-local workers, while each worker imports this helper from its own checkout. If a later change modifies `CTX_OSL` or `GEN_OSL`, base and head can prime different queries while their target case hashes still match, so a priming-state difference can appear as a performance change. Please make the complete priming coordinates stable within this gate, or carry them in the versioned request and hash them.

python/aisimulate/tools/forward_perf_gate/worker.py
```diff
@@ -0,0 +1,372 @@
...
+import platform
+import sys
+import traceback
+from dataclasses import replace
+from pathlib import Path
+
+PYTHON_ROOT = Path(__file__).resolve().parents[2]
+sys.path.insert(0, str(PYTHON_ROOT))
+
+from aiconfigurator.sdk.errors import (
```
jasonqinzhou: A fresh scoped `ruff check` fails `I001` in this file, `measurement.py`, and both new forward-gate test files. The current CI lint job does not scan these tool/test paths, so its green result does not cover these errors. Please apply Ruff's import sorting to the four files and include these paths in the validation command.

jasonqinzhou (changes requested):

## What this PR does

This PR adds a revision-local worker/controller harness and report pipeline for an advisory forward-prediction performance gate. The new head hardens worker failure isolation, broadens typed coverage-miss handling, preserves explicit smoke-run counts, and improves skip diagnostics.

The follow-up commit thoughtfully addresses the existing reviewer feedback with focused tests: empty exceptions and PyO3-style panics are isolated, missing database and modeling coverage use typed results, priming failures are phase-scoped, and smoke overrides are now explicit.

## Why this score

The 6.5 score reflects a substantially more resilient implementation whose 59 focused tests pass, but Codex confirmed a P1 fail-open comparison path: a base case that succeeds during prewarm and then loses coverage in every measured round is still reported as skipped and non-blocking. The earlier priming-protocol issue and four scoped Ruff import-order failures also remain. The canonical recommendation is BLOCKED rather than REQUEST_CHANGES because the mandatory helper requested claude-fable-5 but observed only claude-haiku-4-5-20251001 and claude-sonnet-4-6; the required independent Fable review is therefore unavailable.

jthomson04:
@jasonqinzhou Fixed all three comments in [c0771fa](https://github.com/ai-dynamo/aisimulate/commit/c0771facd181a407ad101a29c89e14744535bd0a).

- [Coverage loss after availability](https://github.com/ai-dynamo/aisimulate/pull/71#discussion_r3965163599): `compare_raw()` now passes each case's successful prewarm disposition into the comparison. A later `DATA_MISS` on either side makes both metrics `INVALID_COMPARISON` and sets `blocking=true`, with the round and failed side in the reason. The regression test covers base, head, or both sides losing coverage in one or all five rounds. Existing skip behavior remains for cases without established availability.
- [Stable priming coordinates](https://github.com/ai-dynamo/aisimulate/pull/71#discussion_r3965163606): priming now uses fixed protocol-v1 coordinates: batch size 2, ISL 2048, prefix 0, context OSL 8, and generation OSL 256. Stride still comes from the shared, hashed request. The test changes both grid OSL values and checks that priming retains its fixed coordinates. The README states that changes to these coordinates require a protocol-version change.
- [Import sorting](https://github.com/ai-dynamo/aisimulate/pull/71#discussion_r3965163608): fixed the import-order errors and checked the full gate directory and both test files explicitly with the project configuration:

```bash
ruff check --config python/aisimulate/pyproject.toml \
  python/aisimulate/tools/forward_perf_gate \
  python/aisimulate/tests/unit/tools/test_forward_perf_gate_compare.py \
  python/aisimulate/tests/unit/tools/test_forward_perf_gate_worker.py
```

Validation: 65 focused tests passed. The scoped Ruff check and format check passed, as did `git diff --check`. [CI on this commit](https://github.com/ai-dynamo/aisimulate/actions/runs/34440538246) passed all 15 jobs.

python/aisimulate/tools/forward_perf_gate/worker.py
jasonqinzhou: Verified fixed in [519f2876](https://github.com/ai-dynamo/aisimulate/commit/519f28768134c40e5450c790cdde38050e5c4b2a) at current head `c0771fac`. Error recording now uses str(exc).split("\n", 1)[0], so an empty exception message stays isolated. The focused test covers MemoryError() with an empty message.

The fresh focused suite passes all 65 tests. Resolving this finding.

python/aisimulate/tools/forward_perf_gate/worker.py
jasonqinzhou: Verified fixed in [519f2876](https://github.com/ai-dynamo/aisimulate/commit/519f28768134c40e5450c790cdde38050e5c4b2a) at current head `c0771fac`. The worker now classifies all four typed coverage errors, including chained causes, as DATA_MISS. Parameterized tests cover each sibling error type.

The fresh focused suite passes all 65 tests. Resolving this finding.

python/aisimulate/tools/forward_perf_gate/measurement.py
jasonqinzhou: Verified fixed in [519f2876](https://github.com/ai-dynamo/aisimulate/commit/519f28768134c40e5450c790cdde38050e5c4b2a) at current head `c0771fac`. A missing database now raises PerfDataNotAvailableError and follows the intended DATA_MISS policy. test_missing_database_is_a_data_miss covers the None return.

The fresh focused suite passes all 65 tests. Resolving this finding.

python/aisimulate/tools/forward_perf_gate/worker.py
jasonqinzhou: Verified fixed in [519f2876](https://github.com/ai-dynamo/aisimulate/commit/519f28768134c40e5450c790cdde38050e5c4b2a) at current head `c0771fac`. Priming failures are now recorded as PRIMING_FAILED for the affected phase only; the other phase continues. Tests cover both missing data and a synthetic panic.

The fresh focused suite passes all 65 tests. Resolving this finding.

python/aisimulate/tools/forward_perf_gate/worker.py
jasonqinzhou: Verified fixed in [519f2876](https://github.com/ai-dynamo/aisimulate/commit/519f28768134c40e5450c790cdde38050e5c4b2a) at current head `c0771fac`. Setup, priming, and case execution now catch BaseException while explicitly re-raising KeyboardInterrupt and SystemExit. Synthetic-panic tests verify setup and per-case isolation.

The fresh focused suite passes all 65 tests. Resolving this finding.

python/aisimulate/tools/forward_perf_gate/compare.py
jasonqinzhou: Verified fixed in [519f2876](https://github.com/ai-dynamo/aisimulate/commit/519f28768134c40e5450c790cdde38050e5c4b2a) at current head `c0771fac`. The subsequent skipped-reason branch is now an elif, so the status-change reason is emitted once. test_changed_skip_reason_is_reported_once verifies the result.

The fresh focused suite passes all 65 tests. Resolving this finding.

python/aisimulate/tools/forward_perf_gate/run.py
jasonqinzhou: Verified fixed in [519f2876](https://github.com/ai-dynamo/aisimulate/commit/519f28768134c40e5450c790cdde38050e5c4b2a) at current head `c0771fac`. Smoke defaults now apply only to unspecified counts. Explicit rounds, warmup, and iterations are preserved and recorded, with tests for both count selection and raw output.

The fresh focused suite passes all 65 tests. Resolving this finding.

python/aisimulate/tools/forward_perf_gate/compare.py
jasonqinzhou: Verified fixed in [c0771fac](https://github.com/ai-dynamo/aisimulate/commit/c0771facd181a407ad101a29c89e14744535bd0a) at current head `c0771fac`. compare_raw now carries successful availability into each case comparison. Later measured DATA_MISS results on either side are blocking INVALID_COMPARISON results. Six regression cases cover base, head, or both sides losing coverage in one or all five rounds.

The fresh focused suite passes all 65 tests. Resolving this finding.

python/aisimulate/tools/forward_perf_gate/measurement.py
jasonqinzhou: Verified fixed in [c0771fac](https://github.com/ai-dynamo/aisimulate/commit/c0771facd181a407ad101a29c89e14744535bd0a) at current head `c0771fac`. Priming uses fixed protocol-v1 coordinates: batch size 2, ISL 2048, prefix 0, context OSL 8, and generation OSL 256. The tests modify both grid OSL constants and verify that priming stays fixed; the README requires a protocol bump for coordinate changes.

The fresh focused suite passes all 65 tests. Resolving this finding.

python/aisimulate/tools/forward_perf_gate/worker.py
jasonqinzhou: Verified fixed in [c0771fac](https://github.com/ai-dynamo/aisimulate/commit/c0771facd181a407ad101a29c89e14744535bd0a) at current head `c0771fac`. The import-order findings are fixed. A fresh scoped Ruff check and format check pass for the entire forward_perf_gate directory and both new test files.

The fresh focused suite passes all 65 tests. Resolving this finding.

jasonqinzhou:
The fresh Fast CI run at `c0771fac` reaches the checks after adding `review-ready`, but current-main workflow steps reference two files absent from this older branch: `python/aisimulate/tests/e2e/cli/test_cli_recommend.py` and `tests/test_ci_workflow_contracts.py`. See [run 34636102108](https://github.com/ai-dynamo/aisimulate/actions/runs/34636102108).

I am updating this branch with current `main`, then validating the resulting exact head before approval and squash merge. The benchmark fixes remain intact; this update brings in the repository CI contracts and current integration base.

jasonqinzhou (approved):
Approved at `e80e4a03d9dcb6226f7bc69fca11dfddfc8894c1` after re-reviewing the fixes and integrating current `main` (`a4c77310`). The benchmark diff is unchanged by the integration update.

The availability-loss, priming-protocol, import-order, worker-isolation, and reporting findings are fixed. All 10 review threads have been verified and resolved.

Fresh validation: 65 focused benchmark tests passed; 29 CI workflow-contract tests passed; all 36 native matrix cases returned OK with finite predictions; scoped Ruff, formatting, and diff checks passed. [Fast CI](https://github.com/ai-dynamo/aisimulate/actions/runs/34636304303) and [Full CI](https://github.com/ai-dynamo/aisimulate/actions/runs/34636308762) passed on this exact head. Full CI finished with 31 successful jobs and the expected main-only staging skip. CODEOWNERS and DCO are green.

This approval supersedes my earlier change request; those findings no longer apply.

PR #74, opened 2026-09-01
feat(memory): accept explicit CUDA graph reservation

simone-chen:
## Summary

- Add an optional rank-local `cuda_graph_reserved_bytes` input to Python and Rust KV-cache estimation.
- Subtract the fixed reservation before backend-specific KV allocation.
- Define the SGLang value as additional to headroom already encoded by `mem_fraction_static`.
- Preserve capacity inputs through replay and native AIC rematerialization.
- Expose `engine.workers.<role>.kv_cache.capacity.cuda_graph_reserved_bytes` through the Python prediction config used by `aisimulate predict`.
- Keep serialized requests and public prediction configs compatible with a zero default; document the Rust source migration.

## Scope

This PR accepts a caller-provided value only. It does not estimate CUDA graph memory and does not include InfX parsing, profile data, Parquet files, or a prediction model.

## Validation

- `cargo test -p aisimulate-core`: 1,082 unit tests passed, 1 ignored; all 50 integration tests passed.
- Rebuilt the Python wheel from the PR head and ran `pytest tests -q -p no:timeout`: 454 passed, 5 expected skips.
- Deterministic embed-Python PyDict forwarding and parsing test passed without a perf database.
- Focused Python coverage validates the public range, rejects no-op fixed-capacity combinations, and traces YAML through Python and native AIC capacity materialization.
- `cargo fmt --all -- --check`, Ruff check, and Ruff format check passed.


python/aisimulate/src/aiconfigurator_core/sdk/memory.py
```diff
@@ -407,7 +423,8 @@ def _estimate_from_breakdown(
         if not (math.isfinite(capacity) and capacity > 0.0):
             raise ValueError(f"GPU capacity must be finite and positive, got {capacity}")
 
-        non_kv = float(breakdown["non_kv_bytes"])
+        cuda_graph = float(cuda_graph_reserved_bytes)
```
tedzhouhk, line 429: [P1] Avoid double-reserving CUDA graph memory for SGLang. `mem_fraction_static` is already derived from capacity after activation and CUDA-graph buffers are reserved (and falls back to 0.88), so `C * f - N - R` counts graph headroom twice. Add a SGLang-specific test and either replace the graph component embedded in `f`, or define this input explicitly as an *additional* reservation.

python/aisimulate/src/aisimulate/runner.py
```diff
@@ -577,6 +595,10 @@ def _materialize_engine_role(
         and timing_model.get("type") == "external"
         and timing_model.get("provider") == "aic"
     )
+    if cuda_graph_reserved_bytes is not None and uses_aic_timing:
```
tedzhouhk, line 593: [P1] Preserve the non-default memory fraction alongside this field. For flat inferred AIC config, Python can first estimate with `gpu_memory_utilization=0.8`, but `capacity_materialized` makes lines 563-575 drop that value. Rust then sees `num_gpu_blocks_is_explicit=false`, re-estimates with the default 0.9 while retaining only this reservation, and overwrites the block count. Carry every capacity input needed by native rematerialization.

python/aisimulate/src/aisimulate/runner.py
```diff
@@ -577,6 +595,10 @@ def _materialize_engine_role(
...
+        aic_timing_overrides["cuda_graph_reserved_bytes"] = (
+            cuda_graph_reserved_bytes
+        )
```
tedzhouhk, line 600: [P2] Do not silently discard this value for nested-rank configurations with fixed/polynomial timing. The presence of `rank` skips Python capacity materialization; `uses_aic_timing=false` then clears this override, and a missing `num_gpu_blocks` falls back to the Rust default (4096) as if no reservation were supplied. Materialize nested inferred capacity or reject this combination.

tedzhouhk (commented):
Agentic review of the CUDA graph reservation path. I found correctness issues in SGLang accounting and native rematerialization, plus replay, compatibility, and coverage gaps. Inline comments below.

crates/core/src/perfmodel/memory.rs
```diff
@@ -46,6 +46,9 @@ pub struct KvCacheEstimateRequest {
     /// Override for unknown SKUs; when `Some`, it wins over the SystemSpec
     /// capacity reported by the native path.
     pub gpu_memory_capacity_bytes_override: Option<u64>,
+    /// Fixed rank-local bytes reserved by CUDA graphs before KV allocation.
+    #[serde(default)]
+    pub cuda_graph_reserved_bytes: u64,
```
tedzhouhk, line 53: [P2] `#[serde(default)]` protects JSON deserialization, not Rust source compatibility. This is a supported public struct without a builder/`Default`, so every existing external struct literal now fails with E0063; the new public `MemoryBreakdown` field similarly breaks exhaustive literals/patterns. Treat this as a next-minor break with migration guidance, or provide a non-breaking construction API rather than claiming complete backward compatibility.

crates/core/src/perfmodel/memory.rs
```diff
@@ -126,6 +129,7 @@ pub struct MemoryBreakdown {
     pub activations_bytes: u64,
     pub runtime_overhead_bytes: u64,
     pub comm_overhead_bytes: u64,
+    pub cuda_graph_reserved_bytes: u64,
```
tedzhouhk, line 135: [P2] Add `#[serde(default)]` here as well. New code cannot deserialize any pre-PR `KvCacheEstimate` JSON whose native `memory_breakdown` lacks this field, even though the request side was made backward-compatible. Please add an old-JSON fixture test.

crates/core/tests/perfmodel/memory_round_trip.rs
```diff
@@ -164,6 +165,10 @@ fn memory_round_trip_forwards_tolerance_and_parses_adjusted() {
     assert_eq!(raw.source, EstimateSource::Native);
     assert!(raw.total_kv_size_bytes > 0 && raw.kv_size_per_token_bytes > 0);
     assert!(raw.memory_breakdown.is_some());
+    assert_eq!(
+        raw.memory_breakdown.unwrap().cuda_graph_reserved_bytes,
```
tedzhouhk, line 169: [P2] This new FFI assertion is not enforced by CI: the default Rust job does not enable `python`/`embed-python`, and this test returns before the assertion on its soft-skip path. More importantly, `fetch_python_estimate` maps every PyO3 call/dict-parse failure to `Unsupported`, which this test also skips, so a missing forwarded key would still appear green. Add a deterministic PyDict/fake-module test that cannot depend on the perf DB.

python/aisimulate/src/aiconfigurator_core/sdk/memory.py
```diff
@@ -127,6 +127,19 @@ def _validate_tolerance(tolerance_fraction: float | None) -> None:
         raise ValueError(f"tolerance_fraction must be finite and in [0, 1), got {tolerance_fraction}")
 
 
+def _validate_cuda_graph_reservation(cuda_graph_reserved_bytes: int) -> None:
+    """Validate the fixed, rank-local CUDA graph reservation."""
+    if (
+        isinstance(cuda_graph_reserved_bytes, bool)
+        or not isinstance(cuda_graph_reserved_bytes, int)
+        or cuda_graph_reserved_bytes < 0
```
tedzhouhk, line 136: [P3] Align this validator with the Rust `u64` contract and the later float conversion. Python currently accepts `1 << 64` but native serde rejects it; values above `2**53` also lose byte precision when converted to `float`, and sufficiently large integers raise an undeclared `OverflowError`. Validate the upper bound and preserve the original integer for output.

docs/core-api.md
```diff
@@ -54,6 +54,20 @@ extension contract:
 The wheel includes `py.typed` and a stub for that native extension. The SDK
 Python modules carry their own annotations.
 
+## KV-cache capacity reservation
+
+`estimate_kv_cache` and `estimate_num_gpu_blocks` accept
+`cuda_graph_reserved_bytes=<rank-local bytes>`. The value must be a
+non-negative integer and defaults to zero. It is treated as fixed non-KV memory
+before the backend-specific KV fraction is applied and is returned as
+`memory_breakdown.cuda_graph_reserved_bytes`.
```
tedzhouhk: [P2] Qualify this as native-only or change the result contract. With `allow_naive_fallback=True`, the reservation is applied but `memory_breakdown` remains `None`, so callers following this documentation cannot read `memory_breakdown.cuda_graph_reserved_bytes`. Add a fallback-path test for whichever contract is intended.

python/aisimulate/src/aiconfigurator_core/sdk/memory.py
```diff
@@ -861,7 +885,9 @@ def estimate(self, *, capacity: int | None, naive_kv_reservation: float) -> dict
         kv_per_token = float(kv_per_token)
         weight_bytes = float(weight_bytes)
 
-        post_weight = max(capacity_bytes - weight_bytes, 0.0)
+        post_weight = max(
+            capacity_bytes - weight_bytes - float(cuda_graph_reserved_bytes), 0.0
+        )
```
tedzhouhk: [P3] Include `cuda_graph_reserved_bytes` in the no-budget diagnostic below. The calculation subtracts weights plus this reservation, but the error reports only `weight_bytes` as non-KV memory, producing a numerically false explanation whenever the graph reservation causes exhaustion.

python/aisimulate/src/aiconfigurator_core/sdk/memory.py
simone-chen: Addressed in 3421b60f. SGLang now explicitly treats this value as an additional reservation beyond headroom already encoded in mem_fraction_static, with SGLang-specific budget coverage.

python/aisimulate/src/aisimulate/runner.py
simone-chen: Addressed in 3421b60f. Inferred-capacity memory fractions are now retained in the AIC timing config so Rust rematerialization uses the same non-default value. The replay regression covers gpu_memory_utilization=0.8.

crates/core/src/perfmodel/memory.rs
simone-chen: Addressed in 3421b60f. The docs now distinguish serde wire compatibility from Rust source compatibility and give migration guidance for direct request literals and exhaustive breakdown literals or patterns as a next-minor API change.

python/aisimulate/src/aiconfigurator_core/sdk/memory.py
simone-chen: Addressed in 3421b60f. Validation now caps the value at 2**53 for exact float conversion, rejects larger values including u64 overflow cases, and returns the original integer in the breakdown.

crates/core/src/perfmodel/memory.rs
simone-chen: Addressed in 3421b60f. MemoryBreakdown now defaults the new field during deserialization, and a pre-change estimate JSON fixture verifies it reads as zero.

docs/core-api.md
simone-chen: Addressed in 3421b60f. The contract now says the breakdown field is native-only; fallback still applies the reservation and returns memory_breakdown=None. A fallback regression verifies both behaviors.

python/aisimulate/src/aisimulate/runner.py
simone-chen: Addressed in 3421b60f. The reservation is extracted from nested rank input, and nested inferred capacity with fixed or polynomial timing is rejected instead of silently falling back. A replay regression covers this case.

crates/core/tests/perfmodel/memory_round_trip.rs
simone-chen: Addressed in 3421b60f. Keyword construction is now isolated and covered by a deterministic PyDict round trip that verifies both forwarding and parsing without importing AIC or requiring a perf database.

python/aisimulate/src/aiconfigurator_core/sdk/memory.py
simone-chen: Addressed in 3421b60f. The naive no-budget diagnostic now reports weights plus the CUDA graph reservation as non-KV memory.

dreamtalen (commented):
Logic looks good to me. We also need to expose cuda_graph_reserved_bytes through the Python config so the standard aisimulate predict CLI can use this new field.

simone-chen:
Addressed the Python config gap in `93ad206b`:

- added `engine.workers.<role>.kv_cache.capacity.cuda_graph_reserved_bytes` to the strict prediction schema;
- validated it as an integer from `0` through `2**53`, defaulting to `0`;
- lowered it through `aisimulate predict` into Python and native AIC KV-capacity materialization;
- documented the YAML path and added range, fixed-capacity, and YAML-to-runtime tests.

Validation: rebuilt the wheel from the PR head; 454 Python tests passed with 5 expected skips; Ruff checks pass.

dreamtalen (approved):

PR #75, opened 2026-09-01
ci(aic): add advisory forward prediction gate

jthomson04:
> Superseded by #189, which carries this workflow unchanged onto `main`. Follow the replacement PR for fresh CI and review; restoring the deleted base is no longer needed.

## Summary

Adds `.github/workflows/performance.yml`, an advisory forward-prediction performance check. The benchmark harness from #71 is now merged into `main` at `d1f4a474288525ab99890477c9e5b42ddb4e481e`.

- Run for relevant source changes on canonical `pull-request/<number>` branches or by manual dispatch.
- Benchmark the trusted canonical commit against the merge base of the PR's base branch.
- Use the base revision's controller, thresholds, and case matrix, with separate revision-local workers, wheels, virtual environments, and source trees.
- Skip with an explanatory summary when the benchmark protocol is missing or incompatible.
- Share a concurrency group between push and manual runs for the same PR.
- Use run-specific worktrees, always clean up, and publish a summary, aggregate blocker annotation, and full artifacts.

The check is advisory and is not required by branch protection. Velonix admits canonical pushes; manual dispatch resolves the same canonical branch for the requested PR number.

## Scope decisions

Keep native-build setup in this workflow, `util-linux` as the explicit source of `taskset`, and the two release builds sequential and uncached. Build optimization remains separate.

## Validation

Reviewed head: `c453ebe7bf09d4715744841c42fcef1b8ee9d8b4`.

- [Full CI run 34440584102](https://github.com/ai-dynamo/aisimulate/actions/runs/34440584102): passed on this head against the original stacked base.
- [Hosted advisory run 34440584190](https://github.com/ai-dynamo/aisimulate/actions/runs/34440584190): passed; compared base `c0771facd181a407ad101a29c89e14744535bd0a` with head `c453ebe7bf09d4715744841c42fcef1b8ee9d8b4`.
- Hosted benchmark: 36 cases, five paired rounds, 72 comparisons; 44 stable (20 faster, 24 unchanged), 28 noisy, zero confirmed regressions, invalid comparisons, skipped comparisons, or run errors.
- [Benchmark artifact](https://github.com/ai-dynamo/aisimulate/actions/runs/34440584190/artifacts/10138009171).
- Fresh local integration check on 2026-09-11: clean merge with `main` at `d1f4a474288525ab99890477c9e5b42ddb4e481e`; the entire resulting change is this one 267-line workflow file.
- On that local integration: actionlint 1.7.12, YAML parsing, all eight embedded shell blocks (`bash -n`), and whitespace checks passed; 29 existing CI workflow contract tests passed.
- All four review threads were verified fixed by `c453ebe7` and resolved.

The hosted runs above predate integration with current `main`. Fresh Full CI and an advisory benchmark are still needed after updating the remote branch.

## Historical recovery blocker

After #71 merged, `copy-pr-bot` deleted its canonical `pull-request/71` branch and automatically closed this stacked PR. This PR is **closed, not merged**.

GitHub refuses to reopen the PR while its base is missing, and both REST and GraphQL refuse to retarget a closed PR. The organization's active `copy-pr-bot` ruleset 4025561 restricts creation and updates under `refs/heads/pull-request/**`; manual restoration was rejected.

A bot/operator with permission under that ruleset must restore `refs/heads/pull-request/71` at its last validated head, `e80e4a03d9dcb6226f7bc69fca11dfddfc8894c1`. Then reopen #75, retarget it to `main`, update its branch, complete fresh validation, and perform the normal protected merge. No protections have been changed.


tianhaox (commented):
The workflow itself looks solid: controller and thresholds from the base revision, canonical `pull-request/<n>` commit only, `persist-credentials: false`, run-scoped worktrees with `always()` cleanup. I cross-checked the CLI flags, output file names and exit codes against #71 and they match. Two things worth addressing, plus one nit.

.github/workflows/performance.yml
```diff
@@ -0,0 +1,218 @@
...
+          if [[ "${actual_head}" != "${PR_HEAD_SHA}" ]]; then
+            echo "checked out ${actual_head}, expected PR head ${PR_HEAD_SHA}" >&2
+            exit 1
+          fi
+          base_tip="$(git rev-parse "refs/remotes/origin/${BASE_REF}")"
+          base_sha="$(git merge-base "${PR_HEAD_SHA}" "${base_tip}")"
+
+          gate_path="python/aisimulate/tools/forward_perf_gate"
+          if ! git cat-file -e "${base_sha}:${gate_path}/run.py" || \
+             ! git cat-file -e "${base_sha}:${gate_path}/worker.py"; then
```
tianhaox: The skip check only tests that `run.py` / `worker.py` exist at the merge base. Any PR that bumps `PROTOCOL_VERSION` (1 → 2) will still run, and the base-side `compare.py` will flag a protocol mismatch → INVALID → red advisory check. Since the protocol is the one shared interface by design, this will happen by construction on every protocol change.

Suggest also comparing `PROTOCOL_VERSION` between `${base_sha}` and `${PR_HEAD_SHA}` (e.g. `git show <sha>:${gate_path}/__init__.py | grep PROTOCOL_VERSION`) and taking the SKIPPED path with an explanatory summary when they differ. Alternatively, document in the README that protocol-bump PRs are expected to go red.

.github/workflows/performance.yml
```diff
@@ -0,0 +1,218 @@
...
+      - "Cargo.toml"
+      - "Cargo.lock"
+      - "crates/**"
+      - "python/aisimulate/**"
+
+permissions:
+  contents: read
+  pull-requests: read
+
+concurrency:
```
tianhaox, line 45: The concurrency key differs between the two triggers for the same PR: `workflow_dispatch` yields `forward-prediction-performance-71`, while a push yields `forward-prediction-performance-refs/heads/pull-request/71`. A manual run and an automatic run for the same PR can therefore execute concurrently, and if they land on the same node they'll both pin to the same CPU and skew each other's timings. Suggest normalising to the PR number, e.g. `${{ inputs.pr_number || github.ref_name }}` with the `pull-request/` prefix stripped in the group expression, or just keying on `github.ref_name` and mapping dispatch to `pull-request/<n>`.

.github/workflows/performance.yml
```diff
@@ -0,0 +1,218 @@
...
+      - name: Remove benchmark worktrees
+        if: always()
+        run: |
+          if [[ -n "${BASE_SRC:-}" ]]; then
+            git worktree remove --force "${BASE_SRC}" || true
+          fi
+          if [[ -n "${HEAD_SRC:-}" ]]; then
+            git worktree remove --force "${HEAD_SRC}" || true
+          fi
+          git worktree prune
```
tianhaox: Nit: the two `git worktree remove` calls are guarded with `|| true` but the final `git worktree prune` isn't. If the checkout step failed there's no repo here and this `always()` step will report a second, unrelated failure. Adding `|| true` keeps the cleanup step quiet.

tianhaox (commented):
One follow-up on the trigger scope.

.github/workflows/performance.yml
```diff
@@ -0,0 +1,218 @@
...
+    inputs:
+      pr_number:
+        description: Pull request number to compare
+        required: true
+        type: number
+  push:
+    # Velonix admits canonical pushes, not untrusted pull_request events.
+    branches:
+      - "pull-request/*"
+    paths:
```
tianhaox, line 17: Suggestion: the `paths` filter is much broader than what the benchmark actually exercises. `crates/**` pulls in `crates/core/src/replay` and `crates/tests`; `python/aisimulate/**` pulls in `tests/` (~280 files), `collector/` (~200), `docs/`, the CLI/replay/sweeper code under `src/aisimulate/`, the `src/aiconfigurator/` facade, and perf data for 13 GPUs other than `b200_sxm`. Each false trigger costs ~9 min of runner time, almost all of it the two uncached release builds.

The measured path (`run_static_latency_only` → Rust engine step) depends on a fairly small set of inputs. Something like this would cover it:

```yaml
paths:
  - ".github/workflows/performance.yml"
  - "Cargo.toml"
  - "Cargo.lock"
  - "crates/core/Cargo.toml"
  - "crates/core/src/lib.rs"
  - "crates/core/src/python.rs"
  - "crates/core/src/engine/**"
  - "crates/core/src/perfmodel/**"
  - "python/aisimulate/pyproject.toml"
  - "python/aisimulate/src/aiconfigurator_core/sdk/**"
  - "python/aisimulate/src/aiconfigurator_core/model_configs/Qwen--Qwen3-32B_config.json"
  - "python/aisimulate/src/aiconfigurator_core/model_configs/Qwen--Qwen3-235B-A22B_config.json"
  - "python/aisimulate/src/aiconfigurator_core/systems/*.yaml"
  - "python/aisimulate/src/aiconfigurator_core/systems/support_matrix/**"
  - "python/aisimulate/src/aiconfigurator_core/systems/fpe_support_matrix/**"
  - "python/aisimulate/src/aiconfigurator_core/systems/data/b200_sxm/**"
  - "python/aisimulate/tools/forward_perf_gate/**"
  - "python/aisimulate/tools/prediction_regression_gate/grid.py"
```

Since the check is advisory, a missed trigger only means one fewer run, whereas a false trigger burns a runner for 9 minutes, so erring on the narrow side seems right. The main maintenance cost is keeping this list in sync if shared code moves (e.g. something factored out of `engine/` or `perfmodel/` into a new module); a one-line note in the harness README would cover that. Build caching (which you've already listed as follow-up) would make this less pressing, and the two are complementary.

jasonqinzhou (approved):
**APPROVE · 9.5/10 · high confidence**

## What this PR does

This PR adds a root advisory workflow that resolves a pull request's trusted canonical head and merge base, builds each revision in isolated worktrees, wheels, and virtual environments, runs the 36-case paired forward-prediction benchmark, and publishes a fail-closed comparison summary plus artifacts. The new head rebases onto the hardened benchmark protocol and refines protocol compatibility, trigger scope, concurrency isolation, and cleanup.

The follow-up is unusually thorough: it incorporates the updated worker's phase-local failure handling, normalizes manual and push concurrency keys, skips incompatible protocol revisions with an explicit reason, narrows expensive runs to the measured dependency surface, and validates the exact rebased head with a complete hosted benchmark.

## Why this score

The independently frozen Codex pass scores the current head 9.5/10 and recommends approval: every previous workflow concern touched by the delta is fixed, no actionable PR #75 defect was found, focused checks passed, and the exact-head hosted benchmark completed cleanly. The canonical recommendation is nevertheless BLOCKED because the helper requested claude-fable-5 but observed only claude-sonnet-4-6 and claude-haiku-4-5-20251001, so the required independent Fable lane is unavailable and this packet is not submittable.

.github/workflows/performance.yml
jasonqinzhou: Verified fixed in https://github.com/ai-dynamo/aisimulate/commit/c453ebe7bf09d4715744841c42fcef1b8ee9d8b4. The preparation step checks the gate files on both revisions, extracts both PROTOCOL_VERSION values, and skips with an explanatory summary when either value is unavailable or the versions differ. The comparison step is gated on run_comparison=true.

.github/workflows/performance.yml
jasonqinzhou: Verified fixed in https://github.com/ai-dynamo/aisimulate/commit/c453ebe7bf09d4715744841c42fcef1b8ee9d8b4. Push runs use github.ref_name (pull-request/<n>), and manual runs map the supplied PR number to the same pull-request/<n> concurrency group. Both triggers now share cancellation for the same PR.

.github/workflows/performance.yml
jasonqinzhou: Verified fixed in https://github.com/ai-dynamo/aisimulate/commit/c453ebe7bf09d4715744841c42fcef1b8ee9d8b4. The always-run cleanup now guards the final git worktree prune with || true, matching both worktree removals.

.github/workflows/performance.yml
jasonqinzhou: Verified fixed in https://github.com/ai-dynamo/aisimulate/commit/c453ebe7bf09d4715744841c42fcef1b8ee9d8b4. The paths filter now names the measured Rust engine/performance-model code, Python facade/core/SDK dependencies, selected Qwen configs, B200 data, and gate tooling instead of the whole crates and Python trees. An inline maintenance note calls out keeping the filter aligned with the measured call path.

jasonqinzhou:
Closeout update: all four review threads are now resolved after checking their fixes at `c453ebe7`. The workflow integrates cleanly with current `main` (`d1f4a474`); actionlint, YAML parsing, all eight shell syntax checks, and 29 CI workflow contract tests pass locally.

The remaining blocker is recovery of this PR: merging #71 caused `copy-pr-bot` to delete `pull-request/71` and close #75. GitHub rejects reopening with the base missing, and rejects changing the base of a closed PR. Organization ruleset 4025561 also rejects manual restoration of the bot-managed branch.

Recovery requires a bot/operator with permission under that ruleset to restore `refs/heads/pull-request/71` at `e80e4a03d9dcb6226f7bc69fca11dfddfc8894c1`. The next steps are then to reopen #75, target `main`, update the branch, run fresh Full CI and the advisory benchmark, and merge with the normal protections. This PR has not been merged. The description now records the verified validation and recovery details.

PR #80, opened 2026-09-02
feat(moe): model DeepEP low-latency decode latency

Harrilee:
## Overview

Implements Stage 1 latency modeling for **DeepEP low-latency decode only** in AISimulate. DeepEP-HT, DeepEP V2, and TensorRT-LLM communication paths retain their existing behavior.

This is the AISimulate migration of the work reviewed in deprecated aiconfigurator PR [#1592](https://github.com/ai-dynamo/aiconfigurator/pull/1592), following the communication-data design introduced by [#1542](https://github.com/ai-dynamo/aiconfigurator/pull/1542).

The decode block is modeled as:

`FP8 DeepEP-LL dispatch -> measured DeepEP expert compute -> BF16 DeepEP-LL combine`

## Modeling boundary

Stage 1 deliberately separates measured local expert compute from modeled LL communication:

- The existing `MoeExpertCompute` predictor owns local expert-compute latency using the WideEP measurements, including the SGLang `DeepEPMoE.run_moe_core` path.
- The new Monte Carlo model applies **only** to dispatch/combine endpoint imbalance. It is not applied a second time to expert compute.
- DeepEP-LL uses the same measured expert-compute contract as the existing large-EP backends in both context and generation phases.
- The performance cost of the deterministic donor Monte Carlo path is tracked separately in [#104](https://github.com/ai-dynamo/aisimulate/issues/104) and does not block Stage 1.

## Calibration and dtype resolution

- Resolves the complete `(phase, dtype, H, K, N, EP, node topology)` shape, trying exact-topology requested/alias/compatible-default rows before single-domain donors in the same dtype order.
- Treats legacy `default` as phase-semantic, not a wildcard: dispatch maps to FP8 and combine maps to BF16. `fp8_block -> fp8` remains an alias.
- Fits valid multi-point curves with unweighted OLS `y = ax + b`, using `t0 = max(b, 0)`; a valid one-point curve borrows the median system-level `t0` and derives a one-shot slope.
- Distinguishes exact/donor and OLS/one-shot provenance. All modeled LL communication is `Estimated`; donor and one-shot paths emit fallback diagnostics.

## Monte Carlo and topology model

- Runs 4,096 deterministic trials for power-law workloads, including `power_law_1.0`, using a fixed `ChaCha8Rng` seed and a bounded cache.
- Preserves exact sampled expert quotas while randomly assigning Top-K distinct experts to source tokens.
- Uses the P50 of complete trial latency. Explicit `balanced`/`uniform` routing keeps every destination rank balanced and reproduces an exact-topology measurement.
- Models `P^2` logical routes as pressure on `P` physical endpoints, with full-duplex `max(TX, RX)` and parallel NVLink/IB paths.
- Exact topology uses measured calibration plus routing skew; a node-1 donor additionally applies the topology-spec bandwidth floor.
- Uses FP8 dispatch payload `H + 4H/128 + 16` and BF16 combine payload `2H`.

## Coverage and errors

- Keeps Python coverage and Rust runtime aligned through shared dtype-order and calibration-viability fixtures.
- Requires measured `MoeExpertCompute` coverage for DeepEP-LL, consistent with the other large-EP backends.
- Rejects DeepEP-LL + EPLB in Stage 1.
- Returns `EmpiricalNotImplemented` after all LL calibration candidates are exhausted in HYBRID mode; SILICON keeps its typed data-miss behavior.
- Restores legacy node-1 LL coordinates using physical node width (EP4 for GB200/GB300 and EP8 for HGX), without changing the HT adapter.

## Numerical parity coverage

Both hardware scenarios are independently pinned across static context, static generation, mixed, and decode surfaces:

| Scenario | Static context (ms) | Static generation (ms) | Mixed (ms) | Decode (ms) |
|---|---:|---:|---:|---:|
| H200 DeepSeek-V3 TP8/EP8, SGLang 0.5.14 | 100.6222 | 30.9112 | 104.0365 | 11.3923 |
| GB200 DeepSeek-R1 ADP32/EP32, SGLang 0.5.14 | 159.2335 | 50.0459 | 166.0868 | 17.0894 |

The graph assertions verify that both context and generation retain measured `MoeExpertCompute`; LL generation surrounds it with FP8 dispatch and BF16 combine.

## Holdout visualizer (NVIDIA Internal Only)

The temporary interactive report is available at [DeepEP-LL single-node -> multi-node holdout validation](https://gitlab-pages-72a7bf.gitlab-master-pages.nvidia.com/deepep-ll-single-node-scaling-validation.html).

**The temporary GitLab Pages link is NVIDIA Internal only** and requires NVIDIA network/access. The static screenshot below is uploaded directly to this PR and does not carry that internal-only restriction.

DeepEP-LL holdout validation visualizer

This is intentionally a **collector-only topology holdout**. It uses node-1 communication measurements to predict held-out node-2/4/8 collector curves and does not introduce an LLM-specific power law, expert hit-rate distribution, or routing trace.

- The node-1 collector donor plus topology bandwidth scaling reduces MAPE from 61.8% (unscaled) to 24.9%.
- This is a 59.7% relative MAPE reduction across 78 held-out dispatch/combine measurements.

The report validates the communication-topology extrapolation only; it does not claim to validate the separate runtime Monte Carlo workload amplification. Exact-topology measurements remain the first choice, donor predictions remain `Estimated`, and Stage 2/3 should validate workload skew separately with real routing or expert hit-rate data.

## Non-blocking follow-up

Tracked in [#104](https://github.com/ai-dynamo/aisimulate/issues/104): benchmark optimized cold/warm donor Monte Carlo queries and realistic `predict`/`recommend` candidate sets, then tune the implementation if it misses the agreed performance budget.

This performance follow-up does not block the Stage 1 accuracy boundary: local expert compute remains measurement-backed, and only LL communication uses the deterministic Monte Carlo model.

## Documentation

`python/aisimulate/docs/DEEPEP_LL_MODELING.md` documents token conventions, endpoint aggregation, calibration, topology, final equations, the measured-compute/communication boundary, fixed Stage 1 assumptions, and Stage 2/3 follow-ups. Module-level Rust documentation links directly to it.

## Validation

- Python unit suite: 4,930 passed, 11 skipped, 201 deselected (excluding the documented torch-only dispatch test).
- DeepEP-LL Python coverage/model-graph/runtime suite: 223 passed.
- Compile-engine parity: 67 passed.
- Rust core library: 1,123 passed, 1 ignored.
- Changed-file Ruff lint and format checks, Rust format, and diff checks pass.
- Golden predictions were regenerated from committed AISimulate code and carry commit `d8f0a17a6f7015bf7043839d57b6689a07170eed` in `post_freeze_pins`.

## Review entry points

- `python/aisimulate/docs/DEEPEP_LL_MODELING.md`
- `crates/core/src/perfmodel/operators/moe_ll_monte_carlo.rs`
- `crates/core/src/perfmodel/perf_database/moe_a2a.rs`
- `crates/core/src/perfmodel/operators/moe_a2a.rs`
- `python/aisimulate/src/aiconfigurator_core/sdk/models/blocks/moe.py`


python/aisimulate/src/aiconfigurator_core/sdk/models/blocks/moe.py
```diff
@@ -616,24 +630,47 @@ def _large_ep_block_ops(
...
-            num_experts=shape.num_experts,
-            moe_ep_size=cfg.moe_ep_size,
-            quant_mode=quant_mode,
-            workload_distribution=workload_distribution,
-            attention_dp_size=cfg.attention_dp_size,
-            inference_phase=inference_phase,
-            num_slots=cfg.wideep_num_slots or None,
-            is_gated=shape.is_gated,
-            enable_eplb=cfg.enable_eplb and is_deepep,
+    if comm_backend == "deepep_ll":
```
Arsene12358: **Blocker.** This branch prices `deepep_ll` decode compute from the ordinary `moe_perf` table (`sglang_fused_moe_triton` on H200, `sglang_flashinfer_trtllm_moe` on B200/GB200) instead of the `deepepmoe` table collected from `DeepEPMoE.run_moe_core`, the layer SGLang actually runs in DeepEP mode. Same token convention, different kernel: rows at identical coordinates are 1.1x to 9.6x slower, and decode step at batch 64 moves 1.5x to 3.3x on shipped H200/GB200/B200 data (details and A/B tables in the review body). Please drop the `deepep_ll` branch and keep `ops.MoEExpertCompute` here for every backend, as on `main`, together with the coverage gating in `moe_comm_resolver.py` / `task_v2.py`.

crates/core/parity_tests/perfmodel/test_compile_engine_parity.py
```diff
@@ -540,35 +540,45 @@ def test_mixed_and_decode_scales_thread_through(self) -> None:
 # --------------------------------------------------------------------------- #
 
 
-_WIDEEP_SGLANG_MODEL = "deepseek-ai/DeepSeek-V3"
-_WIDEEP_SGLANG_SYSTEM = "h200_sxm"
+_WIDEEP_SGLANG_MODEL = "deepseek-ai/DeepSeek-R1"
+_WIDEEP_SGLANG_SYSTEM = "gb200"
```
Arsene12358: Replacing the H200 DeepSeek-V3 tp8/ep8 scenario (and re-pinning `wideep_sglang::*` at this PR's commit) is what kept the compute-leg change out of the parity signal. Please keep the H200 numeric scenario alongside the new GB200 one and re-pin it; once `MoEExpertCompute` is restored the re-pin should show only the intended comm-leg drift.

crates/core/parity_tests/perfmodel/test_compile_engine_parity.py
```diff
@@ -591,23 +601,73 @@ def _build_wideep_sglang():
...
+    )
+    fused_moe_eps = database.moe_compute_coverage(
+        hidden_size=7168,
+        inter_size=2048,
+        topk=8,
+        num_experts=256,
+        quant_mode=common.MoEQuantMode.fp8_block,
+        workload_distribution="power_law_1.01",
+    )
+    assert 8 not in fused_moe_eps, (
```
Arsene12358: This asserts the absence of ordinary fused-MoE EP8 rows on 0.5.6.post2 and treats it as the reason H200 cannot produce an LL number. On `main`, H200 sglang 0.5.14 evaluates DeepSeek-V3 tp8/ep8 LL decode end to end from measured `deepepmoe` rows (static gen 30.43 ms). With the compute op restored, this shape-only test can go back to a numeric one.

Arsene12358 (changes requested):
## Request changes: the DeepEP-LL decode compute leg is now priced from a non-DeepEP kernel table

The LL communication model itself checks out: calibration order, phase-semantic `default`, the GB200/GB300 legacy EP4 fix, EPLB pruning, HYBRID semantics and the deterministic Monte Carlo all read correctly, and `cargo test --workspace` (1088), the Python unit/build suite (5063) and the compile-engine parity suite (65) pass locally on `bf28ccb6`. One change rides along that is not part of the stated design and it blocks the merge as is.

### Blocker: `MoEExpertCompute` -> ordinary `MoE` for `deepep_ll` compute

`_large_ep_block_ops` now emits `ops.MoE` instead of `ops.MoEExpertCompute` when `comm_backend == "deepep_ll"`, and the coverage gating follows it (`PerfDatabase.moe_compute_coverage`, `ll_compute_eps` in `moe_comm_resolver.py` and `task_v2.py`). The two ops read tables collected from different kernels. `wideep_generation_moe_perf.parquet` rows carry `kernel_source = deepepmoe` and are collected from `DeepEPMoE.run_moe_core` (`collector/wideep/sglang/collect_deepep_moe.py`), which is the layer SGLang runs under `--moe-a2a-backend deepep`. The ordinary `moe_perf.parquet` rows for the DeepSeek shape carry `kernel_source = sglang_fused_moe_triton` on H200 and `sglang_flashinfer_trtllm_moe` on B200/GB200, i.e. the non-DeepEP fused path. Both collectors use the same token convention (global tokens = per-rank tokens x EP, rank-0 worst load via the shared power-law helper), so rows at identical coordinates are directly comparable, and the ordinary table is a slower kernel at every large-EP coordinate:

| System / EP | num_tokens (global) | ordinary `moe_perf` (ms) | `deepepmoe` (ms) | ratio |
|---|---:|---:|---:|---:|
| H200 EP64 | 1024 | 0.334 | 0.089 | 3.74 |
| H200 EP64 | 8192 | 2.139 | 0.336 | 6.37 |
| GB200 EP32 | 32 | 0.077 | 0.068 | 1.12 |
| GB200 EP32 | 4096 | 0.917 | 0.159 | 5.77 |
| GB200 EP64 | 8192 | 1.594 | 0.180 | 8.84 |
| B200 EP64 | 8192 | 1.717 | 0.179 | 9.62 |

(fp8_block, power_law_1.01, tp=1, H=7168 I=2048 K=8 N=256; sglang 0.5.14 `moe_perf` vs 0.5.10 `wideep_generation_moe_perf`.)

End to end, evaluated on the same shipped data with a `main` worktree next to this branch (DeepSeek-R1/V3, sglang 0.5.14, `context=deepep_ht`, `generation=deepep_ll`, fp8_block; whole-model ms, main -> PR):

| Scenario | LL comm source | static gen (b1) | decode step b64 | compute op at 256 tok/rank |
|---|---|---:|---:|---:|
| H200 DSV3 tp8 ep8 | exact (8,1) | 30.43 -> 42.81 | 31.98 -> 33.35 | 21.32 -> 23.68 |
| H200 DSR1 dp16 ep16 | exact (16,2) | 70.03 -> 65.76 | 37.37 -> 57.43 (1.54x) | 16.65 -> 85.81 (5.2x) |
| H200 DSR1 dp64 ep64 | exact (64,8) | 48.01 -> 57.40 | 39.11 -> 129.74 (3.32x) | 40.98 -> 269.72 (6.6x) |
| GB200 DSR1 dp32 ep32 | donor (4,1) | 49.65 -> 51.56 | 19.87 -> 40.73 (2.05x) | 10.24 -> 100.93 (9.9x) |
| B200 DSR1 dp16 ep16 | donor (8,1) | 62.99 -> 59.56 | 23.04 -> 39.35 (1.71x) | 11.37 -> 65.52 (5.8x) |

Static context latency is identical to the last digit in every scenario, so the HT path is untouched as claimed. The dispatch/combine movement is the intended new comm model; the compute movement is the kernel swap. It also means the same `DeepEPMoE` layer is now priced from the `deepepmoe` table in context and from the fused-MoE table in generation, and the H200 EP16/32/64 configurations that had a fully measured chain on `main` (exact comm rows plus matching `deepepmoe` rows) lose the measured compute leg.

This did not show up as parity drift because `test_compile_engine_parity.py` replaced the H200 DeepSeek-V3 tp8/ep8 numeric scenario with a GB200 DeepSeek-R1 dp32/ep32 one and re-pinned the `wideep_sglang::*` goldens at `74d16b76`; the retained H200 test asserts graph shape only and pins 0.5.6.post2, which has no ordinary `moe_perf`, so the compute leg is never evaluated. At batch 1 the new golden moves only 4%; the batch-64 doubling has no golden. The PR description covers this with "uses the standard fused MoE predictor for LL expert compute" and the holdout report validates communication only.

### Suggested fix

Keep the DeepEP-mode compute table, exactly as `main` does for every backend:

```diff
-    if comm_backend == "deepep_ll":
-        # LL decode uses the ordinary fused-MoE predictor at the model's
-        # existing power-law curve. ...
-        routed_ops.append(
-            ops.MoE(
-                f"{prefix}_moe", scale_factor,
-                shape.hidden_size, shape.moe_inter_size, shape.topk, shape.num_experts,
-                cfg.moe_tp_size, cfg.moe_ep_size, quant_mode, workload_distribution,
-                cfg.attention_dp_size, is_context=is_context, is_gated=shape.is_gated,
-            )
-        )
-    else:
-        routed_ops.append(
-            ops.MoEExpertCompute(
-                ...
-            )
-        )
+    routed_ops.append(
+        ops.MoEExpertCompute(
+            f"{prefix}_moe",
+            scale_factor,
+            hidden_size=shape.hidden_size,
+            inter_size=shape.moe_inter_size,
+            topk=shape.topk,
+            num_experts=shape.num_experts,
+            moe_ep_size=cfg.moe_ep_size,
+            quant_mode=quant_mode,
+            workload_distribution=workload_distribution,
+            attention_dp_size=cfg.attention_dp_size,
+            inference_phase=inference_phase,
+            num_slots=cfg.wideep_num_slots or None,
+            is_gated=shape.is_gated,
+            enable_eplb=cfg.enable_eplb and is_deepep,
+        )
+    )
```

and the pieces that follow from it:

1. Remove `PerfDatabase.moe_compute_coverage`, `deepep_ll_workload_distribution`, and the `ll_compute_eps` / `ll_compute` special cases in `moe_comm_resolver.py` and `task_v2.py`, so LL compute gating goes back through `moe_expert_compute_coverage` like HT.
2. Restore the `MoEExpertCompute` assertions in `test_large_ep_model_graphs.py`, `test_moe_block_builder_large_ep.py` and `test_g2_data_gated.py` (the `moe_perf.parquet` fixture rows and shipped-data guards can stay).
3. Rewrite section 11 of `DEEPEP_LL_MODELING.md` to say expert compute is the measured DeepEP-mode `moe_expert_compute` table at the model's distribution; the "no second Monte Carlo scale-up" statement still holds.
4. Keep the H200 DeepSeek-V3 tp8/ep8 numeric parity scenario alongside the new GB200 one and re-pin it. With compute restored, that re-pin will show only the comm-leg change (static gen 30.43 -> about 36 ms), which is the drift this PR intends.

If there is evidence that the `deepepmoe` table is wrong, that belongs in the PR as data plus a data fix, not as a swap to a kernel SGLang does not run in this mode. The wide-EP table's known gaps (no EP8 on B200/B300/GB200/GB300, tokens capped at 128 per rank) were equally absent on `main`, so restoring `MoEExpertCompute` regresses nothing relative to `main`; they are collector follow-ups.

Non-blocking observations (donor-mode Monte Carlo cost, skew-factor validation, one-shot provenance wording) are in my full report and I can share them separately.

python/aisimulate/src/aiconfigurator_core/sdk/models/blocks/moe.py
Harrilee: Thanks for tracing the two table contracts and quantifying the kernel delta. The numerical difference is real, but the standard `MoE` selection is intentional in the agreed Stage 1 design, so I have not restored `MoeExpertCompute` for LL.

The modeling boundary is now explicit in code, README, the modeling doc, and the PR body: existing standard-MoE power-law curves own local expert-compute workload efficiency, while this PR's Monte Carlo applies only to LL dispatch/combine endpoint imbalance. We do not claim this reproduces `DeepEPMoE.run_moe_core`, and we do not want LL correctness to require a second dedicated compute collector.

I added a concrete FIXME to migrate the remaining HT/V2 users to standard MoE coverage and then deprecate `MoeExpertCompute` plus its dedicated collector/table; this PR deliberately leaves those existing paths unchanged.

I agree with the parity-visibility part of the review. H200 SGLang 0.5.14 numerical parity is restored alongside GB200, and the generation graph asserts standard `MoE`. Its static-generation golden now moves from about 30.43 ms to 42.8139 ms, so the selected compute-model migration is visible and reviewable instead of hidden behind a shape-only test.

crates/core/parity_tests/perfmodel/test_compile_engine_parity.py
Harrilee: Addressed in `fcff3c99` and `6bcbce1c`. The original H200 DeepSeek-V3 TP8/EP8 numerical scenario is restored under `wideep_sglang::*`, and the GB200 DeepSeek-R1 ADP32/EP32 scenario is retained under a separate `wideep_sglang_gb200::*` namespace. Both now pin static context, static generation, mixed, and decode surfaces.

I did not restore `MoeExpertCompute` for LL because standard `MoE` is the intentional modeling boundary described above. Consequently, the H200 re-pin intentionally exposes that compute-model change: static generation is 42.8139 ms rather than the former ~30.43 ms.

crates/core/parity_tests/perfmodel/test_compile_engine_parity.py
Harrilee: Addressed in `fcff3c99` and `6bcbce1c`. I removed the 0.5.6.post2 shape-only substitute and restored an end-to-end numerical H200 SGLang 0.5.14 DeepSeek-V3 TP8/EP8 scenario. It evaluates the full handle and pins all four parity surfaces.

The graph regression also asserts that context keeps the existing HT `MoeExpertCompute` path, while LL generation is exactly `FP8 dispatch -> standard MoE -> BF16 combine`. This deliberately makes the compute-model choice visible in the numerical golden rather than using missing fused-MoE data as a reason to avoid evaluation.

Arsene12358 (approved):
Approving. The parity items are fully addressed in `fcff3c99` / `6bcbce1c`: the H200 DeepSeek-V3 TP8/EP8 numeric scenario is back under `wideep_sglang::*`, GB200 is pinned separately under `wideep_sglang_gb200::*`, the 0.5.6.post2 shape-only substitute is gone, and the re-pinned H200 goldens (static generation 42.8139 ms, decode 13.8044 ms) match my independent main-vs-PR run to the last digit. 67 parity tests pass locally on `6bcbce1c`.

On the compute leg: I am accepting the standard-`MoE` selection as the stated Stage 1 modeling boundary rather than as a regression, on the basis that it is a deliberate team design decision. One request so the record matches that claim: please link the design doc or discussion where the Stage 1 design was agreed, specifically the choice of the ordinary fused-MoE predictor for LL expert compute and the planned migration/deprecation of `MoeExpertCompute` with its collector and table. The PR body or the `blocks/moe.py` thread is fine. I could not find it in #1592 (CodeRabbit-only reviews), in Linear, or in this PR. If the agreement lives in an offline conversation, a short Linear issue owning the deprecation follow-up would give the new FIXME a tracked target.

For the record, the measured consequence of the selection is the one quantified in my earlier review: at identical table coordinates the ordinary table is 1.1x to 9.6x slower than the `deepepmoe` rows, and decode step at batch 64 moves 1.5x to 3.3x versus `main` on shipped H200/GB200/B200 data. The restored H200 golden now makes that visible, which is what I asked for.

python/aisimulate/src/aiconfigurator_core/sdk/models/blocks/moe.py
Arsene12358: Approving the PR on the basis that this is a deliberate design decision. Please link the design doc or discussion where the standard-`MoE` selection for LL compute and the `MoeExpertCompute` deprecation were agreed (PR body or here). I could not find it in #1592, in Linear, or in this PR; if it was an offline decision, a Linear issue owning the deprecation follow-up would be the right home. Leaving this thread open until the pointer lands.

python/aisimulate/src/aiconfigurator_core/sdk/models/blocks/moe.py
```diff
@@ -616,24 +630,57 @@ def _large_ep_block_ops(
...
+        # SGLang's DeepEPMoE.run_moe_core kernel. The ordinary curve owns the
+        # local-compute workload-efficiency model; Monte Carlo owns only LL
+        # communication endpoint imbalance.
+        #
+        # FIXME(deprecation): migrate the remaining large-EP HT/V2 users to
+        # this standard MoE compute contract, then remove MoeExpertCompute and
+        # its dedicated collector/table dependencies. Do not add new
+        # MoeExpertCompute collection coverage.
+        routed_ops.append(
+            ops.MoE(
```
jasonqinzhou: Thank you for restoring the H200 numerical scenario; that fixes the earlier parity-visibility gap. The underlying accuracy issue remains, though: documenting `ops.MoE` as an intentional abstraction does not establish that its ordinary fused-MoE measurements predict the compute kernel SGLang runs with DeepEP. At matching H=7168/I=2048/K=8/N=256, FP8, power_law_1.01 coordinates, the checked-in ordinary and `deepepmoe` tables still differ by 1.1x to 9.6x, and the new H200 golden deliberately absorbs a 30.43 -> 42.81 ms static-generation change. Please either provide held-out/correlation evidence that validates this abstraction for DeepEP-LL compute, or retain `MoeExpertCompute` and its coverage gates so this communication-model PR continues to use the measured DeepEP kernel contract.

crates/core/src/perfmodel/operators/moe_ll_monte_carlo.rs
```diff
@@ -0,0 +1,887 @@
...
+    Ok(p50_ms)
+}
+
+fn estimate_uncached(request: MonteCarloRequest) -> Result<f64, AicError> {
+    validate_request(request)?;
+    if request.per_rank_tokens == 0 {
+        return Ok(0.0);
+    }
+
+    let trials = trial_count(request);
```
jasonqinzhou, line 162: Could we bound or amortize donor-route sampling before this becomes a sweep path—for example by validating a smaller converged trial count, caching a normalized routing statistic independently of fitted latency/payload, or adding a representative sweep benchmark? This code is unchanged by the update, and in a fresh exact-head optimized process a GB200 EP32 `decode_step_per_op(64, 1024, 4)` call took 6.793 seconds while the identical cached call took 194 microseconds. Because the key varies with token count, fitted variable time, payload, topology, and bandwidth, a sweep can repeatedly pay all 4,096 complete-assignment trials.

jasonqinzhou (dismissed):
**REQUEST_CHANGES · 6.7/10 · low confidence**

## What this PR does

This PR adds Stage 1 DeepEP low-latency decode modeling: exact and donor calibration resolution, OLS and one-shot fits, deterministic Monte Carlo routing pressure, topology-aware NVLink/IB guardrails, coverage gating, provenance, documentation, and current-head H200 plus GB200 parity pins.

The update directly improves reviewability by restoring the H200 numerical scenario alongside GB200 and pinning static, mixed, decode, and graph-shape behavior for both paths.

## Why this score

The completed Codex pass scores the current head 6.7 and would request changes because the explicit standard-MoE abstraction still lacks evidence that it predicts the DeepEP runtime kernel, while the untouched donor Monte Carlo path remains expensive on a cold query. The canonical result is BLOCKED because the independent helper requested claude-fable-5 but observed only claude-sonnet-4-6 and claude-haiku-4-5-20251001, so no compliant dual-review consensus or submittable recommendation can be produced.

jasonqinzhou:
Non-blocking follow-up: the 4,096-trial donor Monte Carlo path has a noticeable cold-query cost. On my machine, using an optimized synthetic merge of this head with current main, the focused EP64 donor test took 0.75 s of test runtime and the EP128 composite test took 1.74 s; repeated identical requests use the bounded cache.

I do not see this as a Stage 1 correctness blocker. Tracking production cold/warm benchmarks, realistic predict/recommend candidate-set cost, an agreed runtime budget, and the standard-MoE/MoeExpertCompute decision record in #104.

crates/core/src/perfmodel/operators/moe_ll_monte_carlo.rs
jasonqinzhou: Tracked as the non-blocking follow-up in #104 and linked from the PR body under “Non-blocking follow-up.” The issue covers optimized cold/warm EP8/32/64/128 benchmarks, realistic predict/recommend candidate-set cost, cache-hit rate, and an explicit runtime budget with remediation if it is missed. Stage 1 keeps Estimated provenance and its documented limitations in the meantime. Resolving this performance thread; the separate compute-accuracy review thread is unaffected.

python/aisimulate/src/aiconfigurator_core/sdk/models/blocks/moe.py
jasonqinzhou: Addressed at 2504559. I restored `MoeExpertCompute` for `deepep_ll` in every backend/phase and restored its coverage gating, so the standard-`MoE` design/deprecation decision is no longer part of this PR. The modeling doc and PR body now state that Monte Carlo applies only to dispatch/combine; #104 is narrowed to a non-blocking performance benchmark. Validation: 4,930 Python unit tests, 223 focused tests, 67 parity tests, and 1,123 Rust core tests passed.

python/aisimulate/src/aiconfigurator_core/sdk/models/blocks/moe.py
jasonqinzhou: Fixed at 2504559 by retaining measured `MoeExpertCompute` for `deepep_ll` and restoring the corresponding coverage gates. I also updated the graph assertions/docs and repinned the four affected H200/GB200 generation/decode goldens. Validation: 4,930 Python unit tests, 223 focused tests, 67 parity tests, and 1,123 Rust core tests passed.

PR #59, opened 2026-08-28
ci: add license compliance gates; hand security scanning to GitLab

pvijayakrish:
Python dependency license checks must pass before wheel staging, and nightly compliance evidence must preserve every dependency version. This PR adds that gate alongside SPDX header enforcement and keeps security scanning in the internal GitLab release-automation pipeline.

- Full CI includes Python compliance in both the wheel-build prerequisites and the aggregate result, using a shared license-check script with `--with-system` covering runtime packages that pip-licenses otherwise excludes.
- Nightly dependency installation runs in separate jobs without release credentials or access to built wheels. Both architectures must pass before the existing direct-to-Artifactory build/stage/download/smoke flow begins. A separate job combines Python inventories and the locked Rust graph into `deps.csv` and a version-aware `deps-diff.csv`. A failed prior-inventory lookup emits a workflow warning while preserving the first-run fallback.
- Scheduled first attempts remain unattended in the workflow. Reruns require an approval output from the current attempt. The production GitLab trigger is restricted to scheduled main runs, requires successful builds, license evidence, and the FPE matrix, and remains gated by `GITLAB_SECURITY_TRIGGER_ENABLED`.
- Failure alerts include timed-out jobs. Artifactory staging remains direct. The existing `nightly-dist-<arch>` GitHub artifacts retain the verified wheel/crate, checksums, and provenance for FPE, accuracy, and installation consumers; they are uploaded after the staged wheel is verified and before dependency execution. FPE reuses that exact amd64 wheel. CI and installation documentation describe the handoff and qualification order.

Validation at `acf3e3c10f273072e868f164dcb6d8f8bd2e5ed3`: 317 workflow, release, and repository-policy tests passed under Python 3.12; Actionlint, Ruff for the shared script and changed contract tests, whitespace, all 1,184 source headers, and packaged legal-file checks passed. Regression cases cover stale rerun approvals, failed/skipped/cancelled gates, concurrent crate versions and license changes, timeout alerts, Full CI aggregation, and the real nightly provenance writer/accuracy-consumer boundary (including altered wheels and wrong revisions). Hosted Fast/Full CI and CodeRabbit results are tracked on this PR. The schedule-only nightly workflow has not been executed at this head.

Deployment configuration remains a separate prerequisite: the current `manual-release-approver` environment has no required reviewers, while `automated-release` still requires a reviewer. The release owner must align those environment protections with the intended manual-rerun/unattended-schedule policy before relying on it. This PR does not change environment settings or approve deployment.

PR #81, opened 2026-09-02
fix(compliance): address OSRB filing requirements

saturley-hall:
## Summary

- remediate the OSRB license, SPDX header, and third-party notice findings
- package repository-level legal files in the Python wheel and enforce copy equality in CI
- stop storing wheels as GitHub Actions artifacts
- stage PR wheels in the dedicated PR Artifactory repository and post-merge wheels in the release repository

## Staging layout

- PR: `<pr-number>/<commit>/<run-id>/*.whl`
- main/release: `post-merge/<commit>/<run-id>/*.whl`

Uploads are checksum-verified and idempotent across reruns.

## Validation

- SPDX source check: 1,047 tracked files pass
- packaged legal-file copies match repository originals
- release artifact manifest check passes
- workflow YAML and staging-shell syntax parse successfully
- PR, post-merge, and rerun staging paths exercised with a mocked Artifactory endpoint

## Operational prerequisite

Replace the dummy `ARTIFACTORY_TOKEN` in the `pr-wheel-staging` environment with the scoped PR-repository token before expecting the wheel staging matrix to pass.


python/aisimulate/collector/wideep/vllm/patches/deepep_73b_nvl4.patch
```diff
@@ -1,3 +1,6 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
jasonqinzhou: These three header lines change this patch's SHA256 from `660b0281a932d56698e00495004bfe285841e48d73d473e180f2ba60fbbcb010` to `eb894098d1ca6520c97a45f6f88bb6b563e237ae71e433cd800e570f8fd1ab80`. The committed GB200/GB300 `collection_meta.yaml` records retain the old hash, while `attest_vllm_runtime()` now derives the new hash from this file and raises `VllmMoeA2ADeclarationError`; the Slurm launcher performs the same comparison against overlay build metadata. I reproduced the failure with the committed GB200 value. Please preserve this external patch's bytes (for example, exempt `.patch` and remove this preamble) or introduce a normalized content hash/migration without invalidating existing overlay and collection provenance.

.github/workflows/ci.yml
```diff
@@ -139,12 +166,132 @@ jobs:
...
       - run: python -m pip install 'maturin>=1.12,<2'
-      - run: python -m maturin build --release --out ../../dist
+      - name: Build application wheel
+        run: python -m maturin build --release --out ../../dist
         working-directory: python/aisimulate
-      - uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
-        with:
-          name: aisimulate-wheel-${{ matrix.arch }}
-          path: dist/*.whl
+      - name: Stage wheel in Artifactory
```
jasonqinzhou, line 189: Could we move this upload behind the validation jobs and smoke-test the exact wheel that is staged? `application-wheel` has no `needs`, so this exact-head run staged both wheels around 20 minutes before `Application Tests` finished; if tests or the release contract fails, a wheel from a failed run is already present in the PR or release repository. Those validation jobs also build/install separate wheels, so even a green run does not validate these exact bytes. A final gated stage job, or install/smoke checks in this job plus dependencies on the other gates, would keep failed or unverified wheels out of Artifactory.

scripts/check_copyright.py
```diff
@@ -0,0 +1,77 @@
...
+COPYRIGHT_MARKER = re.compile(
+    r"SPDX-FileCopyrightText: (?:Modifications )?Copyright \(c\) "
+    r"\d{4}(?:-\d{4})? NVIDIA CORPORATION & AFFILIATES\. All rights reserved\."
+)
+
+
+def is_source(path: Path) -> bool:
+    """Return whether the tracked path is subject to the header policy."""
+
+    return path.suffix in SOURCE_SUFFIXES or path.name.startswith("Dockerfile")
```
jasonqinzhou: This allowlist leaves tracked source files outside the new full-tree policy. For example, `python/aisimulate/docs/e2e-accuracy/app.js`, `python/aisimulate/eslint.config.js`, `python/aisimulate/src/aiconfigurator_core/_aiconfigurator_core.pyi`, and the extensionless shell script `python/aisimulate/collector/wideep/vllm/slurm/host_tools/ibstat` all currently have SPDX headers, but `is_source()` returns false for each, so removing those headers would still pass CI. Could we include the remaining code suffixes and detect tracked shebang scripts, while preserving the hash-stamped patch exception above?

THIRD_PARTY_NOTICES.md
```diff
@@ -0,0 +1,319 @@
...
+These notices apply only to the identified third-party material. Dependencies
+installed separately by package managers are governed by the license material
+distributed with those packages.
+
+## vLLM
+
+The following files are derived from vLLM's attention test utilities at tag
+`v0.11.0` (commit `b8b302cde434df8c9289a2b465406b47ebab1c2d`):
+
+- `collector/vllm/utils.py`
```
jasonqinzhou, line 21: The root and `python/aisimulate/` notice copies are intentionally byte-identical, but every AISimulate path in this notice resolves only relative to `python/aisimulate/`; for example, the repository root has no `collector/vllm/utils.py`. A reader following the canonical root notice cannot resolve these paths as written. Could the opening say that AISimulate file paths are relative to `python/aisimulate/`, or use unambiguous repository-root paths?

jasonqinzhou (changes requested):
**REQUEST_CHANGES · 4.8/10 · medium confidence**

## What this PR does

This PR remediates OSRB licensing findings by adding repository and wheel legal notices, normalizing SPDX headers, and enforcing those files in CI. It also replaces GitHub-hosted wheel artifacts with checksum-verified, branch-scoped Artifactory staging for PR, main, and release pushes.

The latest commit now carries the Transformers v4.57.1 license text verbatim in both synchronized notice copies, complementing the byte-equality and post-upload checksum safeguards already in the PR.

## Why this score

4.8/10 based on revalidated current-head evidence; the code-level decision would be REQUEST_CHANGES for two blocking regressions. The legal-text delta is correct, but the hash-stamped DeepEP patch remains invalidated and wheel bytes are still staged before independent validation; the source-policy and notice-path P2 gaps also remain. The canonical recommendation is BLOCKED because the required claude-fable-5 pass timed out without structured output or verified model evidence.

jasonqinzhou (approved):

pvijayakrish:
My [add-compliance-ci branch](https://github.com/ai-dynamo/aisimulate/pull/59) overlaps on the SPDX checker and ci.yml. I can rebase on yours after it merges and keep only the dependency-license gate + gitlab trigger.

pvijayakrish (approved):

PR #83, opened 2026-09-02
feat(models): add Qwen3.5 vision and Qwen3-VL video support

jasonqinzhou:
## Migration

This is the one-to-one AISimulate continuation of [AIConfigurator PR #1531](https://github.com/ai-dynamo/aiconfigurator/pull/1531) at source head `c4833f10757fec9501373b219adb2e79dbf58437`.

AIConfigurator is being refrozen for feature development, so review and landing continue here. The source PR retains its original discussion and history.

## Summary

- model Qwen3.5 dense and MoE vision encoders using the Qwen3-VL-derived tower, RoPE, PatchMerger, projector, and encoder DP/TP behavior
- support image workloads for Qwen3.5 and homogeneous video workloads for Qwen3.5 and Qwen3-VL through SDK, task, CLI, API, latency, memory, energy, and TTFT paths
- retain corrected language/GDN/MoE configuration while preserving the nested vision contract
- reject incomplete or mixed visual requests explicitly and keep support-matrix probes image-bearing
- cover the supported Qwen3.5 checkpoints plus the bundled 122B architecture

The port follows AISimulate's canonical package layout, compiled op-spec path, and generator-module contracts.

## Validation

- 626 focused unit tests passed across all changed files and affected Qwen, backend, task, CLI, generator, sweep, and support-matrix paths
- Ruff lint and format checks passed on changed Python files
- Python compilation and `git diff --check` passed

Silicon power and energy calibration remains a separate data follow-up; functional propagation is covered by tests.

Relates to AIC-1741


Arsene12358 (approved):
Reviewed at head `26d3b8a1`. **Approving: no merge blockers found.**

**What I verified**

- Port fidelity: per-file changed-line comparison against ai-dynamo/aiconfigurator#1531 at `c4833f10` — 23/25 files identical; the two test files differ only by one tuple-arity adaptation and one blank line.
- Ran the 12 touched test files in a fresh `uv sync --extra dev` venv (maturin-built Rust engine): 626 passed. Ruff check/format clean, `git diff --check` clean.
- Text-only parity vs `main` (Qwen3-VL-8B and Qwen3.5-27B, static + agg, h200_sxm vllm 0.24.0): TTFT/TPOT/memory identical to 4 decimals.
- Modeling cross-checked against HF transformers `modeling_qwen3_5.py` / `vision_utils.py`: single PatchMerger with no deepstack list (matches `supports_deepstack=False`; every bundled Qwen3.5/3.6 config has `deepstack_visual_indexes: []`), Conv3d patch embed, interpolated learned pos-embed, and `cu_seqlens = repeat_interleave(h*w, t)` — one attention sequence per temporal patch, exactly what `_encoder_shape` models. Video math checks out: 16 frames @ 448² → 8 × 196 = 1568 tokens, agg ctx 2592; 1 frame ≡ 1 image (ceil padding).
- All bundled Qwen3-VL / Qwen3.5 / Qwen3.6 configs load; the new `out_hidden_size == hidden_size` check never fires on them.
- The five CodeRabbit "Major" items from #1531 (temporal floor, agg cache key, AFD forwarding, support-matrix fail-fast, partial-video boundaries) are all addressed in this head.

**Soft findings (non-blocking) — all intentional, but none is mentioned in the PR body and each changes existing outputs, so please call them out before merging:**

1. **Support-matrix rows will flip PASS → FAIL on `a100_sxm` and `b60`.** VL probes are now image-bearing, but those two systems have no `encoder_attention` perf data, and HYBRID has no empirical fallback for that op. Confirmed through the real path: `SupportMatrix.run_single_test("Qwen/Qwen3-VL-2B-Instruct", "a100_sxm", ...)` is PASS on `main` and FAIL on this branch for both sglang 0.5.10 and vllm 0.14.0 (h200_sxm stays PASS). 68 currently published PASS rows are affected (40 on a100_sxm, 28 on b60). The flip is the honest outcome — those rows were text-only false positives — but two things would help: (a) a one-line warning in the PR body so the next `support-matrix-main` run isn't a surprise, and (b) a preflight in `_create_task` that raises `PerfDataNotAvailableError("<system>/<backend>/<version>: no encoder_attention perf data")` when the architecture is encoder-bearing but the system lacks the family. Today the FAIL row stores `sweep_agg`'s arbitrary "selected exception" — on a100_sxm that surfaced an unrelated sglang CP-constraint message, which will misdirect whoever triages the new FAILs. Collecting `encoder_attention` on a100_sxm/b60 is the proper long-term fix.

2. **`estimate --mode agg` default `ctx_tokens` changed from `isl` to `isl + visual_context_tokens`** (`cli_estimate` → `_run_agg_estimate`, and `Task`'s agg path). This aligns estimate with default-mode sweeps, which already used `isl_eff`, so it is a consistency fix — but it moves existing Qwen3-VL numbers: Qwen3-VL-8B, one 448² image, isl 1024 / osl 128 / bs 8 on h200_sxm vllm goes from TTFT 114.78 ms (ctx 1024) to 89.06 ms (ctx 1220), −22%. Suggest one sentence in `docs/cli_user_guide.md` and the PR body noting the new default and that `--ctx-tokens <isl>` reproduces the old budget.

3. **New `encoder_patch_embed_gemm` / `encoder_position_embed` / `encoder_merger_norm` ops shift existing Qwen3-VL image cost** (+0.65% encoder latency, +3.3 MiB encoder memory on the same probe). Fidelity fix, worth a line in the description.

4. Nit: `_run_encoder_phase` re-derives `has_video_workload` and the temporal ceil-division that `_encoder_pre_merge_per_visual` already computed, and re-checks a divisibility invariant the helper already enforced. Returning `temporal_sequences_per_visual` from the helper (1 for images) would drop ~20 lines and one duplicated invariant. Fine to defer.

**Pre-existing, not attributable to this PR (for awareness only):** `tests/e2e/cli/test_cli_api_equivalence.py::test_cli_default_api_vs_command` fails identically on `main` (`_moe_comm_fallbacks` / `_per_ops_source` column mismatch); an image on a non-VL model still degrades silently to text-only while a video now raises; encoder weights are only counted in memory when a visual workload is present.

PR #84, opened 2026-09-02
feat(models): model Kimi K2.5 image and video encoder

jasonqinzhou:
## Migration

This is the AISimulate continuation of [AIConfigurator PR #1527](https://github.com/ai-dynamo/aiconfigurator/pull/1527) at source head `406f064994e07a23f211292d88705a58d141b24b`.

## Summary

- model the Kimi K2.5 MoonViT3D tower, pooled PatchMerger/projector, and encoder communication
- preserve BF16 vision precision while retaining checkpoint-specific language quantization
- account for image and video encoder latency, memory, energy, context tokens, and TTFT
- model each supported video (at most four sampled frames) as a joint spatial-temporal attention sequence before temporal pooling; reject longer videos, including token overrides
- apply Kimi resize/padding and processor budgets; preserve independent native image/video defaults and reject unsupported mismatched side limits
- keep the projector replicated under encoder TP and reject positive but incompatible vision head geometry
- fail closed when model-level quantization does not explicitly exclude vision components, even when root and text quantization dictionaries have equal contents
- preserve declared quantization scope across repeated loads while retaining existing language quantization behavior
- pin adapted Transformers and vLLM revisions in both packaged third-party notices

The branch is reconciled with the current shared video/runtime and compiled engine-step APIs; obsolete duplicate CLI/task plumbing was dropped.

## Validation

Reviewed head: `03bc3f53c1b94fab31bd8cb50fc3ecd20b51bf88`, after merging main `e456bea46a4e54806d2bb2551db51d529b05009b` (Llama 4 PR #82).

- Resolved the shared configuration/backend/parser and notice conflicts while preserving both parent implementations. Main's positional `VisionEncoderConfig` fields remain unchanged, and Kimi's fields are appended. Pooled Kimi video retains one joint spatial-temporal attention sequence in the shared workload representation.
- Fresh independent task review passed 491 tests; 120 native encoder scenarios match the appropriate parents, including 18 expected rejections. All 72 Llama root/nested quantization constructor controls and 24 language graph cases match their respective parent.
- Fresh whole-branch review passed 516 tests and matched six language graphs and twelve Qwen native SOL predictions to updated main. Root separately matched all 24 Qwen SOL/SILICON predictions.
- Implementer validation passed all 1,380 selected tests, with one package-relative documentation test rerun from the package directory. Focused lint/format, actual Fast CI lint scope, legal parity and whitespace checks passed. Broad existing lint/format findings match main exactly.
- Earlier source and numerical validation remains recorded in the prior reviews. The merge retains the quantization-scope correction from `6dc7ac48`, changes no Rust implementation or numerical golden, and preserves both packaged notice sets byte-for-byte.

Two nonblocking findings remain documented: projector input normalization overcounts temporal rows by 0.0018384125 ms (0.0547% of encoder SOL latency in the four-frame 448-square b200/TRT-LLM TP1 case), and the existing complete Transformers attribution for pooling should also be repeated in the `vit.py` header. Both source/package notices retain complete immutable sources, owners and licenses. The timing comparison is analytical, not a measured GPU bound.

The four sibling encoder PRs overlap in shared runtime/configuration and notice files. Land them one at a time and refresh/revalidate the remaining PRs after each squash merge. Required code-owner approval is still outstanding.

This is architecture and analytical-accounting coverage, not measured Kimi K2.5 encoder silicon accuracy. Performance-data collection and calibration remain follow-up work under AIC-1742.

Current-head Fast CI and external review are complete for `03bc3f53c1b94fab31bd8cb50fc3ecd20b51bf88`. The null-quantization and EPD summary warnings were reproduced identically on both parents and the merged head; the remaining schema-documentation omission is a nonblocking follow-up already present before the merge. [Full CI](https://github.com/ai-dynamo/aisimulate/actions/runs/34473966623) passed on this commit, including both application wheels and release artifact contracts. Each architecture passed 5,674 package tests (12 skipped, one expected failure), 551 root tests (five skipped), and five smoke tests. [Approval was renewed on this exact head](https://github.com/ai-dynamo/aisimulate/pull/84#pullrequestreview-5167086651). Prior green runs and approval on `6dc7ac48` remain historical evidence.

Public EPD sweep support remains an existing limitation for these Kimi and Llama checkpoints; colocated encoder and language-only worker validation does not establish end-to-end EPD support. The model-info README and complete dataclass attribute documentation also remain a nonblocking follow-up.


python/aisimulate/src/aiconfigurator_core/sdk/models/blocks/vit.py
```diff
@@ -140,6 +159,16 @@ def _vit_transformer_ops(enc_cfg: common.VisionEncoderConfig, tp_size: int) -> l
         rope_dim = 6 * (n_vit // tp_size) * head_size_vit
         result.append(ops.ElementWise("encoder_rope_apply", depth, rope_dim, rope_dim, 0.8))
 
+    if enc_cfg.final_norm:
+        result.append(ops.ElementWise("encoder_final_norm", 1, h_vit, h_vit, 0.8))
+    if enc_cfg.projector_pre_norm:
+        result.append(ops.ElementWise("encoder_merger_pre_norm", 1, h_vit, h_vit, 0.8))
+    if enc_cfg.pool_temporal:
+        # Kimi's sd2_tpool merger performs temporal mean pooling and spatial
```
jasonqinzhou, line 173: The Kimi K2.5 tower, temporal-pooling, and precision/exclusion behavior is adapted from external Transformers/vLLM implementations, but this PR does not identify immutable source revisions or extend the packaged notices. Add commit-pinned adjacent provenance and update both THIRD_PARTY_NOTICES copies.

jasonqinzhou: Fixed in d117033c: adjacent source notes now pin the exact Transformers and vLLM commits, and both byte-identical packaged notices include the source and license references.

jasonqinzhou (commented):
Exact-head self-review after the current-main refresh: no new findings. The merge was clean; the shared engine/model/support-matrix regression suite passed locally (315 tests), and trusted Full CI is terminal green on amd64 and arm64.

jasonqinzhou:
Readiness workflow finding at a994e74a3540d5d1eb8e275a6709bc96496a7333: this non-draft PR is missing the repository-required review-ready label. The direct PR-event Fast CI run therefore skips its substantive jobs and fails its aggregate gate; the separately dispatched exact-SHA Full CI is not label-gated. Applying review-ready now to admit the normal PR checks and automated review. This is workflow metadata only, not a human approval or merge-readiness claim. The exact-head Full CI run has already passed.

THIRD_PARTY_NOTICES.md
```diff
@@ -46,6 +46,17 @@ The following files are derived from vLLM's attention test utilities at tag
 Upstream source:
 https://github.com/vllm-project/vllm/blob/v0.11.0/tests/v1/attention/utils.py
 
+The Kimi K2.5 vision-tower topology and pooled multimodal adapter modeled in
+`src/aiconfigurator_core/sdk/models/blocks/vit.py` and parsed in
```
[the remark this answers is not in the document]
jasonqinzhou: The path basis is explicitly stated in the preamble at lines 13-14: AISimulate file paths in this document are relative to python/aisimulate/ unless otherwise stated. Therefore src/aiconfigurator_core/... resolves to the correct repository files, and the identical packaged notice remains correct as well. The proposed prefix would double python/aisimulate/ under that declared basis. No source-path change is needed here. @coderabbitai please re-evaluate and resolve this finding using the documented path basis.

python/aisimulate/src/aiconfigurator_core/sdk/utils.py
```diff
@@ -565,6 +568,79 @@ def _parse_qwen_vision_encoder_config(
...
+    if out_hidden_size != expected_out_hidden_size:
+        raise ValueError(
+            f"Kimi K2.5 vision text_hidden_size ({out_hidden_size}) does not match "
+            f"text_config.hidden_size ({expected_out_hidden_size})"
+        )
+    return VisionEncoderConfig(
+        depth=int(vision_cfg["vt_num_hidden_layers"]),
+        hidden_size=hidden_vit,
+        num_heads=int(vision_cfg["vt_num_attention_heads"]),
+        intermediate_size=int(vision_cfg["vt_intermediate_size"]),
```
jasonqinzhou, line 713: [P2] Reject non-positive Kimi vision head counts at the parse boundary. A checkpoint with vt_num_attention_heads=0 currently reaches the generic ViT builder and divides hidden_size by zero, producing ZeroDivisionError instead of the typed invalid-configuration ValueError. Reproduced against this head. Validate a positive integer before arithmetic or encoder construction and cover invalid head-count values.

python/aisimulate/tests/unit/sdk/models/test_kimi_k25_vision.py
```diff
@@ -0,0 +1,237 @@
...
+    encoder_attention = next(op for op in model.encoder_ops if op._name == "encoder_attention")
+    assert encoder_attention._fmha_quant_mode == common.FMHAQuantMode.bfloat16
+
+
+def test_model_level_quantization_must_explicitly_exclude_kimi_vision_components():
+    raw = deepcopy(get_model_config_from_model_path("nvidia/Kimi-K2.5-NVFP4")["raw_config"])
+    raw["quantization_config"]["ignore"] = ["language_model.lm_head"]
+
+    with pytest.raises(ValueError, match="does not explicitly exclude both vision_tower and mm_projector"):
+        _parse_hf_config_json(raw)
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in 1a2a1ed4. Added parameterized tests for all four unsupported topology guards and a pooled-video token override through the real runtime shape resolver. The latter asserts attention (2,3200), projector (2,100), and 100 language visual tokens. The expanded local suite passes. @coderabbitai please re-evaluate this finding against the pushed fix and resolve if satisfied.

python/aisimulate/src/aiconfigurator_core/sdk/utils.py
```diff
@@ -565,6 +568,79 @@ def _parse_qwen_vision_encoder_config(
...
+
+    Sources: Transformers commit cbc1651a032b923da7f4b44b3d0e6f68e6ba6b55
+    and vLLM commit d2906091bfc579cebefe3d8e8fb9077397ce9882.
+    """
+    if not vision_cfg:
+        return None
+
+    merge_kernel = vision_cfg.get("merge_kernel_size", [2, 2])
+    if not (isinstance(merge_kernel, list) and len(merge_kernel) == 2 and merge_kernel[0] == merge_kernel[1]):
+        raise ValueError(f"Kimi K2.5 requires a square merge_kernel_size, got {merge_kernel!r}")
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in 1a2a1ed4. The quantization exclusion container is validated as a list/tuple of strings before iteration. Invalid null/string/scalar/mapping/mixed containers raise ValueError; valid list and tuple forms remain accepted. The expanded local suite passes. @coderabbitai please re-evaluate this finding against the pushed fix and resolve if satisfied.

python/aisimulate/src/aiconfigurator_core/sdk/utils.py
jasonqinzhou: Fixed in 1a2a1ed4. The parser rejects non-positive and non-integer head counts before division/modulo or tower construction. All six malformed-head regression cases now raise the intended ValueError, and the expanded suite passes.

python/aisimulate/tests/unit/sdk/models/test_kimi_k25_vision.py
```diff
@@ -0,0 +1,306 @@
...
+def test_kimi_quantization_accepts_valid_ignore_container(container):
+    raw = deepcopy(get_model_config_from_model_path("nvidia/Kimi-K2.5-NVFP4")["raw_config"])
+    raw["quantization_config"]["ignore"] = container(["vision_tower.*", "mm_projector.*"])
+    assert _parse_hf_config_json(raw)["encoder_config"].hidden_size == 1152
+
+
+@pytest.mark.parametrize("heads", [0, -1, False, None, 1.5, "16"])
+def test_kimi_k25_parser_rejects_invalid_vision_head_count(heads):
+    raw = deepcopy(get_model_config_from_model_path("moonshotai/Kimi-K2.5")["raw_config"])
+    raw["vision_config"]["vt_num_attention_heads"] = heads
```
jasonqinzhou, line 282: The new strict head-count regression covers False but not True. Because bool is an int subclass and False is also zero, these cases would still pass if the boolean-specific rejection were accidentally removed. Add True to protect the positive-boolean path, matching the equivalent Gemma review finding.

jasonqinzhou: Fixed in 50f81cf4: the invalid-head cases now include both True and False, and all targeted model tests pass. This is test-only coverage; the strict production guard was already correct.

python/aisimulate/src/aiconfigurator_core/sdk/utils.py
```diff
@@ -565,6 +568,85 @@ def _parse_qwen_vision_encoder_config(
...
+        raise ValueError("Kimi K2.5 vision vt_num_attention_heads must be a positive integer")
+    spatial_merge_size = int(merge_kernel[0])
+    merger_dim = hidden_vit * spatial_merge_size**2
+    out_hidden_size = int(vision_cfg["text_hidden_size"])
+    if out_hidden_size != expected_out_hidden_size:
+        raise ValueError(
+            f"Kimi K2.5 vision text_hidden_size ({out_hidden_size}) does not match "
+            f"text_config.hidden_size ({expected_out_hidden_size})"
+        )
+    return VisionEncoderConfig(
```
Arsene12358, line 709: [P1] Apply Kimi image and video resizing instead of Qwen rounding

The new Kimi encoder configuration enters the existing Qwen smart-resize path. The pinned Kimi processor pads up to the patch/merge stride after its own resolution limits; Qwen rounds to the nearest stride. The new image/video support consequently returns incorrect encoder shapes and language context lengths for ordinary non-aligned inputs. This violates the goal to account for image/video context tokens, latency and memory.

| Current | Required correction |
| --- | --- |
| 448 × 449 image<br>LM tokens: 256<br>ViT patches: 1,024<br><br>4,000 × 4,000 image<br>LM tokens: 20,449 | Use Kimi navit_resize and checkpoint processor limits.<br>448 × 449 → padded 448 × 476<br>LM tokens: 272; ViT patches: 1,088<br>4,000-square at pinned default limits: 4,225 LM tokens.<br>Retain the Qwen path for Qwen models. |

The tuple is `(projected_tokens_per_visual, encoder_patches_per_visual, visual_count)`. Executing the exact `navit_resize` function from the pinned upstream image processor pads 448×449 to 448×476, yielding 272 projected tokens and 1,088 patches. A 4,000-square input additionally yields 20,449 projected tokens here versus 4,225 with the pinned processor defaults. The mismatch is in the newly enabled Kimi path; Qwen’s existing rounding behavior should remain unchanged.

Reproducer (run in the installed PR environment):

```python
from aiconfigurator_core.sdk.config import RuntimeConfig
from aiconfigurator_core.sdk.backends.base_backend import BaseBackend
from aiconfigurator_core.sdk.utils import get_model_config_from_model_path
enc = get_model_config_from_model_path("moonshotai/Kimi-K2.5")["encoder_config"]
print(BaseBackend._encoder_pre_merge_per_visual(RuntimeConfig(image_height=448, image_width=449), enc))
# Actual: (256, 1024, 1); upstream geometry: (272, 1088, 1)
```

Sources: [Pinned Kimi image processor](https://github.com/huggingface/transformers/blob/cbc1651a032b923da7f4b44b3d0e6f68e6ba6b55/src/transformers/models/kimi_k25/image_processing_kimi_k25.py#L50-L91), [Pinned Kimi video processor](https://github.com/huggingface/transformers/blob/cbc1651a032b923da7f4b44b3d0e6f68e6ba6b55/src/transformers/models/kimi_k25/video_processing_kimi_k25.py).

python/aisimulate/src/aiconfigurator_core/sdk/backends/base_backend.py
```diff
@@ -370,17 +370,21 @@ def _smart_resized_spatial_tokens(height: int, width: int) -> tuple[int, int]:
                 # frame, so a partial group still produces one temporal patch.
                 temporal_patches = -(-video_frames // enc_cfg.temporal_patch_size)
                 spatial_post_merge, spatial_pre_merge = _smart_resized_spatial_tokens(video_height, video_width)
-                tokens_per_visual = temporal_patches * spatial_post_merge
+                tokens_per_visual = (
```
Arsene12358, line 511: [P1] Split video into temporal chunks before attention and pooling

With pool_temporal=True the PR pools the entire video into one spatial output and evaluates one joint attention sequence. The pinned Kimi video processor splits sampled frames into chunks of up to four frames first. The public video_frames input accepts longer videos, so this produces incorrect context and attention work beyond four frames. The PR goal of joint spatial-temporal attention is valid within each processor chunk.

| Current | Required correction |
| --- | --- |
| 8 frames, each 448 × 448<br>Attention: 1 × 8,192 patches<br>Projected context: 256 tokens<br>The projector runs on 256 rows. | Represent two 4-frame chunks.<br>Attention: 2 × 4,096 patches<br>Projected context: 512 tokens<br>The projector runs on 512 rows in total.<br>Alternatively reject >4 frames explicitly until chunk modeling exists. |

The runtime attention shape was also captured at `_run_encoder_phase_with_rust`: `(batch, sequence_length)=(1,8192)`, whereas the pinned processor produces two independent four-frame chunks with shape `(2,4096)`. The four-frame case agrees. At eight frames the current code halves projected context and doubles the attention-pair count relative to the two equal chunks.

Reproducer (run in the installed PR environment):

```python
from aiconfigurator_core.sdk.config import RuntimeConfig
from aiconfigurator_core.sdk.backends.base_backend import BaseBackend
from aiconfigurator_core.sdk.utils import get_model_config_from_model_path
enc = get_model_config_from_model_path("moonshotai/Kimi-K2.5")["encoder_config"]
rt = RuntimeConfig(num_images_per_request=0, num_videos_per_request=1, video_frames=8, video_height=448, video_width=448)
print(BaseBackend._encoder_pre_merge_per_visual(rt, enc))
# Actual: (256, 8192, 1)
# Upstream: 512 projected tokens, 8192 total patches in two 4096-patch attention sequences.
```

Sources: [Pinned processor chunk split](https://github.com/huggingface/transformers/blob/cbc1651a032b923da7f4b44b3d0e6f68e6ba6b55/src/transformers/models/kimi_k25/video_processing_kimi_k25.py#L171-L180).

python/aisimulate/src/aiconfigurator_core/sdk/models/deepseek.py
```diff
@@ -121,8 +123,15 @@ def __init__(
         backend_name: str = "",
         attention_quant_exclusions: frozenset = frozenset(),
         shared_expert_quant_mode: common.GEMMQuantMode | None = None,
+        encoder_config: common.VisionEncoderConfig | None = None,
     ) -> None:
         super().__init__(*args)
+        if encoder_config is not None:
+            self.encoder_config = encoder_config
```
Arsene12358, line 130: [P2] Keep the Kimi projector replicated under encoder TP

The newly attached Kimi tower delegates projector construction to the Qwen builder. With enable_encoder_dp=False, this shards both projector GEMMs and inserts an all-reduce. The pinned vLLM PatchMerger uses two ReplicatedLinear layers. Predictions therefore use the wrong per-rank matrix sizes, weight bytes and communication for this supported topology.

| Current | Required correction |
| --- | --- |
| Kimi K2.5, encoder TP2<br>fc0 (N,K): (2,304, 4,608)<br>fc1 (N,K): (7,168, 2,304)<br>encoder_projector_ar: TP2 | Use the Kimi projector topology.<br>fc0 (N,K): (4,608, 4,608)<br>fc1 (N,K): (7,168, 4,608)<br>No projector all-reduce.<br>Keep the tower’s own TP communication. |

The new call to the generic builder makes this topology reachable for Kimi K2.5. The pinned vLLM PatchMerger declares both projections as `ReplicatedLinear`; consequently encoder TP2 must retain the full projector matrix sizes per rank and has no projector all-reduce. This affects per-rank weights, GEMM work and communication.

Reproducer (run in the installed PR environment):

```python
from aiconfigurator_core.sdk.config import ModelConfig
from aiconfigurator_core.sdk.models import get_model
model = get_model("moonshotai/Kimi-K2.5", ModelConfig(tp_size=2, moe_tp_size=2, enable_encoder_dp=False), "vllm")
for op in model.encoder_ops:
    if "encoder_projector" in op._name:
        print(op._name, getattr(op, "_n", None), getattr(op, "_k", None), getattr(op, "_tp_size", None))
# fc0 GEMM: N=2304, K=4608; fc1 GEMM: N=7168, K=2304; projector AR: TP2.
```

Sources: [Pinned PatchMerger implementation](https://github.com/vllm-project/vllm/blob/d2906091bfc579cebefe3d8e8fb9077397ce9882/vllm/model_executor/models/kimi_k25_vit.py#L918-L945).

python/aisimulate/src/aiconfigurator_core/sdk/utils.py
```diff
@@ -565,6 +568,85 @@ def _parse_qwen_vision_encoder_config(
...
+
+    # The Moonshot checkpoint scopes quantization to text_config. NVIDIA's
+    # model-level NVFP4 config explicitly excludes both vision components.
+    quant_is_text_only = root_quant_cfg is not None and root_quant_cfg == nested_text_quant_cfg
+    if root_quant_cfg and not quant_is_text_only:
+        exclusions = root_quant_cfg.get("ignore", [])
+        if not isinstance(exclusions, (list, tuple)) or any(not isinstance(pattern, str) for pattern in exclusions):
+            raise ValueError("Kimi K2.5 quantization_config.ignore must be a list or tuple of strings")
+        ignore = tuple(pattern.lower() for pattern in exclusions)
+        vision_ignored = any("vision_tower" in pattern for pattern in ignore)
```
Arsene12358: [P2] Require quantization exclusions to cover each complete vision component

The stated fail-closed contract for model-level quantization is checked with substring membership. An exclusion naming one child of vision_tower and one child of mm_projector is accepted as excluding both complete components. The new builder then models every vision operation in BF16 even though the metadata does not establish that precision for the remaining modules. This is a narrower configuration-validation defect in addition to the default-workload errors above.

| Current | Required correction |
| --- | --- |
| quantization_config.ignore = [<br>  'vision_tower.encoder.layers.0',<br>  'mm_projector.linear_1'<br>]<br>Parser accepts the entire BF16 vision graph. | Verify that the backend exclusion patterns cover both whole components, using their actual matching semantics. Reject partial or ambiguous coverage. Add the narrow-child exclusions as a negative regression while preserving the bundled fully excluded NVFP4 configuration. |

The two exclusion patterns cover specific child modules, not the complete vision tower and projector. The parser’s substring checks accept them as proof that both entire components are excluded, then the builder forces the whole vision graph to BF16. The PR explicitly promises to fail closed on ambiguous model-level quantization; this accepted input violates that contract.

Reproducer (run in the installed PR environment):

```python
from copy import deepcopy
from aiconfigurator_core.sdk.utils import get_model_config_from_model_path, _parse_hf_config_json
raw = deepcopy(get_model_config_from_model_path("nvidia/Kimi-K2.5-NVFP4")["raw_config"])
raw["quantization_config"]["ignore"] = ["vision_tower.encoder.layers.0", "mm_projector.linear_1"]
print(_parse_hf_config_json(raw)["encoder_config"])
# Accepted: complete VisionEncoderConfig; build_encoder_ops models all vision GEMMs as BF16.
```

Sources: [PR quantization exclusion check](https://github.com/ai-dynamo/aisimulate/blob/50f81cf40f91a90728497ef49f016f31477af384/python/aisimulate/src/aiconfigurator_core/sdk/utils.py#L606-L617).

Arsene12358 (dismissed):
Requesting changes on `50f81cf40f91a90728497ef49f016f31477af384` for the 4 blockers detailed in the inline comments.

- [P1] Apply Kimi image and video resizing instead of Qwen rounding (spec).
- [P1] Split video into temporal chunks before attention and pooling (spec).
- [P2] Keep the Kimi projector replicated under encoder TP (spec).
- [P2] Require quantization exclusions to cover each complete vision component (spec).

Independent validation: 988 selected package tests passed. Six existing text prefill/decode operation graphs and twelve Qwen SILICON-query static/aggregate predictions match base `b6b0a386`. The inline findings come from additional input/shape reproductions and pinned upstream or repository-rule checks; passing suites do not resolve them.

Only findings introduced by this PR are requested here. Existing EPD candidate-selection limitations and unmeasured GPU calibration are excluded.

python/aisimulate/src/aiconfigurator_core/sdk/utils.py
```diff
@@ -565,6 +575,157 @@ def _parse_qwen_vision_encoder_config(
...
+                "Kimi K2.5 has model-level quantization but does not explicitly exclude both "
+                "vision_tower and mm_projector; refusing to infer encoder precision from the language model"
+            )
+
+    hidden_vit = int(vision_cfg["vt_hidden_size"])
+    vision_heads = vision_cfg["vt_num_attention_heads"]
+    if not isinstance(vision_heads, int) or isinstance(vision_heads, bool) or vision_heads <= 0:
+        raise ValueError("Kimi K2.5 vision vt_num_attention_heads must be a positive integer")
+    spatial_merge_size = int(merge_kernel[0])
+    merger_dim = hidden_vit * spatial_merge_size**2
```
Arsene12358, line 702: [P2] Reject non-integral Kimi K2.5 vision head widths

The new parser accepts positive but incompatible geometry. Setting the bundled checkpoint vt_num_attention_heads=17 while keeping vt_hidden_size=1152 succeeds, but encoder construction emits QKV n=3417 and an attention width of 17 × 67 = 1139. Reject incompatible hidden/head geometry before creating the vision config instead of silently truncating it. I reproduced this through the actual parser and serialized encoder consumer while checking the related Llama finding; Kimi K3 and Gemma already validate their corresponding head geometry. Add public-loader regressions for positive non-divisors with both language-only settings.

Arsene12358 (dismissed):
Requesting changes for one additional independently reproduced parser defect at `10b3f78c46dd2228de03de2e7212b4418cbe46bc`: incompatible positive vision head counts silently truncate QKV geometry. The original four blockers are fixed; this narrow follow-up is being corrected and independently re-reviewed.

Arsene12358 (commented):
Codex follow-up review of `9261f7c4f37366fd45e3bae22ddd4d2b53f8abab`: all four original blockers and the additional head-width validation finding are fixed. Independent task and whole-branch reviews found no remaining blocking findings.

- 5,542 full unit tests passed (11 skipped) at `10b3f78c` before the final head-width guard. The guard passed 1,058 broad model/encoder/utility/support-matrix tests and 224 independent focused tests; the final documentation-only commit preserves identical runtime and test code. Local full runs exclude the torch-dependent dispatch test and documented non-TTY plain-output tests.
- Independent task review and a separate whole-branch Codex review found no remaining blocking findings at this exact head.
- Final independent checks include positive-but-incompatible head geometry through real BF16/NVFP4 local loaders in normal/language-only modes, 200 geometry cases against freshly fetched upstream processor code, and 24 static plus 24 mixed native SOL dimension/token-override comparisons.
- Both frozen parity suites passed all 363 cases without numerical golden changes. Their operation round-trip test now reflects the unchanged separate encoder FFI contract; native encoder execution is tested separately.
- Six language operation graphs and 24 Qwen predictions (12 SILICON and 12 configured SOL) match the pinned base exactly. Cases cover static/aggregate text/image/video, h200_sxm/vLLM 0.24.0, TP1, batch 4, ISL 1024 / OSL 8, 448×449 images and four 448-square video frames.
- Changed files pass lint, formatting, packaged legal parity, and whitespace checks. Current-head Fast CI, CodeRabbit, and Full CI must be evaluated separately from earlier runs.

One independently verified minor issue remains: the video projector normalization is charged for all pre-pool temporal patches. For one four-frame 448-square video on b200_sxm/TRT-LLM TP1, the excess is 0.0018384125 ms, approximately 0.0547% of encoder SOL latency. It is recorded as nonblocking; this analytical case is not a measured GPU bound.

Correction to my earlier validation wording: the original Qwen probe used a raw database template with SILICON query mode, despite passing SOL as its loading-policy argument. Final validation separately reran both actual query modes. Required CODEOWNER approval remains outstanding.

python/aisimulate/src/aiconfigurator_core/sdk/utils.py
Arsene12358: Kimi now uses its own resize, processor budgets and ceiling padding. The reproduced 448×449 image produces 272 language tokens/1,088 patches; 4000-square images produce 4,225/16,900; four 1024-square video frames produce 1,089/17,424. Local/remote native defaults remain independent, mismatched side limits reject explicitly, and legacy shared limits remain supported. Independent upstream geometry and actual native consumers pass.

Verified head: `9261f7c4f37366fd45e3bae22ddd4d2b53f8abab`; independent task and whole-branch review gates pass.

python/aisimulate/src/aiconfigurator_core/sdk/backends/base_backend.py
Arsene12358: Implemented the explicit-rejection option from this review: more than four sampled frames per video now raises before shape calculation, including when num_video_tokens is supplied. Supported one-chunk videos keep the correct pooled context and joint attention shapes. Multi-chunk video modeling remains unsupported and the CLI guide states this limit.

Verified head: `9261f7c4f37366fd45e3bae22ddd4d2b53f8abab`; independent task and whole-branch review gates pass.

python/aisimulate/src/aiconfigurator_core/sdk/models/deepseek.py
Arsene12358: Both Kimi projector GEMMs remain full-sized under encoder TP: (4608,4608) and (7168,4608), with no projector AllReduce. Native DP2/TP2 tests verify replicated parameter counts and latency at equal local token loads while preserving tower and DP communication. The separately documented minor normalization overcount is nonblocking and does not undo this replication correction.

Verified head: `9261f7c4f37366fd45e3bae22ddd4d2b53f8abab`; independent task and whole-branch review gates pass.

python/aisimulate/src/aiconfigurator_core/sdk/utils.py
Arsene12358: Replaced substring inference with conservative supported ModelOpt full-component exclusions. Partial child paths, ambiguous prefixes/case and unsupported matching semantics reject; complete vision_tower/mm_projector coverage and bundled BF16/NVFP4 behavior pass regression tests.

Verified head: `9261f7c4f37366fd45e3bae22ddd4d2b53f8abab`; independent task and whole-branch review gates pass.

python/aisimulate/src/aiconfigurator_core/sdk/utils.py
Arsene12358: Added the parser divisibility guard in 104e30d9ea868624b643e6f6e12de523686cda01. Twelve public-loader incompatible-head cases failed before and pass after the fix; four valid controls preserve full 3456×1152 QKV and 16×72 attention. Independent old-parser counterfactuals and valid 16/24-head controls pass. The broad suite passed 1,058 tests and final independent review passed 224 focused tests.

Verified head: `9261f7c4f37366fd45e3bae22ddd4d2b53f8abab`; independent task and whole-branch review gates pass.

python/aisimulate/docs/cli_user_guide.md
```diff
@@ -473,6 +473,9 @@ Beyond `--ttft`, `--tpot`, `--isl`, `--osl`, and `--prefix`, `default` mode acce
 The image and video input flags are also available in `recommend` mode.
 
 Image and video workloads must currently be estimated separately. Qwen3-VL dense/MoE and Qwen3.5 video token accounting uses each model's temporal patch size before the spatial merge.
+
+Kimi K2.5 resizes to its processor's patch and side limits, then pads to the patch/merge stride. The bundled checkpoints use 16,384 image patches, 4,096 patches per video frame, and 512 patches per side before padding. Local and downloaded checkpoints may override these limits through `preprocessor_config.json` (legacy `media_proc_cfg` or native Transformers settings) and `video_preprocessor_config.json`. Video estimates support one processor chunk of up to four sampled frames per video, including when `--num-video-tokens` is supplied; longer videos are rejected until temporal chunk modeling is available. Kimi's projector remains replicated under encoder tensor parallelism.
```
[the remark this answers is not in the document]
Arsene12358: Documented in `9261f7c4f37366fd45e3bae22ddd4d2b53f8abab`: effective native image/video side limits must match; each otherwise retains its independent 512-patch default, while legacy media_proc_cfg shares the limit. Independent review checked the sentence against the parser and local/remote fixtures.

python/aisimulate/src/aiconfigurator_core/sdk/utils.py
```diff
@@ -565,6 +575,159 @@ def _parse_qwen_vision_encoder_config(
...
+        )
+    if vision_cfg.get("video_attn_type") != "spatial_temporal":
+        raise ValueError(
+            "Kimi K2.5 vision modeling requires video_attn_type='spatial_temporal', "
+            f"got {vision_cfg.get('video_attn_type')!r}"
+        )
+
+    # The Moonshot checkpoint scopes quantization to text_config. NVIDIA's
+    # model-level NVFP4 config explicitly excludes both vision components.
+    quant_is_text_only = root_quant_cfg is not None and root_quant_cfg == nested_text_quant_cfg
```
Arsene12358: [P2] Do not infer quantization scope from equal root/text dictionaries

A fresh checkpoint can contain equal but independently specified `quantization_config` dictionaries at the root and under `text_config`; equality does not establish that the root config was promoted from text-only metadata. Starting from the bundled NVFP4 config, removing the vision exclusions correctly rejects loading, but copying that same root dictionary into `text_config.quantization_config` makes the public loader accept it and model both projector GEMMs in BF16. I reproduced this through local `get_model` in both language-only modes, and the full worker reaches native SOL with the same projector latency and weights as the fully excluded control. At the pinned vLLM revision, ModelOpt quantization is passed to the projector and `mm_projector` is not automatically excluded, so this violates the promised fail-closed vision-precision contract. Preserve the original scope explicitly (or avoid mutating the cached raw config when promoting text quantization), then always validate genuinely root-level configurations even when their contents equal the nested text config. Add public-loader controls for equal root/text dictionaries with missing and complete exclusions, while preserving repeated loads of the bundled text-only-quantized checkpoint.

Sources: [pinned upstream](https://github.com/vllm-project/vllm/blob/d2906091bfc579cebefe3d8e8fb9077397ce9882/vllm/model_executor/models/kimi_k25.py#L318-L355), [pinned upstream](https://github.com/vllm-project/vllm/blob/d2906091bfc579cebefe3d8e8fb9077397ce9882/vllm/model_executor/layers/quantization/modelopt.py#L201-L225)

Arsene12358 (dismissed):
Final approval review found one additional independently reproduced blocker: equal root/text quantization dictionaries bypass model-level vision exclusion validation. Holding approval until this is corrected and independently reverified. The previously resolved findings remain fixed.

Arsene12358 (commented):
Final Codex re-review of `6dc7ac4891fb4b422bc29762a707122e366a08fd`: the quantization-scope bypass is corrected, and independent task, standards and whole-branch reviews found no remaining blocking findings.

Actual root-level quantization now always undergoes complete-component vision exclusion validation, even when it equals text metadata. Raw configuration retains declared scope; the shared language reader preserves root-first/text-fallback behavior without mutating that scope. Unsafe public-loader controls reject in both language-only modes; supported native SOL controls preserve projector latency and weights.

Validation: 1,150 broad tests (including 126 Kimi cases); 260 independently rerun focused tests; all 88 bundled language graphs and precision modes match freshly executed main; 24 Qwen predictions match in configured SOL and SILICON. Independent standards checks additionally passed 67 tests and 48 precedence/nonmutation cases. No Rust, numerical golden, performance-table or schema change was introduced by this correction.

The previously documented normalization overcount and provenance-header placement are P3/nonblocking. Current-head hosted checks must complete before approval; older Full CI runs are not evidence for this head.

python/aisimulate/src/aiconfigurator_core/sdk/utils.py
Arsene12358: Fixed in `6dc7ac4891fb4b422bc29762a707122e366a08fd`. The parser and inference helpers preserve the checkpoint's actual root/text scope instead of promoting nested quantization or comparing dictionary contents. Every genuine root declaration now goes through complete-component exclusion validation. The eight language readers use one root-first/text-fallback accessor, preserving quantization behavior for all 88 bundled models.

Public-loader regressions cover equal dictionaries with missing/partial/malformed exclusions and complete-exclusion controls, both worker modes, and repeated bundled/local/mock-downloaded text-only loads. Independent task and final whole-branch reviews reverified the correction, actual native SOL controls, unchanged language graphs/modes and Qwen predictions. The focused and broad checks passed; approval remains pending current-head hosted CI.

python/aisimulate/src/aiconfigurator_core/sdk/models/qwen35.py
```diff
@@ -47,7 +48,7 @@ def _qwen35_mixed_precision_gemm_modes(
         quantization = hf_quant.get("quantization")
         if isinstance(quantization, dict):
             layer_maps.append(quantization.get("quantized_layers"))
-    quantization = raw_config.get("quantization_config")
+    quantization = _get_language_quantization_config(raw_config)
```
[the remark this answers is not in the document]
Arsene12358: Independently checked against pinned main `b6b0a38618aa59181a6bdc80ae9d6f43082c17c1`, pre-fix `9261f7c4`, and current `6dc7ac48`: this mapping predates the PR. `qwen35.py` is unchanged between main and `9261f7c4`; the scope fix changes only how this consumer reads language metadata. The old loader already promoted nested quantization into the root before the consumer read it.

Four public Qwen3.5 controls (root/nested MXFP8 metadata × both language-only settings) produce identical context/generation Rust operation graphs, including `fp8` projection modes, and identical native SOL predictions on all three revisions. The tested context latency is 4.663 ms and summary memory is 24.161 GiB; temporary checkpoint path labels are the only output metadata difference.

A change to the MXFP8 mapping would therefore address preexisting behavior. It is outside this PR's introduced-regression scope and should be assessed separately. Resolving this conversation for this PR; the new accessor does not introduce a new mapping or newly expose the nested path.

python/aisimulate/src/aiconfigurator_core/sdk/utils.py
```diff
@@ -1250,9 +1418,19 @@ def _infer_quant_dynamic(quant_cfg: dict) -> bool | None:
...
 
 
+def _get_language_quantization_config(raw_config: dict) -> dict | None:
+    """Read root-first language quantization without changing its declared scope."""
+    if "quantization_config" in raw_config:
+        return raw_config["quantization_config"]
+    architecture = (raw_config.get("architectures") or [None])[0]
+    text_key = MULTIMODAL_TEXT_CONFIG_KEY.get(architecture)
+    nested = raw_config.get(text_key, {}) if text_key else {}
+    return nested.get("quantization_config", {}) if isinstance(nested, dict) else {}
```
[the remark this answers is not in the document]
Arsene12358: The current helper returns explicitly present root or nested `quantization_config` values unchanged. It does not convert `null`, lists, scalars or booleans into `{}`. Both MoE callers still invoke `.get(...)` directly on that value.

Independent public-loader checks compared actual main `b6b0a386`, pre-fix `9261f7c4`, and current `6dc7ac48`: all 48 Mixtral/Llama controls have identical results. With explicit `fp8_block` MoE mode, malformed root/nested values still raise `AttributeError` on all three revisions; absent, empty and valid dictionaries retain their existing successful behavior. A separate direct helper matrix also confirms that explicit malformed values are returned unchanged.

Improving the exception type/message is a separate validation change; the alleged new silent fallback is not present. Resolving this conversation because no PR-introduced acceptance or defaulting regression is established.

Arsene12358 (approved):
Final Codex review of `6dc7ac4891fb4b422bc29762a707122e366a08fd` against `b6b0a38618aa59181a6bdc80ae9d6f43082c17c1`: no remaining blocking findings. All requested blocking changes are resolved, including the quantization-scope bypass discovered during this final round. Independent task, standards and whole-branch reviews examined the correction and its actual native consumers.

The correction passed 1,150 broad tests; fresh independent task and final reviewers passed 1,082 and 260 tests respectively. All 88 bundled language operation graphs and quantization modes match freshly executed pinned main, as do 24 Qwen predictions across configured SOL and SILICON. Public-loader and native controls reject unsafe root quantization while preserving supported repeated loads and BF16 projector accounting. The two subsequent CodeRabbit comments were independently checked against main and the preceding head; neither establishes an introduced regression, and both conversations are resolved.

Direct Fast CI, CodeRabbit and [Full CI on amd64 and arm64](https://github.com/ai-dynamo/aisimulate/actions/runs/34457527634) passed on this exact head. This establishes architecture and analytical consistency, not measured GPU predictive accuracy.

Two nonblocking P3 findings remain documented: video projector normalization overcounts by 0.0018384125 ms (0.0547% of encoder SOL latency in the four-frame 448-square b200/TRT-LLM TP1 case), and the existing complete Transformers pooling attribution should also appear in the vit.py header. Both packaged notices retain immutable sources, owners, licenses and modification notices.

Approved with those nonblocking notes. Video support remains limited to four sampled frames. Merge the overlapping sibling PRs sequentially and revalidate the remaining heads after each merge.

Arsene12358 (commented):
Codex review of merge commit `03bc3f53c1b94fab31bd8cb50fc3ecd20b51bf88`, tree `191da997596a471a7498f638bb98179bab1391ed`, against both parents: prior PR84 `6dc7ac4891fb4b422bc29762a707122e366a08fd` and updated main `e456bea46a4e54806d2bb2551db51d529b05009b`. No remaining conflict or newly introduced blocking finding. Separate fresh task and whole-branch reviewers verified the merged behavior.

The resolution preserves both notice sets, Llama tiling/CLS/prompt accounting, Kimi processor budgets and joint attention before temporal pooling, and the previously fixed quantization scope. Main's existing positional encoder configuration fields remain unchanged; Kimi's fields are appended. The shared workload representation remains the single source of encoder shapes passed to Rust.

Independent task review passed 491 tests and compared 120 native encoder scenarios against the appropriate parents, including 18 expected input rejections. All 72 Llama root/nested quantization constructor controls and 24 language graph cases match their parent. Independent whole-branch review passed 516 tests and matched six language graphs plus twelve Qwen native SOL predictions to fresh main. Root additionally matched 24 Qwen SOL/SILICON predictions. Implementer validation covered 1,380 selected tests, with one package-relative documentation test rerun successfully from its required working directory. Focused lint/format, actual Fast CI lint scope, legal parity, and whitespace checks passed; broader preexisting lint/format findings are identical to main.

The two previously documented P3 notes remain unchanged: source-reference placement and the small Kimi video normalization overcount. No numerical golden, Rust implementation, or performance data was changed. This is native analytical consistency evidence, not measured GPU accuracy. Current-head external review and hosted CI are being tracked separately before approval is renewed.

Arsene12358:
@coderabbitai review

Please review the conflict-resolution merge at `03bc3f53c1b94fab31bd8cb50fc3ecd20b51bf88`. The new base is `e456bea46a4e54806d2bb2551db51d529b05009b`; prior review and CI on `6dc7ac48` do not cover this merge. Independent task and whole-branch review results are recorded in the current-head Codex review.

Arsene12358:
External-review triage for `03bc3f53c1b94fab31bd8cb50fc3ecd20b51bf88` ([CodeRabbit review](https://github.com/ai-dynamo/aisimulate/pull/84#pullrequestreview-5166698317)): no newly introduced blocker.

The null-quantization warning is a preexisting input-hardening issue. A fresh independent reviewer ran 384 actual Llama constructor controls on main `e456bea4`, premerge PR84 `6dc7ac48`, and merged PR84 `03bc3f53`: each gives the same 324 successes and 60 AttributeErrors. Main already retains an explicitly present null root value, then calls `.get` on it when FP8-block construction is requested. Nested quantization was promoted only when the root key was absent. Both checkpoints, all three backends, and both language-only modes are covered. A controlled error message can be a separate hardening follow-up; changing null into an empty dictionary would change defaulting behavior.

The summary's EPD warning also reproduces identically on both parents: public agg/disagg EPD calls for both Kimi and both Llama checkpoints, plus a Qwen3.5 worker-candidate control, all retain the same existing rejection. No EPD code changed, and this merge does not add end-to-end EPD sweep support. The model-info README and complete dataclass attribute documentation remain a minor follow-up already present before the merge.

The actual Rust consumer evidence includes native operation shapes, latency, energy, source, visual context and memory, compared to each appropriate parent. Both original notice sets survive unchanged in byte-identical packaged/root copies. This is analytical consistency evidence, not measured GPU calibration. Detailed scripts/data are retained in the local review artifact; the compact machine-readable result below is published here for visibility.

```json
{
  "heads": {
    "main": "e456bea46a4e54806d2bb2551db51d529b05009b",
    "prior84": "6dc7ac4891fb4b422bc29762a707122e366a08fd",
    "merged84": "03bc3f53c1b94fab31bd8cb50fc3ecd20b51bf88"
  },
  "null_quantization": {
    "cases_per_revision": 384,
    "successes": 324,
    "same_AttributeErrors": 60,
    "all_equal": true
  },
  "epd_public_controls": {
    "cases_per_revision": 9,
    "same_existing_rejection": true
  },
  "native_encoder_parent_comparison": {
    "successes": 102,
    "expected_rejections": 18,
    "all_equal": true
  },
  "llama_quantization_parent_comparison": {
    "cases": 72,
    "all_equal": true
  },
  "qwen_SOL_SILICON_parent_comparison": {
    "cases": 24,
    "all_equal": true
  }
}
```

Arsene12358 (approved):
Final approval of conflict-resolution merge `03bc3f53c1b94fab31bd8cb50fc3ecd20b51bf88` against updated main `e456bea46a4e54806d2bb2551db51d529b05009b`. No remaining merge conflict or newly introduced blocking finding.

Fresh independent task and whole-branch reviews verified the combined implementation. The task reviewer passed 491 tests, matched 120 native encoder scenarios and 72 Llama quantization controls to the appropriate parents, and preserved all existing positional configuration fields. The whole-branch reviewer passed 516 tests. Root also matched all 24 Qwen SOL/SILICON predictions to fresh main. Both original notice sets are intact and byte-identical in the source/package copies.

The subsequent external null-quantization warning was independently reproduced identically across main, premerge PR84, and this commit in 384 constructor controls per revision. Public EPD limitations also reproduce unchanged on both parents. These are existing limitations, not regressions introduced by this PR or merge. Source-reference placement, the small Kimi normalization overcount, and schema documentation remain nonblocking follow-ups.

Fast CI, CodeRabbit review, and [Full CI on amd64 and arm64](https://github.com/ai-dynamo/aisimulate/actions/runs/34473966623) passed on this exact commit. Approved with the documented limits. This validation establishes native analytical consistency, not measured GPU predictive accuracy. Remaining sibling PRs must be updated and revalidated after each merge.

PR #85, opened 2026-09-02
feat: exercise multimodal encoders in support matrix

jasonqinzhou:
## Migration

This is the one-to-one AISimulate continuation of [AIConfigurator PR #1529](https://github.com/ai-dynamo/aiconfigurator/pull/1529) at source head `e12feed572599e60b6893d6ac92caaaf157f8783`.

AIConfigurator is being refrozen for feature development, so review and landing continue here. The source PR retains its original discussion and history.

## Summary

- detect declared vision encoders independently from implementation support
- exercise a canonical nonzero image workload for implemented encoders in aggregate and disaggregate runs
- require positive encoder latency and memory evidence before a multimodal row can pass
- persist and validate visual-workload evidence in the support-matrix schema and documentation
- classify declared-but-unimplemented encoders as `FAIL / ENCODER_UNSUPPORTED`

The port is adapted to AISimulate's package-local Dockerfile and compiled-only engine-step path. It remains independent from the model-implementation PRs in this migration set.

## Validation

- 131 focused unit tests passed across support-matrix generation, comparison, failure classification, documentation, and encoder paths
- Ruff lint and format checks passed on changed Python files
- Python compilation and `git diff --check` passed

Historical checked-in support-matrix CSVs are intentionally not regenerated; the next scheduled full generation will emit the new image-evidence schema.

Linear: AIC-1738


python/aisimulate/tools/support_matrix/support_matrix.py
```diff
@@ -642,7 +829,12 @@ def run_single_test(
             unsupported_modes = set(modes_to_test) - {"agg", "disagg"}
             if unsupported_modes:
                 raise ValueError(f"Unsupported support-matrix mode(s): {sorted(unsupported_modes)}")
-        constraints = _get_test_constraints(model)
+        try:
+            constraints = _get_test_constraints(model)
+        except Exception as exc:
+            raise ModelMetadataError(f"Failed to load model sizing metadata for {model!r}: {exc}") from exc
+        encoder_coverage = _get_encoder_coverage(model)
+        image_workload = encoder_coverage.workload
```
Arsene12358, line 874: **B1 (blocker):** this is now the only run for encoder-implemented models, but the image workload is only answerable where the database has `encoder_attention` perf data. a100_sxm and b60 have none, so every Qwen3-VL row there goes from PASS on main to a raw-traceback FAIL (A/B table in the review body), and the failure happens inside the sweep before `_assert_encoder_evidence` can turn it into `ENCODER_NOT_EXERCISED`.

Suggest a second deterministic preflight right after the `ENCODER_UNSUPPORTED` one, following the same pattern (explicit prefix, preflight replay command, no text-backbone run):

```python
if image_workload is not None and not _encoder_perf_data_available(system, backend, version):
    reason = (
        "ENCODER_DATA_UNAVAILABLE: "
        f"{system}/{backend} v{version} has no encoder_attention perf data; the canonical "
        f"{image_workload.num_images}x{image_workload.image_height}x{image_workload.image_width} "
        f"image workload for {model!r} cannot be exercised."
    )
    statuses = dict.fromkeys(modes_to_test, STATUS_FRAMEWORK_INCOMPATIBLE)
    error_messages = dict.fromkeys(modes_to_test, reason)
    commands = {
        mode: _encoder_preflight_row_command(  # generalise _encoder_unsupported_row_command
            model=model, system=system, backend=backend, version=version, mode=mode,
            expect_status=STATUS_FRAMEWORK_INCOMPATIBLE,
            expect_error_prefix="ENCODER_DATA_UNAVAILABLE:",
        )
        for mode in modes_to_test
    }
    if include_commands:
        return statuses, error_messages, commands, provenance
    return statuses, error_messages
```

with

```python
def _encoder_perf_data_available(system: str, backend: str, version: str) -> bool:
    """True when the encoder-attention table has rows for this database."""
    database = perf_database.get_database(system, backend, version)
    if database is None:
        return False
    EncoderAttention.load_data(database)  # aiconfigurator_core.sdk.operations.attention
    view = getattr(database, "_encoder_attention_data", None)
    return view is not None and len(view) > 0
```

`FRAMEWORK_INCOMPATIBLE` keeps these rows out of the Phase-2 sequential retry (only `STATUS_FAIL` is retried) and matches how missing perf-data files are classified for the rtx_pro sm120 gaps; a dedicated status is fine too. `check_csv_sanity` needs the mirror of the `ENCODER_UNSUPPORTED` replay-contract branch for the new prefix/status pair, and `model-roster.md` should define this third outcome.

python/aisimulate/tools/support_matrix/compare_support_matrix.py
```diff
@@ -338,7 +461,36 @@ def find_blocking_status_transitions(changed_rows: list[tuple]) -> list[str]:
...
+        except (TypeError, ValueError):
+            return True
+
+    encoder_migration_keys = set()
+    for row in new_data_rows or []:
+        key = tuple(row[:6])
+        if (
+            len(row) > 7
+            and row[6] == STATUS_FAIL
+            and row[7].strip().startswith("ENCODER_UNSUPPORTED:")
```
Arsene12358: **B1 (blocker, gate side):** the one-time waiver only recognises `FAIL` + `ENCODER_UNSUPPORTED:`. The a100_sxm/b60 Qwen3-VL rows regenerate as plain `FAIL` with a traceback, so against the checked-in matrix `find_blocking_status_transitions` reports 64 `Unexpected PASS -> FAIL transition` errors and `compare_support_matrix.py` exits 2 (simulated with `save_results_to_csv` + the exact `support-matrix-main.yml` invocation; with only the intended reclassifications it is 0 errors).

Once the preflight from the other comment exists, extend the waiver to the new pair, still restricted to legacy backbone-only old rows:

```python
_ENCODER_MIGRATIONS = (
    (STATUS_FAIL, "ENCODER_UNSUPPORTED:", _has_declared_unimplemented_encoder),
    (STATUS_FRAMEWORK_INCOMPATIBLE, "ENCODER_DATA_UNAVAILABLE:", _has_implemented_encoder),
)
encoder_migration_keys: set[tuple[tuple[str, ...], str]] = set()
for row in new_data_rows or []:
    key = tuple(row[:6])
    for status, prefix, predicate in _ENCODER_MIGRATIONS:
        if (
            len(row) > 7
            and row[6] == status
            and row[7].strip().startswith(prefix)
            and _is_legacy_backbone_only_row(old_rows.get(key))
            and predicate(row[0])
        ):
            encoder_migration_keys.add((key, status))
...
    if (key, new_status) in encoder_migration_keys:
        continue
```

Please add a compare test for the new waiver next to `test_explicit_encoder_unsupported_migration_is_not_blocking`, and keep `test_encoder_unsupported_after_image_backed_pass_is_blocking` semantics for the new status (an image-backed PASS regressing to `ENCODER_DATA_UNAVAILABLE` must still block).

python/aisimulate/tests/e2e/support_matrix/test_pr_support_matrix.py
```diff
@@ -35,6 +35,10 @@
     ("meta-llama/Meta-Llama-3.1-8B", "b200_sxm", "sglang"),
     ("MiniMaxAI/MiniMax-M2.5", "b200_sxm", "vllm"),
     ("openai/gpt-oss-20b", "h100_sxm", "trtllm"),
+    # Multimodal acceptance: dense and MoE backbones, with run_single_test
+    # exercising the canonical image workload in both agg and disagg modes.
+    ("Qwen/Qwen3-VL-8B-Instruct", "b200_sxm", "vllm"),
+    ("Qwen/Qwen3-VL-30B-A3B-Instruct", "b200_sxm", "vllm"),
```
Arsene12358, line 41: **B1 (test coverage):** these two cases cover one of the eleven support-matrix systems, which is how the a100_sxm/b60 flip stayed invisible in both #1529 and this port. The e2e budget probably cannot absorb an a100 case, so please pin the behaviour at unit level instead: a test in `test_multimodal_encoder_coverage.py` that monkeypatches an empty encoder-attention view (or `_encoder_perf_data_available` returning `False`) for `Qwen/Qwen3-VL-8B-Instruct` and asserts `FRAMEWORK_INCOMPATIBLE`, an `ENCODER_DATA_UNAVAILABLE:` reason, a preflight replay command, and that `_run_mode` is never called. A companion test asserting that a database *with* encoder data still takes the image path keeps the two preflights from shadowing each other.

Arsene12358 (changes requested):
## Review: one merge blocker (regression verified against main)

Thanks for the port. The feature works as specified on every system that has encoder-attention perf data: I ran `Qwen/Qwen3-VL-8B-Instruct` and `Qwen/Qwen3-VL-30B-A3B-Instruct` through `SupportMatrix.run_single_test` on all 45 system/backend/version combinations present in the checked-in matrix, and on b200_sxm, b300_sxm, gb200, gb300, h100_sxm, h200_sxm, l40s and rtx_pro_6000_server every one of the 160 mode-results is PASS with the image flags persisted in the replay command and silicon provenance. The two new e2e cases pass, the 173 `tests/unit/support_matrix` tests pass on a fresh venv against main-merged, ruff is clean, and the `ENCODER_UNSUPPORTED` preflight plus its `generate_support_matrix.py --expect-status/--expect-error-prefix` replay command work end-to-end.

### B1 (blocker): systems without `encoder_attention` perf data lose all Qwen3-VL coverage, and the compare gate rejects the regenerated matrix

`run_single_test` now uses the 1024x1024x1 image run as the *only* run for encoder-implemented models. `encoder_attention` perf data exists under 9 of the 11 support-matrix systems; **a100_sxm and b60 have none** (`EncoderAttention.load_data(db)` binds an `_encoder_attention_data` view with 0 rows for every a100/b60 database; `src/aiconfigurator_core/systems/data/{a100_sxm,b60}/` have no `encoder_attention/` family at all). With the image workload the sweep has no answerable perf data for any batch point and raises *before* `_assert_encoder_evidence` runs, so the row becomes a generic `FAIL` carrying a raw traceback rather than `ENCODER_NOT_EXERCISED` or `ENCODER_UNSUPPORTED`, its persisted replay command reproduces a crash, and Phase 2 retries it sequentially as if it were transient.

A/B with the same installed package and only the `tools/` tree swapped (main `4b1b5dc5` vs this PR at `664b5965`), agg / disagg:

| Model | System / backend / version | main tools | PR tools |
|---|---|---|---|
| Qwen3-VL-8B | a100_sxm / trtllm 1.0.0 | PASS / PASS | FAIL / FAIL |
| Qwen3-VL-8B | a100_sxm / sglang 0.5.10 | PASS / PASS | FAIL / FAIL |
| Qwen3-VL-8B | a100_sxm / vllm 0.14.0 | PASS / PASS | FAIL / FAIL |
| Qwen3-VL-30B-A3B | a100_sxm / trtllm 1.0.0 | HYBRID_PASS / HYBRID_PASS | FAIL / FAIL |
| Qwen3-VL-30B-A3B | a100_sxm / sglang 0.5.10 and vllm 0.14.0 | PASS / PASS | FAIL / FAIL |
| Qwen3-VL-8B and 30B-A3B | b60 / vllm 0.20.0 and 0.26.0 | PASS / PASS | FAIL / FAIL |
| both | b200_sxm / vllm 0.27.0 (control) | PASS / PASS | PASS / PASS |

Persisted error (agg; disagg is the prefill twin):

```
aiconfigurator_core.sdk.errors.NoFeasibleConfigError: sweep_agg: no results — 53325 batch point(s) had no answerable perf data
(e.g. FPM queries outside the collected domain, or no collected cell matches the model identity/quant modes). ...
```

Same-process control with the PR tools: Qwen3-VL-8B on a100_sxm/trtllm/1.0.0 with `image_workload=None` returns 416 Pareto rows (encoder_latency 0, i.e. the text-only PASS this PR wants to retire); with the canonical image workload it raises `NoFeasibleConfigError`. The flip is caused by the workload switch, not by the environment.

Pipeline consequence: the checked-in matrix has 64 Qwen3-VL PASS rows on a100_sxm (40) and b60 (24). Regenerating with this PR and running `compare_support_matrix.py --old … --new …` exactly as `support-matrix-main.yml` does yields **64 `Unexpected PASS -> FAIL transition` errors and exit code 2**, the branch the workflow turns into `::error::Validation errors detected`, so no daily PR is produced. The `ENCODER_UNSUPPORTED:` migration waiver does not cover this class. The same simulation with only the intended reclassifications (the 20 declared-but-unimplemented roster models plus image evidence on Qwen3-VL rows) produces 0 blocking errors, so the gate itself is sound; it is this one transition class that is unhandled.

### Suggested fix

The PR already has the right shape for "deterministic, cannot exercise the encoder": a preflight, an explicit reason prefix, a replayable preflight command, no Phase-2 retry, and a compare waiver limited to legacy backbone-only rows. Extend that shape to "encoder implemented, but this system/backend/version has no encoder perf data". Classifying it as `FRAMEWORK_INCOMPATIBLE` keeps it out of the Phase-2 retry loop for free (only `STATUS_FAIL` rows are retried) and matches how missing perf-data files are already classified for the rtx_pro sm120 gaps; a dedicated status would work equally well. The inline comments cover `run_single_test`, the compare waiver, and the tests to add. If you would rather close the gap by collecting `encoder_attention` data for a100_sxm and b60, that data has to land before this PR merges; given the legacy databases involved (trtllm 1.0.0, vllm 0.14.0, sglang 0.5.10) the code path looks like the practical route.

Minimal repro from `python/aisimulate` with the PR tools:

```
python - <<'EOF'
from tools.support_matrix.support_matrix import SupportMatrix
print(SupportMatrix.run_single_test(model="Qwen/Qwen3-VL-8B-Instruct", system="a100_sxm", backend="trtllm", version="1.0.0"))
EOF
```

python/aisimulate/tools/support_matrix/support_matrix.py
Arsene12358: Follow-up: I put the fix for B1 up as a stacked PR onto this branch so it can be merged into #85 with one click or cherry-picked as a single signed-off commit: https://github.com/ai-dynamo/aisimulate/pull/125

It adds the `ENCODER_DATA_UNAVAILABLE` preflight (FRAMEWORK_INCOMPATIBLE, replayable `generate_support_matrix.py` command, no backbone run, not retried), makes the compare gate's preflight contract and migration waiver table-driven and row-aware (a row is only accepted or waived when `_encoder_perf_data_available(system, backend, version)` is actually `False`), and pins it all with 15 tests. Verified: 188 unit tests green, the two PR e2e cases still PASS on b200_sxm, the ten a100_sxm/b60 combinations return FRAMEWORK_INCOMPATIBLE in under a second, and regenerating the checked-in matrix yields 0 blocking transitions. Happy to adjust the status choice or naming if you prefer a different shape.

Arsene12358: Update: the fix commit (`3421538d`, signed off) is now pushed directly onto this PR branch instead of the stacked PR (#125), which GitHub marked as merged once the commit landed on this branch. Head is `664b5965` + `3421538d`. Same content as described above: `ENCODER_DATA_UNAVAILABLE` preflight, row-aware compare contract and waiver, docs paragraph, 15 tests; 188 unit tests green, PR e2e cases still PASS, a100_sxm/b60 combinations return FRAMEWORK_INCOMPATIBLE, regenerated matrix yields 0 blocking transitions.

Arsene12358 (approved):
Approving now that B1 is resolved by the `fix(support-matrix)` commit on this branch and the branch is merged with current main (the `#83` Qwen3.5 encoder is recognised through its nested `vision_config`, so Qwen3.5 rows now run the image workload too) (supersedes my earlier request for changes).

Verified on the branch merged with main, fresh editable venv: `tests/unit/support_matrix` 195 passed; `tests/e2e/support_matrix/test_pr_support_matrix.py -k Qwen3-VL` 2 passed (b200_sxm/vllm still takes the image path with positive encoder evidence); real `run_single_test` for Qwen3-VL-8B and 30B-A3B across all 45 checked-in system/backend/version combinations gives PASS with image flags and silicon provenance on the 9 systems that have `encoder_attention` data and FRAMEWORK_INCOMPATIBLE / `ENCODER_DATA_UNAVAILABLE:` on a100_sxm and b60; regenerating the checked-in matrix and running `compare_support_matrix.py` as the daily workflow does yields CSV sanity clean and 0 blocking transitions (the only remaining validation errors are the range-check ones that already exist on the unmodified checked-in matrix). ruff check/format and `git diff --check` clean.

Non-blocking notes for follow-up, not required for this merge: 20 of the 75 default-roster models (Qwen3.5/3.6, Gemma-4, Llama-4, Kimi-K2.5/K2.6/K2.7/K3, MiniMax-M3, Step-3.7) now regenerate as `FAIL / ENCODER_UNSUPPORTED:` (1800 rows, 872 of them currently PASS or HYBRID_PASS, plus 350 HW_INCOMPATIBLE and 8 FRAMEWORK_INCOMPATIBLE rows preempted). The migration waiver absorbs this, but a dedicated status would avoid prefix-sniffing in the gate and the UI. The rtx_pro sm120 framework-gap test now branches on encoder coverage, so its gemma-4 and Llama-4-Scout cases no longer assert the framework gap. `test_support_matrix_docs.py` reads a CWD-relative path (fine under CI's `python/aisimulate` working directory).

PR #87, opened 2026-09-02
feat(models): model Kimi K3 image and video encoder

jasonqinzhou:
## Summary

Enable Kimi K3 image and video encoder modeling while retaining its KDA, MLA, LatentMoE and DSPARK language paths. This continues [AIConfigurator PR #1530](https://github.com/ai-dynamo/aiconfigurator/pull/1530) at source head `cdf6a077a8eaac77be7c7916c3d8b1fe4d472e02`.

- Model MoonViT3D, architecture-specific QKV geometry, spatial-temporal attention, temporal pooling, replicated PatchMergerV2 and communication.
- Apply processor budgets and independent native image/video defaults. Validate strict integer dimensions, positional-embedding constraints, the four-frame temporal limit and the actual padded 512×512 rotary grid boundary.
- Account for encoder latency, memory, energy, context tokens and TTFT. The activation estimate covers actual per-rank QKV output while retaining the existing generic estimate, projected embeddings and minimum allocation.
- Preserve visual context on `language_only=True` workers while omitting encoder execution and memory. Reject malformed outer vision metadata, including falsy non-objects, while retaining supported absent/null/empty-object configurations.
- Feed K3’s nested encoder configuration to the compatibility API/CLI EPD candidate consumer. Mark both prefill and decode EPD workers as language-only; keep the existing encoder support boundary for other model families.
- Integrate with the landed Llama 4, Kimi K2.5 and Gemma 4 shared workload/parser implementations; preserve existing positional configuration arguments through keyword-only K3 additions.

## Validation

Reviewed head: `c7b41c0f5bf5d46af9fc17d75d2986485f6b6c17`, based on main `0d5ece792dcecf1fa5da3873b292d45283ea1218`.

- 936 affected tests pass under the actual CI unit/build selector. New tests reproduce invalid scalar, positional and spatial geometry through public loaders and consumers, including language-only workers, and verify native QKV buffer accounting across TP/encoder-DP modes.
- Independent task and whole-branch re-reviews are recorded in the PR discussion. Reviewers reran tests, executed pinned upstream processor/constructor checks and verified real native static/aggregate behavior. Final outer-object, EPD consumer and decode-worker and sweep corrections were independently re-reviewed after these checks.
- Twelve public before/after cases confirm that the memory correction changes only memory outputs. The shipped four-frame 896×896 TP16/encoder-DP case changes from 110 to 158 MiB activation memory, covering its 144 MiB QKV buffer plus 14 MiB projected embeddings. Latency, energy and six language-only controls are unchanged.
- Six text graph pairs and 48 existing-model static/aggregate predictions match main, including Qwen SOL/SILICON and sibling encoder SOL controls. Geometry was independently compared against immutable upstream processor code; valid boundary controls remain accepted.
- Real K3 public EPD aggregate and disaggregated calls pass for image/video workloads, including separate encoder and LM parallelism. Real K3 configuration sweeps pass for image/video workloads with both plain configurations and callable factories. Four complete existing Qwen EPD aggregate/disaggregated result rows are exactly unchanged before/after the decoder flag correction.
- Lint, formatting, packaged legal parity and whitespace checks pass. Rust, native API/stubs, performance data and numerical goldens are unchanged from main.
- [Fast CI](https://github.com/ai-dynamo/aisimulate/actions/runs/34505710866) passed on the final commit. Formal CodeRabbit and independent Codex reviews completed on the same commit; no actionable comments remain. The remaining processor-file coverage warning was independently verified with 56 public-loader cases: 40 malformed inputs reject correctly and 16 valid/absence controls pass ([reproduction](https://github.com/ai-dynamo/aisimulate/pull/87#issuecomment-5622596010)). [Full CI](https://github.com/ai-dynamo/aisimulate/actions/runs/34506770365) passed on that exact commit: all 18 required validation jobs succeeded, with the expected manual-run release-staging skip. Both amd64 and arm64 passed native builds and application-wheel validation; each passed 6,059 selected package tests, 551 root tests and 5 compatibility tests (reported separately). [Final approval](https://github.com/ai-dynamo/aisimulate/pull/87#pullrequestreview-5170514062) covers this commit; GitHub reports the PR as approved and mergeable.

### Evidence classification and reproduction

Local evidence uses the real native extension and synthetic checkpoint/runtime inputs; it includes unit regressions, upstream-function comparisons and before/after parity. It is not fake-runner evidence or production tracing. The native Rust tree is identical to main; hosted Full CI rebuilt and tested the final source on Linux amd64 and arm64. No measured Kimi K3 production-silicon evidence is supplied, and no production accuracy is claimed.

The [public reproduction](https://github.com/ai-dynamo/aisimulate/pull/87#issuecomment-5621834064) contains the complete native script, fixture, pinned before/after revisions, machine-readable JSON output commands and numerical memory table. Its corrected revision is `55b24f2c`; the final parser-object and EPD metadata-consumer corrections do not alter valid-input geometry, operation construction or memory accounting. The independent [technical review](https://github.com/ai-dynamo/aisimulate/pull/87#pullrequestreview-5169584361) records the larger geometry/memory checks; the [final technical review](https://github.com/ai-dynamo/aisimulate/pull/87#pullrequestreview-5170053515) and [approval](https://github.com/ai-dynamo/aisimulate/pull/87#pullrequestreview-5170514062) record the final guard and EPD corrections.

<details>
<summary>Exact selected-test, lint and formatting commands</summary>

The local commands below ran from `python/aisimulate` on macOS with Python 3.13.3 and the existing dependency interpreter; the package source was explicitly set to this fresh checkout. `-p no:timeout` follows the repository's macOS guidance. With the documented package environment activated and the native extension built, the same invocations are reproducible as:

```sh
PYTHONDONTWRITEBYTECODE=1 PYTHONPATH="$PWD/src" python -m pytest -p no:timeout tests/unit/sdk/models/test_kimi_k3_vision.py tests/unit/sdk/models/test_kimi_k25_vision.py tests/unit/sdk/models/test_llama4_vision.py tests/unit/sdk/models/test_gemma4_vision.py tests/unit/sdk/backends/test_encoder.py tests/unit/sdk/backends/test_base_backend.py tests/unit/sdk/test_utils.py tests/unit/sdk/sweep/test_sweep.py tests/unit/sdk/task_v2/test_epd_real_database.py -m 'unit or build' -q --tb=short
ruff check src/aiconfigurator_core/sdk/utils.py src/aiconfigurator_core/sdk/common.py src/aiconfigurator_core/sdk/backends/base_backend.py tests/unit/sdk/models/test_kimi_k3_vision.py src/aiconfigurator/sdk/sweep.py src/aiconfigurator/sdk/task_v2.py tests/unit/sdk/sweep/test_sweep.py tests/unit/sdk/task_v2/test_epd_real_database.py
ruff format --check src/aiconfigurator_core/sdk/utils.py src/aiconfigurator_core/sdk/common.py src/aiconfigurator_core/sdk/backends/base_backend.py tests/unit/sdk/models/test_kimi_k3_vision.py src/aiconfigurator/sdk/sweep.py src/aiconfigurator/sdk/task_v2.py tests/unit/sdk/sweep/test_sweep.py tests/unit/sdk/task_v2/test_epd_real_database.py
```

From the repository root:

```sh
python3 scripts/check_packaged_legal_files.py
git diff --check 0d5ece792dcecf1fa5da3873b292d45283ea1218..HEAD
```

Hosted evidence comes only from the root `.github/workflows/fast-ci.yml` and `.github/workflows/ci.yml` workflows linked above, on the stated reviewed commit. Test totals from overlapping local reviews are not additive.

</details>

This validates architecture and analytical accounting, not measured Kimi K3 silicon accuracy. Activation memory remains an approximation with an explicit QKV buffer lower bound; missing SILICON encoder profile coverage is unchanged. Scalar token overrides retain their existing contract and do not supply a spatial aspect ratio.

Relates to [AIC-1736](https://linear.app/nvidia/issue/AIC-1736/kimi-k3-enable-image-and-video-encoder-modeling).


python/aisimulate/src/aiconfigurator_core/sdk/utils.py
```diff
@@ -693,6 +693,62 @@ def _parse_hf_config_json(config: dict) -> dict:
...
+                    f"vt_num_attention_heads ({vision_heads})"
+                )
+            merger_dim = vision_cfg["mm_hidden_size"] * merge_kernel[0] * merge_kernel[1]
+            out_hidden = vision_cfg["text_hidden_size"]
+            if out_hidden != hidden_size:
+                raise ValueError(
+                    f"Kimi-K3 vision text_hidden_size ({out_hidden}) must match "
+                    f"the language hidden_size ({hidden_size})"
+                )
+            kimi_vision_config = VisionEncoderConfig(
```
jasonqinzhou, line 1120: The Kimi K3 vision parser and operation topology are adapted from the upstream MoonViT3D / PatchMergerV2 implementation, but this port does not identify an immutable source revision or extend the packaged third-party notices. Please pin the exact Transformers/vLLM revisions beside the adapted logic and add the corresponding source/license entries to both byte-identical THIRD_PARTY_NOTICES.md copies.

jasonqinzhou: Fixed in ac05cd41: the MoonViT3D and PatchMergerV2 logic now pins immutable Transformers/vLLM revisions beside the implementation, and both packaged notices contain the corresponding source and license entries.

jasonqinzhou (commented):
Exact-head self-review after the current-main refresh: no new findings. The merge was clean; the shared engine/model/support-matrix regression suite passed locally (317 tests), and trusted Full CI is terminal green on amd64 and arm64.

python/aisimulate/src/aiconfigurator_core/sdk/models/kimi_k3.py
```diff
@@ -121,6 +122,15 @@ def __init__(self, backend_name: str, *args) -> None:
 
         self._build_context_ops()
         self._build_generation_ops()
+        if cfg.vision_config is not None:
+            self.encoder_config = cfg.vision_config
+            self.encoder_ops.extend(
```
jasonqinzhou: [P1] Honor ModelConfig.language_only when building the vision tower. A real Kimi K3 model created with language_only=True still retains 21 encoder ops / 886,202,368 bytes of encoder weights. Existing EPD language workers use this flag to omit encoder execution and memory while retaining encoder_config for visual context sizing, so this double-counts encoder work on the language worker. Gate only encoder-op construction on not language_only; keep the visual configuration. Kimi K2.5 in #84 and the existing Qwen3-VL builders already honor this contract. Add a regression for both flag states, visual-token retention, and absent encoder latency/memory.

jasonqinzhou: Fixed in da68423e. The model retains encoder_config for visual token sizing but omits the vision tower when language_only=True. The new regression failed before this fix and now passes for both worker configurations, including absent encoder latency, energy, and memory on the language-only worker.

jasonqinzhou:
Readiness workflow finding at da68423e11642027e1e661ca643da901a9ba10f7: this non-draft PR is missing the repository-required review-ready label. The direct PR-event Fast CI run therefore skips its substantive jobs and fails its aggregate gate; the separately dispatched exact-SHA Full CI is not label-gated. Applying review-ready now to admit the normal PR checks and automated review. This is workflow metadata only, not a human approval or merge-readiness claim. The exact-head Full CI run has already passed.

python/aisimulate/src/aiconfigurator_core/sdk/utils.py
```diff
@@ -753,6 +753,64 @@ def _parse_hf_config_json(config: dict) -> dict:
...
+                    f"Kimi K3 vision modeling requires merge_type='sd2_tpool', got {vision_cfg.get('merge_type')!r}"
+                )
+            if vision_cfg.get("mm_projector_type") != "patchmergerv2":
+                raise ValueError(
+                    "Kimi K3 vision modeling requires mm_projector_type='patchmergerv2', "
+                    f"got {vision_cfg.get('mm_projector_type')!r}"
+                )
+
+            vision_hidden = int(vision_cfg["vt_hidden_size"])
+            vision_heads = int(vision_cfg["vt_num_attention_heads"])
```
jasonqinzhou: [P2] Reject non-positive Kimi vision head counts at the parse boundary. A checkpoint with vt_num_attention_heads=0 currently reaches the QKV divisibility check and evaluates modulo zero, producing ZeroDivisionError instead of the typed invalid-configuration ValueError. Reproduced against this head. Validate a positive integer before arithmetic or encoder construction and cover invalid head-count values.

python/aisimulate/src/aiconfigurator_core/sdk/utils.py
```diff
@@ -753,6 +753,64 @@ def _parse_hf_config_json(config: dict) -> dict:
...
+                spatial_merge_size=merge_kernel[0],
+                out_hidden_size=out_hidden,
+                projector_dims=((merger_dim, merger_dim), (merger_dim, out_hidden)),
+                partial_rotary_factor=1.0,
+                in_channels=3,
+                qkv_hidden_size=qkv_hidden,
+                final_norm=True,
+                pool_temporal=True,
+                video_attention_type="spatial_temporal",
+                max_temporal_patches=int(vision_cfg.get("init_pos_emb_time") or 0),
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in d41fbe86. The temporal-position limit is validated as a positive integer at parse time, before encoder construction. Fourteen regressions cover missing/null/zero/negative/bool/fraction/string values on normal and language_only workers. The expanded local suite passes. @coderabbitai please re-evaluate this finding against the pushed fix and resolve if satisfied.

python/aisimulate/src/aiconfigurator_core/sdk/utils.py
jasonqinzhou: Fixed in d41fbe86. The parser rejects non-positive and non-integer head counts before division/modulo or tower construction. All six malformed-head regression cases now raise the intended ValueError, and the expanded suite passes.

python/aisimulate/tests/unit/sdk/models/test_kimi_k3_vision.py
```diff
@@ -0,0 +1,280 @@
...
+        tp_size=tp_size,
+        attention_dp_size=1,
+        moe_tp_size=tp_size,
+        moe_ep_size=1,
+        enable_encoder_dp=enable_encoder_dp,
+        nextn=nextn,
+    )
+
+
+@pytest.mark.parametrize("heads", [0, -1, False, None, 1.5, "12"])
```
jasonqinzhou: The new strict head-count regression covers False but not True. Because bool is an int subclass and False is also zero, these cases would still pass if the boolean-specific rejection were accidentally removed. Add True to protect the positive-boolean path, matching the equivalent Gemma review finding.

jasonqinzhou: Fixed in dd525dfe: the invalid-head cases now include both True and False, and all targeted model tests pass. This is test-only coverage; the strict production guard was already correct.

python/aisimulate/src/aiconfigurator_core/sdk/utils.py
```diff
@@ -753,6 +753,73 @@ def _parse_hf_config_json(config: dict) -> dict:
...
+                )
+            max_temporal_patches = vision_cfg.get("init_pos_emb_time")
+            if (
+                not isinstance(max_temporal_patches, int)
+                or isinstance(max_temporal_patches, bool)
+                or max_temporal_patches <= 0
+            ):
+                raise ValueError("Kimi K3 vision_config.init_pos_emb_time must be a positive integer")
+            merger_dim = vision_hidden * merge_kernel[0] * merge_kernel[1]
+            kimi_vision_config = VisionEncoderConfig(
```
Arsene12358, line 1120: [P1] Use Kimi processor geometry for the new K3 encoder

The new MoonViT3D configuration routes raw dimensions through Qwen nearest-stride rounding. The pinned Kimi processor used as this PR’s architecture reference pads upward after Kimi resizing. Non-aligned images therefore get incorrect projected token counts and encoder query shapes, contrary to the stated image/video accounting goal.

| Current | Required correction |
| --- | --- |
| 448 × 449 image<br>Language context: 256 tokens<br>Encoder attention: 1 × 1,024 patches | Normalize with the Kimi processor geometry.<br>Padded dimensions: 448 × 476<br>Language context: 272 tokens<br>Encoder attention: 1 × 1,088 patches<br>Add a non-aligned image case at the runtime consumer. |

Executing the exact `navit_resize` function from the PR’s pinned Kimi processor source pads 448×449 to 448×476. The actual new K3 backend returns 256 projected tokens/1,024 patches rather than 272/1,088. This concerns the newly attached K3 encoder. Its intentional four-frame limit is not part of this finding.

Reproducer (run in the installed PR environment):

```python
from aiconfigurator_core.sdk.config import RuntimeConfig
from aiconfigurator_core.sdk.backends.base_backend import BaseBackend
from aiconfigurator_core.sdk.utils import get_model_config_from_model_path
enc = get_model_config_from_model_path("moonshotai/Kimi-K3")["extra_params"].vision_config
print(BaseBackend._encoder_pre_merge_per_visual(RuntimeConfig(image_height=448, image_width=449), enc))
# Actual: (256, 1024, 1); pinned Kimi processor geometry: (272, 1088, 1)
```

Sources: [Pinned MoonViT processor math](https://github.com/huggingface/transformers/blob/cbc1651a032b923da7f4b44b3d0e6f68e6ba6b55/src/transformers/models/kimi_k25/image_processing_kimi_k25.py#L50-L91).

python/aisimulate/src/aiconfigurator_core/sdk/models/blocks/vit.py
```diff
@@ -278,3 +292,20 @@ class EncoderOnlyModel:
...
+    """Build the complete MoonViT3D + PatchMergerV2 graph for Kimi K3."""
+    if enc_cfg.encoder_type != "kimi_k3_moonvit3d_patchmergerv2":
+        raise ValueError(f"Expected Kimi K3 encoder config, got {enc_cfg.encoder_type!r}")
+    if not (enc_cfg.final_norm and enc_cfg.pool_temporal and enc_cfg.projector_post_norm):
+        raise ValueError("Kimi K3 requires complete MoonViT3D and PatchMergerV2 operation semantics")
+    if enc_cfg.video_attention_type != "spatial_temporal":
+        raise ValueError("Kimi K3 requires spatial_temporal vision attention")
+    if enc_cfg.qkv_hidden_size <= 0 or enc_cfg.max_temporal_patches <= 0:
+        raise ValueError("Kimi K3 requires explicit QKV and temporal-position geometry")
+    return build_encoder_ops(enc_cfg, tp_size, enable_encoder_dp)
```
Arsene12358, line 652: [P2] Build PatchMergerV2 with its actual normalization and TP topology

The new K3 builder delegates to the generic Qwen projector. At encoder TP2 it halves the projector GEMM dimensions and adds an all-reduce, whereas the referenced PatchMergerV2 uses replicated linears. It also inserts encoder_merger_norm even though the V2 branch has only its final RMSNorm. This fails the stated PatchMergerV2 modeling goal.

| Current | Required correction |
| --- | --- |
| K3, encoder TP2<br>fc0 (N,K): (2,048, 4,096)<br>fc1 (N,K): (7,168, 2,048)<br>Extra pre-merger norm and TP2 all-reduce<br>Final RMSNorm | Use the actual PatchMergerV2 sequence:<br>fc0 (N,K): (4,096, 4,096)<br>GELU<br>fc1 (N,K): (7,168, 4,096)<br>Final RMSNorm<br>No pre-merger norm or projector all-reduce. |

The referenced `patchmergerv2` branch constructs two `ReplicatedLinear` operations, GELU and a final RMSNorm. It returns before the other merger variant constructs `pre_norm`. Thus delegating to the generic Qwen projector adds both unsupported TP sharding/communication and an extra pre-merger normalization to newly supported K3 workloads.

Reproducer (run in the installed PR environment):

```python
from aiconfigurator_core.sdk.config import ModelConfig
from aiconfigurator_core.sdk.models import get_model
model = get_model("moonshotai/Kimi-K3", ModelConfig(tp_size=2, moe_tp_size=2, enable_encoder_dp=False), "sglang")
for op in model.encoder_ops:
    if "encoder_projector" in op._name or op._name == "encoder_merger_norm":
        print(op._name, getattr(op, "_n", None), getattr(op, "_k", None), getattr(op, "_tp_size", None))
# fc0 GEMM: N=2048, K=4096; fc1 GEMM: N=7168, K=2048.
# Includes encoder_merger_norm and encoder_projector_ar (TP2).
```

Sources: [Pinned PatchMergerV2 branch](https://github.com/vllm-project/vllm/blob/d2906091bfc579cebefe3d8e8fb9077397ce9882/vllm/model_executor/models/kimi_k25_vit.py#L895-L944).

Arsene12358 (dismissed):
Requesting changes on `dd525dfe48dbc6ea7567a3cfbc387547b5d50945` for the 2 blockers detailed in the inline comments.

- [P1] Use Kimi processor geometry for the new K3 encoder (spec).
- [P2] Build PatchMergerV2 with its actual normalization and TP topology (spec).

Independent validation: 996 selected package tests passed. Six existing text prefill/decode operation graphs and twelve Qwen SILICON-query static/aggregate predictions match base `b6b0a386`. The inline findings come from additional input/shape reproductions and pinned upstream or repository-rule checks; passing suites do not resolve them.

Only findings introduced by this PR are requested here. Existing EPD candidate-selection limitations and unmeasured GPU calibration are excluded.

Arsene12358 (commented):
Codex follow-up review of `d6b908ebf6adef3fad3986361efb0cb1e953724f`: both original blockers are fixed, along with the native processor-default issue caught by independent task review. A separate whole-branch review found no remaining blocking or deferred findings.

- 5,533 full unit tests passed (11 skipped) on this exact head; the final processor correction also passed 1,033 model/encoder/utility/support-matrix tests. Local full runs exclude the torch-dependent dispatch test and documented non-TTY plain-output tests.
- Independent task re-review and a separate whole-branch Codex review found no remaining blocking or deferred findings.
- Final independent checks include 200 geometry comparisons against freshly fetched upstream processor code, 24 static and 24 mixed native SOL dimension/token-override comparisons, and public local/remote loader controls for native and legacy metadata.
- Replicated PatchMergerV2 dimensions, absent pre-norm/projector AllReduce, final normalization, TP/DP communication and language-only behavior are covered through the actual engine consumer.
- Six language operation graphs and 24 Qwen predictions (12 SILICON and 12 configured SOL) match the pinned base exactly. Qwen cases use static/aggregate text/image/video, h200_sxm/vLLM 0.24.0, TP1, batch 4, ISL 1024 / OSL 8, 448×449 images and four 448-square video frames.
- Changed files pass lint/format/legal/whitespace checks. Direct Fast CI passed on this exact head; current-head CodeRabbit and Full CI status must be assessed separately from historical runs.

Correction to my earlier validation wording: the original Qwen probe used a raw database template with SILICON query mode, despite passing SOL as its loading-policy argument. Final validation separately reran both actual query modes. Required CODEOWNER approval remains outstanding.

python/aisimulate/src/aiconfigurator_core/sdk/utils.py
Arsene12358: Kimi resizing now applies the pinned processor budgets and ceiling padding; 448×449 produces 272 language tokens and 1,088 patches. Local/remote processor sidecars preserve independent native defaults, and unsupported mismatched image/video side limits reject explicitly while legacy shared limits remain supported. Independent upstream geometry and native runtime comparisons pass.

Verified head: `d6b908ebf6adef3fad3986361efb0cb1e953724f`; 5,533 full unit tests and independent task/whole-branch reviews pass.

python/aisimulate/src/aiconfigurator_core/sdk/models/blocks/vit.py
Arsene12358: PatchMergerV2 now keeps both full GEMM dimensions under encoder TP, omits the extra pre-normalization and projector AllReduce, and retains only the required final normalization. Actual engine tests verify weights, latency, memory and DP/TP communication; the tower QKV geometry is unchanged.

Verified head: `d6b908ebf6adef3fad3986361efb0cb1e953724f`; 5,533 full unit tests and independent task/whole-branch reviews pass.

Arsene12358 (approved):
Final Codex review of `d6b908ebf6adef3fad3986361efb0cb1e953724f` against `b6b0a38618aa59181a6bdc80ae9d6f43082c17c1`: no remaining blocking findings. Separate independent standards and functional reviews examined the complete branch and reverified the consequential runtime contracts. All previously requested blocking changes are resolved.

Independent final checks passed 77 focused tests, 200 freshly fetched upstream geometry comparisons, and 24 static plus 24 mixed native SOL counterfactuals.

Direct Fast CI, CodeRabbit, and [Full CI on amd64 and arm64](https://github.com/ai-dynamo/aisimulate/actions/runs/34449126144) passed on this exact head. This is architecture and analytical consistency evidence, not measured GPU predictive-accuracy validation.

Nonblocking P3: synchronize the existing provenance references in vit.py and utils.py into file headers as AGENTS.md requests. Complete immutable sources, owners, licenses and modification notices are already preserved in the packaged notices; this is a placement deviation, not missing attribution.

Approved with that nonblocking note. Merge the overlapping sibling PRs sequentially and revalidate the remaining heads after each merge.

Arsene12358:
@coderabbitai review

Please review current head `1796b59381be70dc51f01327bd7185972d6c1f0f` after merging main `0d5ece792dcecf1fa5da3873b292d45283ea1218`. The resolution preserves Kimi K3 processor defaults, temporal limits, PatchMergerV2 topology and language-only behavior alongside Llama 4, Kimi K2.5 and Gemma 4. Five K3-specific config additions are keyword-only to preserve existing positional constructors.

Local validation: 675 relevant tests passed; 64 native static and 192 mixed calls per revision match the previously approved K3 head. Six text graph pairs and 48 existing-model static/aggregate predictions match main. The intervening main update contains only three README/FPM documentation files; the complete PR-specific diff is unchanged. This is analytical/accounting validation, not measured Kimi K3 silicon accuracy.

Arsene12358 (commented):
Codex final technical review of `1796b59381be70dc51f01327bd7185972d6c1f0f` against main `0d5ece792dcecf1fa5da3873b292d45283ea1218`: no remaining blockers or deferred findings. Independent standards, task, and whole-branch reviews are clean. The final main documentation refresh leaves the complete PR-specific diff unchanged.

Validation includes 675 implementation tests, independent 398-test task and 199-test final review suites, freshly fetched upstream geometry comparisons, actual native static/aggregate/mixed consumer checks, temporal/input guards, and sibling regression controls. K3 matches the previously approved branch across 64 static and 192 mixed calls per revision; six text graph pairs and 48 existing-model static/aggregate predictions match main. Existing positional constructors and native public API contracts remain intact.

Direct Fast CI passed on this head. Fresh CodeRabbit and exact-head Full CI remain separate gates before renewed approval. This is architecture and analytical accounting validation, not measured K3 silicon accuracy.

Arsene12358 (changes requested):
Requesting changes on `1796b59381be70dc51f01327bd7185972d6c1f0f` for a confirmed P2: unsupported Kimi K3 vision geometry reaches native predictions.

- Require strict positive integer dimensions. The public local loader accepts depth `0`, producing zero transformer QKV latency, and converts `True` into one layer. On B200/TRT-LLM/current configured SOL, TP1, batch 1, ISL 128/OSL 1, a 448×448 image costs 0.7053777935 ms with the shipped 27 layers, 0.0172670338 ms with depth 0, and 0.0427526175 ms with depth True. Malformed configurations must fail before construction, including language-only workers.
- Enforce actual positional-embedding constraints. QKV width 1560 with 12 heads gives head width 130, rejected by upstream 2D rotary embedding because it must be divisible by 4. Tower width 1025, with matching `mm_hidden_size`, is rejected by upstream temporal sinusoidal initialization because tower width must be even. Public predictions currently accept both. Retain valid controls such as head width 132 and tower width 1026, and the supported absent/null QKV fallback.
- Validate the actual padded K3 spatial grid against the upstream 512×512 rotary table. A local processor override `patch_limit_on_one_side=1024` and image 28×14336 produces a 2×1024 patch grid; native prediction succeeds, but upstream lookup rejects it. Validate after merge padding, which can itself cross the boundary. Preserve a larger processor ceiling when the actual image still fits, and keep unrelated Kimi K2.5 behavior unchanged.

The unconditional and runtime constraints were checked against the immutable [vLLM MoonViT implementation](https://github.com/vllm-project/vllm/blob/d2906091bfc579cebefe3d8e8fb9077397ce9882/vllm/model_executor/models/kimi_k25_vit.py), with its validation paths independently executed. These are PR-introduced missing-validation cases, not a measured silicon-accuracy claim. They remain one geometry-validation blocker and are being corrected before independent re-review and approval.

Arsene12358 (changes requested):
Requesting changes for a second confirmed P2 on `1796b59381be70dc51f01327bd7185972d6c1f0f`: encoder activation memory ignores the new K3 QKV width.

Independent public-consumer verification found that the shipped K3 configuration (QKV hidden width1536), four896×896 frames, B200/TRT-LLM/current configured SOL, TP16 with encoder DP, batch1, reports110MiB encoder activations. Its BF16 QKV output tensor alone requires144MiB (16,384 patch rows ×4,608 QKV columns ×2 bytes). The complete worker estimate is101.013GiB, so this is reachable in a fitting deployment. Valid QKV width3072 doubles that buffer while leaving the activation estimate unchanged.

The earlier reviews dismissed the inherited formula as approximate; that does not justify falling below a newly introduced, explicitly modeled buffer requirement. Preserve the existing generic approximation and projected-embedding buffer, but ensure the activation width covers actual per-rank QKV output, respecting encoder DP/TP. Add native-consumer width/parallelism regressions and explain expected before/after memory changes.

This correction and strict geometry validation are being implemented in a single fix wave, followed by independent re-review. Existing-model defaults and predictions should remain unchanged.

Arsene12358:
@coderabbitai review

Both requested P2 corrections are implemented on `55b24f2cc8052f9711a66e21f316b23c4b00e18a`. Please review this revision. The K3 parser now requires strict positive geometry and the upstream positional-width constraints; the runtime rejects padded K3 grids beyond the 512×512 rotary table. Valid absent/null QKV defaults, valid boundary widths, language-only behavior, and unaffected sibling models are retained. The generic activation estimate now covers the actual per-rank packed QKV output, while retaining the legacy estimate, projected-embedding buffer and 32 MiB minimum.

The new regression cases first reproduced failures on the old implementation and then passed after correction. Independent re-review results are reported separately. The following public reproduction supplies commands, fixtures, pinned revisions and numerical outputs for the memory correction; it does not claim measured silicon accuracy.

### Before/after evidence

Baseline: `1796b59381be70dc51f01327bd7185972d6c1f0f`. Corrected head: `55b24f2cc8052f9711a66e21f316b23c4b00e18a`. Both use the same performance-data tree `4bda4b53d1a6f939217c429a11fd428911674b77` and unchanged Rust implementation. The fixture starts from the bundled `moonshotai/Kimi-K3` configuration; only the QKV width is varied. Runtime: B200/TRT-LLM/current, explicitly configured SOL, ISL 128, OSL 1, four 896×896 frames, one video per request. Geometry is 16,384 pre-pool patch rows and 1,024 output tokens per video.

| QKV width | Model TP | Encoder DP | Batch | Packed QKV MiB/rank | Activations before MiB | Activations after MiB | Worker memory before → after GiB |
| --- | --- | --- | --- | --- | --- | --- | --- |
| 1536 | 16 | True | 1 | 144 | 110 | 158 | 101.012592793 → 101.059467793 |
| 1536 | 2 | True | 3 | 288 | 234 | 330 | 700.553997993 → 700.647747993 |
| 1536 | 2 | False | 3 | 216 | 330 | 330 | 700.278607368 → 700.278607368 |
| 3072 | 16 | True | 1 | 288 | 110 | 302 | 101.328999043 → 101.516499043 |
| 3072 | 2 | True | 3 | 576 | 234 | 618 | 700.870404243 → 701.245404243 |
| 3072 | 2 | False | 3 | 432 | 330 | 474 | 700.436810493 → 700.577435493 |

The shipped TP16/encoder-DP case is a fitting deployment: the old 110 MiB activation estimate was below its single 144 MiB packed QKV buffer; the corrected 158 MiB includes that buffer plus the existing 14 MiB projected embeddings. TP2 cases are shape/accounting controls, not deployment recommendations. All native latency and energy values, and all non-memory result fields, match before/after exactly after excluding the temporary local checkpoint path. Six corresponding language-only cases remain unchanged and allocate no encoder memory. A separate rerun of 48 existing-model static/aggregate predictions and six text graph pairs matches main exactly.

<details>
<summary>Exact native reproduction script and commands</summary>

Save the following as `/tmp/k3-memory-probe.py`. It records complete public native results, including latency, energy, weights, activation memory and worker memory, in JSON.

```python
"""Public local-checkpoint/native prediction evidence for the K3 QKV fix."""
import json
import sys
from copy import deepcopy
from pathlib import Path
from tempfile import TemporaryDirectory

from aiconfigurator_core.sdk.config import ModelConfig, RuntimeConfig
from aiconfigurator_core.sdk.models import get_model
from aiconfigurator_core.sdk.utils import get_model_config_from_model_path
from aiconfigurator_core.sdk.backends.factory import get_backend
from aiconfigurator_core.sdk.perf_database import get_database_view

ROOT = Path(__file__).parent
raw = get_model_config_from_model_path('moonshotai/Kimi-K3')['raw_config']
db = get_database_view('b200_sxm', 'trtllm', 'current', database_mode='SOL', allow_missing_data=True)
records = []
for qkv in [1536, 3072]:
    for tp, dp, batch in [(16, True, 1), (2, True, 3), (2, False, 3)]:
        for language in [False, True]:
            config = deepcopy(raw)
            config['vision_config']['qkv_hidden_size'] = qkv
            with TemporaryDirectory(dir=ROOT) as checkpoint:
                Path(checkpoint, 'config.json').write_text(json.dumps(config))
                model = get_model(checkpoint, ModelConfig(tp_size=tp, moe_tp_size=tp, moe_ep_size=1,
                                  enable_encoder_dp=dp, language_only=language), 'trtllm')
                runtime = RuntimeConfig(batch_size=batch, isl=128, osl=1, num_images_per_request=0,
                                        num_videos_per_request=1, video_frames=4,
                                        video_height=896, video_width=896)
                summary = get_backend('trtllm').run_static(model, db, runtime, mode='static_ctx')
                # 896 / 14 = 64 patches per side, four frames, then temporal
                # pooling and 2x2 spatial merging yield 1024 language tokens.
                patch_rows = 16384 * ((batch + tp - 1) // tp if dp else batch)
                qkv_columns = 3 * qkv // (1 if dp else tp)
                records.append(dict(qkv=qkv, tp=tp, encoder_dp=dp, batch=batch, language_only=language,
                    patch_rows=patch_rows, qkv_columns=qkv_columns,
                    qkv_tensor_MiB=2 * patch_rows * qkv_columns / 2**20,
                    result=summary.get_result_dict(), encoder_memory=summary.get_encoder_memory(),
                    encoder_latency=summary.get_encoder_latency_dict(),
                    encoder_energy=summary.get_encoder_energy_wms_dict()))
Path(sys.argv[1]).write_text(json.dumps(records, indent=2, sort_keys=True, default=str) + '\n')
print(f'{len(records)} public native predictions recorded in {sys.argv[1]}')

```

From a clone containing both revisions, create independent environments using the repository's documented setup:

```sh
git worktree add --detach /tmp/k3-before 1796b59381be70dc51f01327bd7185972d6c1f0f
git worktree add --detach /tmp/k3-after 55b24f2cc8052f9711a66e21f316b23c4b00e18a
uv sync --project /tmp/k3-before/python/aisimulate --extra dev
uv sync --project /tmp/k3-after/python/aisimulate --extra dev
PYTHONPATH=/tmp/k3-before/python/aisimulate/src /tmp/k3-before/python/aisimulate/.venv/bin/python /tmp/k3-memory-probe.py /tmp/k3-before.json
PYTHONPATH=/tmp/k3-after/python/aisimulate/src /tmp/k3-after/python/aisimulate/.venv/bin/python /tmp/k3-memory-probe.py /tmp/k3-after.json
```

The output JSON supplies per-case values; the only expected result changes are `memory` and `encoder_memory`. The temporary local checkpoint `model` path is non-deterministic and should be excluded when comparing. The test suite also exercises actual native static/aggregate QKV shapes and language-only memory subtraction under the CI `unit or build` selection.

</details>

The positional constraints are checked against the pinned [vLLM MoonViT implementation](https://github.com/vllm-project/vllm/blob/d2906091bfc579cebefe3d8e8fb9077397ce9882/vllm/model_executor/models/kimi_k25_vit.py). Processor geometry was independently compared with freshly fetched, unchanged `navit_resize` functions from the pinned [Transformers image processor](https://github.com/huggingface/transformers/blob/cbc1651a032b923da7f4b44b3d0e6f68e6ba6b55/src/transformers/models/kimi_k25/image_processing_kimi_k25.py) and [video processor](https://github.com/huggingface/transformers/blob/cbc1651a032b923da7f4b44b3d0e6f68e6ba6b55/src/transformers/models/kimi_k25/video_processing_kimi_k25.py). Independent task review executed 960 geometry comparisons; the bounded fixes preserve supported geometry and add rejection of upstream-invalid cases.

This is reproducible analytical/accounting evidence. SOL energy can legitimately be zero; missing SILICON encoder profiles and measured K3 GPU calibration remain outside these claims.

Arsene12358 (commented):
Both requested P2 blockers are resolved on `55b24f2cc8052f9711a66e21f316b23c4b00e18a`: strict K3 geometry now includes positional widths and the actual padded rotary-grid boundary; activation memory now covers actual per-rank packed QKV output. Valid optional-QKV defaults, boundary geometries, language-only context, and sibling behavior are preserved.

The independent task re-review passed 824 affected tests plus upstream-derived public boundary checks and unchanged sibling predictions. The independent final whole-branch re-review passed 533 selected tests, executed pinned upstream constructor constraints, and verified 48 spatial cases including 32 real native valid predictions. Independent native before/after memory checks confirm the shipped case changes from 110 to 158 MiB activations, covering its 144 MiB QKV buffer; latency/energy and language-only outputs remain unchanged. Counts overlap and are not additive. Lint, formatting, legal parity, whitespace and signed commits pass. No unresolved technical or deferred findings remain.

Fresh external review and exact-commit hosted CI are pending. The earlier REQUEST_CHANGES reviews will be superseded by approval only after those gates pass.

python/aisimulate/src/aiconfigurator_core/sdk/utils.py
```diff
@@ -1055,6 +1056,88 @@ def _parse_hf_config_json(config: dict) -> dict:
...
+            # Kimi K3 reuses the MoonViT3D tower and PatchMergerV2 implemented
+            # for Kimi K2.5. Sources: Transformers commit
+            # cbc1651a032b923da7f4b44b3d0e6f68e6ba6b55 and vLLM commit
+            # d2906091bfc579cebefe3d8e8fb9077397ce9882.
+            def positive_int(value, name):
+                if not isinstance(value, int) or isinstance(value, bool) or value <= 0:
+                    raise ValueError(f"Kimi K3 vision_config.{name} must be a positive integer")
+                return value
+
+            merge_kernel = vision_cfg.get("merge_kernel_size")
```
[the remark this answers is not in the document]
Arsene12358: Fixed in `3d16594c21645e08d2abe3113cf3fdd73641fd61` with a dictionary guard before field access. The focused public-loader regressions reproduce the old AttributeError and now receive a clear ValueError; supported missing/null/empty metadata and valid configurations retain their behavior. The840-test affected suite and independent re-reviews pass.

Arsene12358 (commented):
The final P3 parser-object follow-up is independently clean on `3d16594c21645e08d2abe3113cf3fdd73641fd61`. The production change is a dictionary guard before K3 field access; malformed truthy metadata gets a clear ValueError, while existing absent/null/empty and valid model behavior is preserved. Sixteen new public-loader cases cover both language-only settings. The complete affected suite passes840 tests. Independent reviewers rechecked the focused cases and verified that all geometry, operation, runtime and memory code is unchanged from the prior reviewed revision, preserving its numerical evidence. No remaining technical or deferred findings.

Both prior P2s remain resolved. Exact-current-commit external review and hosted CI will precede final approval.

Arsene12358:
@coderabbitai review

The remaining minor parser-object finding is fixed on `3d16594c21645e08d2abe3113cf3fdd73641fd61`: a truthy non-dictionary `vision_config` now raises a clear ValueError before field access. Valid dictionaries and supported missing/null/empty behavior are preserved. Both independent re-reviews are clean, and the seven-file affected suite passes 840 tests on this revision. Lint, formatting and whitespace also pass. The final diff is two files: three production lines and sixteen parameterized regression cases. Geometry, native operation construction and memory accounting are unchanged from the reviewed `55b24f2c`.

The PR description now contains the exact selected-test/lint/format commands, exact revisions and evidence classification. Local native tests use synthetic checkpoints/runtime inputs; before/after and upstream-function comparisons establish analytical accounting and parity. No fake-runner or production-silicon evidence is claimed. The complete public native reproduction and numeric table remain in https://github.com/ai-dynamo/aisimulate/pull/87#issuecomment-5621834064. Fresh Fast CI and post-review Full CI are tracked separately on the current commit. Please verify the small correction and updated evidence description.

Arsene12358 (changes requested):
Requesting changes on `3d16594c21645e08d2abe3113cf3fdd73641fd61` for a remaining P2 in K3 metadata validation: falsy non-object `vision_config` values bypass the dictionary guard and silently disable vision accounting. Public local checkpoints containing false,0,an empty string,or an empty array all load and return actual native predictions for an image workload with zero encoder operations/latency. The shipped dictionary control builds20 encoder operations. Reproduction uses B200/TRT-LLM/current configured SOL, TP16/encoder-DP, batch1, ISL128/OSL1, one448-square image; both colocated and language-only modes reach the defect.

Move object validation before the truthiness branch, preserving missing/None/empty-dictionary absence while rejecting malformed non-dictionary values. Add these negative cases through the public loader/consumer. The earlier strict scalar, positional and grid corrections remain correct; this closes the malformed outer-object gap. The new K3 vision contract makes silently treating malformed metadata as a text-only model incorrect. This is analytical prediction evidence, not measured silicon accuracy.

Arsene12358 (changes requested):
Requesting changes on `3d16594c21645e08d2abe3113cf3fdd73641fd61` for a confirmed P2: the EPD sweep does not consume the K3 vision configuration introduced by this PR. At `python/aisimulate/src/aiconfigurator/sdk/sweep.py:859–862`, `_get_encoder_worker_candidates` requires extra_params itself to be VisionEncoderConfig; K3 correctly retains its language metadata in KimiK3Config and nests vision_config. The public compatibility API `Task.run_single_agg` with enable_epd therefore rejects the bundled moonshotai/Kimi-K3 as having no vision encoder.

Independent actual-native reproduction: B200/TRT-LLM/current configured SOL, TP16/moeTP16,batch1,encoderTP1,224-square image. Both direct candidate construction and the public Task call raise the false no-vision error. A bounded in-process extraction-only counterfactual returns real encoder candidates for TP1/2/4 and a successful E+agg result. Main has no K3 vision contract, so its prior rejection was expected; this PR adds that contract without wiring this consumer. The unified CLI routes EPD users to the supported compatibility CLI/API, so its own EPD restriction does not exclude this caller.

Add a K3-specific extraction fallback while retaining existing direct-config behavior. Avoid broadly admitting other nested families through the generic encoder builder: Llama4 requires its own operation topology. Add real bundled-K3 candidate and public Task EPD tests, with no candidate-function mock, plus a control preserving unsupported sibling behavior. This is analytical consumer validation, not a measured silicon-accuracy claim.

Independent final review of local correction `2ff8329a087a1112f08a8c5e7b6bbefbd8ee11fc` confirms the candidate/E+agg fix but finds one remaining E+P+D caller gap: `Task.run_single_disagg` marks prefill as language_only under EPD, but leaves the decode ModelConfig unchanged. K3 LM TP16 then constructs an unused decoder encoder and fails because its12 vision heads do not divide16. A real B200/TRTLLM/SOL public call with prefill/decode TP16, decode batch2 and encoder TP4/batch2 reproduces this before reaching candidate generation. An isolated counterfactual setting only decode.language_only=True makes the same call succeed (one E/P/D worker each,36 GPUs,0.417ms encoder latency,31.2ms TTFT). Mirror the existing prefill flag for the EPD decode worker and add an actual disaggregated K3 public regression. This extends the same open EPD integration finding; all branches of the supported EPD API must work before approval.

The same decode-worker omission also exists in the supported configuration-search entry point `sweep_disagg`: only prefill_model_config receives `_language_only_config`. On local `1f06cd55`, independent real native sweeps with both plain ModelConfig and the actual Task-supplied decoder factory fail at the same12-head/LM-TP16 check. Applying only the existing language-only wrapper to decode yields a valid36-GPU row. The correction therefore covers both single-point Task E+P+D and sweep E+P+D, with image/video and plain/factory controls. This remains the same EPD integration blocker, not a separate modeling formula change.

python/aisimulate/src/aiconfigurator_core/sdk/utils.py
```diff
@@ -1055,6 +1056,91 @@ def _parse_hf_config_json(config: dict) -> dict:
                 "silently dropped and produce a wrong hybrid layer plan."
             )
         layer_types = tuple("linear_attention" if (i + 1) in kda_layer_ids else "full_attention" for i in range(layers))
+        kimi_vision_config = None
+        if vision_cfg:
+            if not isinstance(vision_cfg, dict):
```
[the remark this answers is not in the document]
Arsene12358: Fixed in `c7b41c0f5bf5d46af9fc17d75d2986485f6b6c17`: object validation now precedes truthiness and rejects falsy non-objects. Missing/None/empty-dictionary absence is preserved. Ten new falsy public-loader cases first reproduced the failure, then passed; the full936-test affected suite and independent re-reviews are clean.

Arsene12358 (commented):
Independent task and final whole-branch re-reviews are clean on `c7b41c0f5bf5d46af9fc17d75d2986485f6b6c17`. Both current-head REQUEST_CHANGES findings are resolved: the parser rejects falsy malformed vision objects while preserving supported missing/null/empty-object metadata; the supported compatibility EPD consumer now reads K3 nested vision config with a K3-specific extraction fallback, and both prefill and decode workers omit the separately hosted encoder in single-point and sweep EPD paths. Direct Qwen behavior and unsupported nested Llama4 behavior remain unchanged.

The nine-file affected suite passes936 tests, including actual K3 encoder candidates and public Task E+agg/E+P+D image/video paths through native SOL. Independent review reran the prior failures and valid/sibling controls. Root reran both fixed real EPD consumers; their complete outputs match the independent extraction-only counterfactual. All geometry, operation builders, runtime, memory and native sources outside the bounded corrections are unchanged, so prior accounting/boundary evidence remains applicable. Lint, formatting, legal parity and whitespace pass; no remaining technical or deferred findings.

Fresh exact-head CodeRabbit review, Fast CI and Full CI remain the final gates before approval.

Arsene12358:
@coderabbitai review

Both findings from the last review are fixed on `c7b41c0f5bf5d46af9fc17d75d2986485f6b6c17`. The K3 outer-object check now runs before truthiness: missing/null/empty-dictionary metadata remains supported, while false, zero, empty strings, arrays and all other non-null non-objects fail clearly. The EPD consumer now unwraps KimiK3Config through the existing helper. This fallback is K3-specific because other nested families such as Llama4 require specialized encoder builders; existing direct-config models and unsupported sibling boundaries are preserved.

The new public regressions first reproduced12 failures (ten falsy cases and two real K3 EPD caller failures), with18 controls passing. After correction, all936 affected tests pass across the nine-file suite, including real native K3 encoder candidates, public Task EPD, Qwen controls and retained Llama4 rejection. Both independent re-reviews are clean. Exact commands, head and analytical/synthetic/parity classification are in the PR description. These changes do not alter the previously reviewed geometry, operation builders, native engine, data or QKV memory formula.

The final whole-branch review also exercised E+P+D and found that the decode worker still built an unused encoder. The final commit mirrors the existing prefill `language_only` flag for the EPD decode worker. Actual K3 image/video E+P+D tests now pass at language TP16 and encoder TP4; four full existing Qwen EPD result rows are exactly unchanged. The same missing decode-worker wrapper was corrected in `sweep_disagg`; four native sweep regressions cover image/video and plain/callable configurations. The actual Task-supplied factory is independently exercised. The bounded role audit confirms both language workers omit the encoder in the single-point and sweep EPD paths. Both independent final re-reviews include these corrections.

### Public EPD before/after reproduction

Baseline `3d16594c21645e08d2abe3113cf3fdd73641fd61` raises a false "has no vision encoder" ValueError. Corrected head `c7b41c0f5bf5d46af9fc17d75d2986485f6b6c17` returns one encoder worker and one aggregate worker, encoder TP1, encoder latency0.223ms and composed TTFT46.451706742857134ms. This is actual native analytical SOL with synthetic workload inputs, not measured silicon evidence. Run the following in either revision's documented package environment:

```python
from aiconfigurator.sdk.task_v2 import Task

task = Task(
    serving_mode="agg", model_path="moonshotai/Kimi-K3",
    system_name="b200_sxm", backend_name="trtllm", database_mode="SOL",
    enable_epd=True, isl=128, osl=2, image_height=224, image_width=224,
    num_images_per_request=1, engine_step_backend="rust",
)
result = task.run_single_agg(tp=16, moe_tp=16, batch_size=1, encoder_tp=1)
print({key: result[key] for key in ["(e)workers", "(a)workers", "encoder_latency", "ttft"]})
```

The actual encoder-candidate consumer also now returns TP1/2/4 candidates with latency0.223/0.161/0.130ms and encoder memory0.857/0.487/0.303GiB for this image/batch. Root reran both actual fixed consumers and their outputs exactly match the independently isolated extraction-only counterfactual from the failing revision. The regression suite exercises the actual candidate function and native engine, without substituting a fake candidate result.

The previous complete [QKV-memory reproduction](https://github.com/ai-dynamo/aisimulate/pull/87#issuecomment-5621834064) remains applicable: the memory/runtime implementation is unchanged by these metadata-consumer corrections. Fresh Fast CI and post-review Full CI will be recorded on the final commit. Please review this correction and resolve the remaining falsy-metadata thread.

Arsene12358:
@coderabbitai review

Arsene12358:
The remaining CodeRabbit Cross-Layer Contract warning is an evidence-only suggestion, not a failing product path. Independent verification on `c7b41c0f5bf5d46af9fc17d75d2986485f6b6c17` exercised 56 cases through the actual public loader: both processor filenames, local files and mocked remote transport, ten malformed scalar/list values, and missing/null/empty-object/populated-object controls. All40 malformed cases raise the filename-specific ValueError; all16 valid/absence controls preserve K3 encoder metadata and independent image/video budgets. Parsing and validation are real; only remote transport is mocked. This is loader evidence, not network/authentication testing or a native/silicon claim.

The relevant non-null/dictionary guard is unchanged from main, and the K3 architecture route reaches it correctly. No product defect was found. A checked-in matrix would add coverage, but no corrective source change is needed to close this review. Both independent code reviews remain clean, and Full CI is running on the same commit.

<details>
<summary>Complete public-loader reproduction</summary>

Run this script in an empty working directory using the documented project environment. It writes `processor-file-review.json` with every input, route, output/error and count; stdout reports56 total,40 rejected and16 accepted.

```python
import json,logging,tempfile
from copy import deepcopy
from pathlib import Path
from unittest.mock import patch
logging.disable(logging.CRITICAL)
from aiconfigurator_core.sdk import utils
raw=deepcopy(utils.get_model_config_from_model_path('moonshotai/Kimi-K3')['raw_config'])
rows=[];count=0
cases=[('invalid',x) for x in (True,False,0,0.0,1,1.5,'','vision',[],[1])]+[('missing',None),('null',None),('empty',{}),('populated',{'max_patches':1024})]
for remote in (False,True):
    for filename in ('preprocessor_config.json','video_preprocessor_config.json'):
        for kind,value in cases:
            count+=1;files={'config.json':deepcopy(raw)}
            other='video_preprocessor_config.json' if filename=='preprocessor_config.json' else 'preprocessor_config.json';files[other]={}
            if kind!='missing':files[filename]=deepcopy(value)
            row=dict(route='mocked_remote' if remote else 'local',filename=filename,kind=kind,value=value)
            with tempfile.TemporaryDirectory() as td:
                if remote:model_path=f'synthetic-kimi-k3/sidecar-review-{count}'
                else:
                    model_path=td
                    for name,payload in files.items():Path(td,name).write_text(json.dumps(payload))
                downloads=[]
                def download(path,name,**kwargs):downloads.append(name);return deepcopy(files.get(name))
                with patch.object(utils,'_download_hf_config',lambda path:deepcopy(raw)),patch.object(utils,'_download_hf_json',download):
                    try:
                        info=utils.get_model_config_from_model_path(model_path);enc=utils.get_vision_encoder_config_from_model_info(info)
                        row.update(accepted=True,encoder_type=enc.encoder_type,image_max_patches=enc.image_max_patches,video_max_patches=enc.video_max_patches)
                        assert kind!='invalid',row
                        assert enc.encoder_type=='kimi_k3_moonvit3d_patchmergerv2'
                        assert enc.image_max_patches==(1024 if filename=='preprocessor_config.json' and kind=='populated' else 16384)
                        assert enc.video_max_patches==(1024 if filename=='video_preprocessor_config.json' and kind=='populated' else 4096)
                    except ValueError as exc:
                        row.update(accepted=False,error_type=type(exc).__name__,error=str(exc));assert kind=='invalid',row;assert str(exc)==f'Kimi {filename} must be an object'
                if remote:
                    assert filename in downloads;row['downloaded_files']=downloads
                else:assert not downloads
            rows.append(row)
result=dict(head='c7b41c0f5bf5d46af9fc17d75d2986485f6b6c17',procedure='Actual public get_model_config_from_model_path; local disk JSON and transport-only _download_hf_config/_download_hf_json replacements for simulated remote data. Parsing and validation remain real. No network/authentication or native engine claims.',counts=dict(total=len(rows),rejected=sum(not row['accepted'] for row in rows),accepted=sum(row['accepted'] for row in rows)),cases=rows)
Path('processor-file-review.json').write_text(json.dumps(result,indent=2));print(json.dumps(result['counts']));print('Every malformed scalar/list sidecar rejected with filename-specific ValueError; absent/null/empty/valid object controls preserved.')

```

</details>

Arsene12358 (approved):
Final review of `c7b41c0f5bf5d46af9fc17d75d2986485f6b6c17`: no remaining merge blockers. The main-branch conflicts are resolved, and the confirmed metadata/geometry, QKV activation-memory and EPD integration findings are fixed. Independent task and whole-branch reviewers reverified the final changes, including actual native image/video EPD predictions and configuration sweeps.

Validation: 936 affected local tests passed. [Fast CI](https://github.com/ai-dynamo/aisimulate/actions/runs/34505710866) and [Full CI](https://github.com/ai-dynamo/aisimulate/actions/runs/34506770365) passed on this exact commit, including native builds and wheel validation on amd64 and arm64. Each architecture passed 6,059 selected package tests, 551 root tests and 5 compatibility tests; these overlapping suites are reported separately. Full CI completed with 18 successful jobs and the expected manual-run release-staging skip.

The formal CodeRabbit review covers this commit, with no actionable comments and all nine threads resolved. Its remaining processor-file coverage suggestion was independently checked with [56 public-loader cases](https://github.com/ai-dynamo/aisimulate/pull/87#issuecomment-5622596010); all behaved correctly and no product defect was found. Approval covers software behavior and analytical modeling, without claiming measured Kimi K3 silicon accuracy.

PR #89, opened 2026-09-02
feat(moe): add model expert popularity collector and database

Harrilee:
## What this adds

A production-oriented SGLang collector and packaged database for model-specific routed-expert popularity:

- 15 MoE models, 809 MoE layers, and 197,024 layer/expert rows
- original per-layer/per-expert activation counts in Parquet, with validated YAML metadata
- native fused routing observation; the collector does not replace the serving backend or recompute top-k in Python
- immutable model/checkpoint, tokenizer, image, collector, and validation provenance

## Reproducibility and input-sequence stability

The collection separates two different reproducibility questions:

1. **Repeat reproducibility:** each shard is run twice with identical token IDs. Eleven models are bit-for-bit exact. The four models with native kernel nondeterminism pass the release gate of Pearson `>= 0.999` and JSD `<= 0.001` bits.
2. **Input-sequence robustness:** every model uses four independently seeded random input-sequence shards. For every shard pair, we compare each MoE layer's hottest expert. The visualizer's **Mean cross-sequence top-1 stability** is the mean fraction of layers whose top-1 expert ID is unchanged.

Every published model exceeds **90% mean cross-sequence top-1 stability** (range **90.31%–100%**). This means the hottest expert is reproducible across changed input sequences for the great majority of layers, rather than being an artifact of one workload sample. The model detail view also retains the more conservative worst-pair result. For transparency, DeepSeek-V4-Flash has a 90.31% mean and an 88.37% worst pair because several near-tied experts exchange first place; this is not a claim that every layer is 100% invariant.

Pearson and JSD additionally verify that the full popularity distribution remains close across shards. These metrics do not yet guarantee the complete hot-to-cold ordering of the low-count tail.

## Visualizer

> [!IMPORTANT]
> The interactive visualizer is temporarily hosted on an NVIDIA Internal-only page. The screenshot below is stored directly in this GitHub PR branch so it renders independently of the internal GitLab URL.

[Open the NVIDIA Internal Expert Popularity Database visualizer](https://gitlab-pages-72a7bf.gitlab-master-pages.nvidia.com/expert-popularity-database.html)

Expert Popularity Database with cross-sequence top-1 stability

## Validation

- [Full CI: 14/14 jobs passed](https://github.com/ai-dynamo/aisimulate/actions/runs/33688677562)
- 123 targeted collector, SDK, package-data, and compatibility tests passed
- native-path cluster smokes passed for standard TopK, FP8, MXFP4, and DeepSeek-V4 HashTopK + fused learned routing
- no WideEP collector or `compile_engine.json` changes

This supersedes ai-dynamo/aiconfigurator#1596.


Harrilee:
Full CI completed successfully on head `ebcd7409fc3ee251c440c5d6a728949847cc201b`: all 14 amd64/arm64 jobs passed, including application tests, wheel builds, release artifact contracts, lint, Rust, public API, and cargo-deny. Run: https://github.com/ai-dynamo/aisimulate/actions/runs/33680832246

Harrilee:
Final validation for `976630c5b15019c4d443c27ae59a3b30742721c6`:

- [Full CI: 14/14 jobs passed](https://github.com/ai-dynamo/aisimulate/actions/runs/33688677562) on the exact PR head (amd64 + arm64 application tests, wheels, release artifact contract, Rust, lint, and dependency policy).
- 123 targeted collector, publisher, SDK, package-data, compatibility, and import-contract tests passed locally.
- The built wheel contains exactly 15 expert-popularity metadata YAML files and 15 Parquet files.
- All 15 packaged bundles validate: 809 MoE layers and 197,024 layer/expert rows; no dense-layer rows, old replication fields, `alpha_exp`, or cluster identity are published.
- Native serving-path smokes passed for standard unquantized TopK, FP8 FlashInfer TRT-LLM routing, MXFP4 FlashInfer routing, and DeepSeek-V4 HashTopK + fused learned routing. Recording-off/on requests produced identical output tokens, all recorder endpoints returned HTTP 200, assignment conservation passed, and repeat stability passed.
- DeepSeek-V4-Flash was recollected through the native recorder. Its repeat minimum Pearson is `0.999993262` and maximum JSD is `0.000024799` bits.
- The NVIDIA Internal-only visualizer and English preview were refreshed to the new database.

python/aisimulate/collector/expert_popularity/verify_routing_equivalence.py
```diff
@@ -0,0 +1,328 @@
...
+    if not isinstance(canonical_map, dict) or not isinstance(checkpoint_map, dict):
+        raise TypeError("safetensors index is missing weight_map")
+    if set(canonical_map) != set(checkpoint_map):
+        raise RuntimeError("canonical and checkpoint tensor key sets differ")
+    if canonical_index.get("metadata", {}).get("total_size") != checkpoint_index.get("metadata", {}).get("total_size"):
+        raise RuntimeError("canonical and checkpoint safetensors total_size differ")
+    router_keys = sorted(key for key in canonical_map if _ROUTER_KEY.search(key))
+    if not router_keys:
+        raise RuntimeError("no router tensors found")
+    canonical_router_digest, canonical_router_tensors = _tensor_digest(
```
jasonqinzhou, line 183: The equivalence report still hashes only tensors matching the router-key regex, then separately checks `wo_a`; for every other common tensor it compares only key presence. A repacked checkpoint can therefore change embeddings, attention projections, norms, or expert weights and still receive `status: PASS`, even though those changes alter the hidden states presented to later routers and can change expert popularity. This remains directly relevant because both DeepSeek-V4 bundles are labeled with canonical model IDs while being collected from `sgl-project/*-FP8` checkpoints on the strength of this report. Please byte-verify every common tensor (applying explicit, verified transforms for declared repacked tensors), or run a canonical-vs-checkpoint route-equivalence comparison over the collection inputs, before accepting the repack as canonical. A negative test that mutates an unrelated shared tensor should make the report fail.

python/aisimulate/src/aiconfigurator_core/sdk/expert_popularity.py
```diff
@@ -0,0 +1,251 @@
...
+
+
+def _require_mapping(value: Any, name: str) -> Mapping[str, Any]:
+    if not isinstance(value, Mapping):
+        raise ExpertPopularityDataError(f"metadata.{name} must be a mapping")
+    return value
+
+
+def _validate_public_metadata(metadata: Mapping[str, Any]) -> None:
+    provenance = _require_mapping(metadata.get("provenance"), "provenance")
```
jasonqinzhou, line 101: `_validate_public_metadata` still rejects unexpected provenance keys but does not require the production fields that make a bundle trustworthy. The passing unit fixture has no model revision, measurement block, validation gates/status, immutable image/checkpoint identity, collection digests, or publication timestamp, yet `load_expert_popularity_metadata()` returns it as validated metadata. Please define and enforce the required schema-v1 fields, types, and immutable digest/revision formats (including the canonical-vs-repacked evidence rule), with negative tests for omitted or malformed fields, so a truncated bundle cannot cross the SDK validation boundary.

python/aisimulate/src/aiconfigurator_core/model_data/expert_popularity/deepseek-ai--DeepSeek-V4-Flash/metadata.yaml
```diff
@@ -0,0 +1,300 @@
...
+      SGLANG_OPT_FP8_WO_A_GEMM: '0'
+    schema_version: 1
+    scope: 'Artifact-level routing-equivalence evidence for an official weight-only
+      repack: routing configuration, weight-index identity, accounted physical tensor
+      deltas, every router tensor, exact canonical-FP8-to-checkpoint-BF16 wo_a dequantization,
+      tokenizer artifacts, required runtime semantics, and the checkpoint''s no-retraining
+      provenance claim are verified.'
+    status: PASS
+  framework_version: 0.5.14
+  observation_source: recorder
```
jasonqinzhou, line 288: This bundle uses `observation_source: recorder` together with aggregate repeat validation, but `_run_recorder_window` retains only each window's aggregate `logical_count` tensor plus response/output IDs; unlike `_run_response_window`, it never persists token-level expert IDs. That conflicts with the README's stated aggregate-mode contract that raw per-request routes remain part of the evidence, and removes the ability to audit an anomalous aggregate back to individual requests. Please either retain token-level routes for recorder-mode aggregate collections, or narrow the documented evidence contract and enforce the chosen requirement when publishing aggregate bundles.

jasonqinzhou (changes requested):

## What this PR does

This PR adds a production-oriented SGLang expert-routing collector, publisher, SDK loaders and validators, and packaged popularity bundles for 15 MoE models. The new head adds a self-contained visualizer preview whose database totals and table content summarize the packaged artifacts.

The new preview is polished and useful: it explains the coverage metric, exposes the per-model stability table, and its 15-model, 809-layer, 197,024-row, and 4.02M-token headline totals all reconcile with the checked-in metadata.

## Why this score

Codex scores the current head 6.5 and would request changes because the unchanged repacked-checkpoint verifier still permits unrelated shared weights to differ even though they can change downstream router inputs; the SDK metadata validator remains incomplete; and recorder-mode aggregate evidence still omits the raw per-request routes promised by the documentation. The image-only delta is valid and introduces no new finding. The canonical recommendation is BLOCKED because the required Claude Fable review timed out without structured or model-verifiable evidence, so no finding is eligible for posting from this packet.

python/aisimulate/collector/expert_popularity/verify_routing_equivalence.py
Harrilee: Thanks. We are intentionally keeping two identities here: `model.id` is the canonical, user-facing database lookup key, while `provenance.collection_checkpoint.{id,revision,quantization}` records the exact immutable artifact that served the workload. Quantization/repack variants are provenance rather than lookup dimensions because this database models expert popularity at the model level; variant-specific top-level IDs would fragment one model across checkpoint names and make normal lookup ambiguous. For example, a model distributed natively in FP4 should still be queried under its canonical model name, while the FP4 checkpoint remains explicit in provenance. The report is therefore supporting provenance for an official weight-only/no-retraining repack, not an attempt to rename the collected checkpoint or claim that quantization is a separate model. We retain the immutable checkpoint ID and SHA, exact router tensors/config/tokenizer, declared `wo_a` transformation, and empirical repeat/cross-sequence stability. I clarified this identity contract in 8bf4775f. We are not expanding this PR to make quantization/checkpoint variants separate database keys or to byte-verify every non-routing tensor.

python/aisimulate/src/aiconfigurator_core/sdk/expert_popularity.py
Harrilee: The intended validation boundary is split: `publish.py` applies the production workload/stability gates before a bundle can be produced, while the SDK loader keeps the reusable data-format contract and privacy allowlist lightweight. To make the repository-owned production guarantee explicit in CI without widening the SDK policy for external/custom data roots, 8bf4775f adds `test_packaged_database_has_complete_production_metadata`. It traverses all 15 shipped bundles and requires immutable model/tokenizer/checkpoint revisions, measurement and PASS validation evidence, four shards/two repeats, conservation, immutable image identity, collection digests, timezone-aware publication time, and matching PASS evidence whenever canonical and collection identities differ. A truncated checked-in production bundle now fails CI.

python/aisimulate/src/aiconfigurator_core/model_data/expert_popularity/deepseek-ai--DeepSeek-V4-Flash/metadata.yaml
Harrilee: Agreed that the README overstated the recorder evidence contract. Production collection intentionally uses SGLang native `stat` recorder aggregates; we do not need token-level routes to build or validate this model-level database and do not have capacity to analyze that volume. In 8bf4775f the README now states that each repeat/shard retains the raw aggregate `logical_count` dump, exact workload input IDs, responses/output IDs, endpoint results, and hashes. `response_routed_experts` is explicitly a focused diagnostic path: per-request token-level routes, when available, are diagnostic-only and are not a publication requirement. This matches what recorder-mode collections actually preserve and removes the conflicting promise.

jasonqinzhou (changes requested):
Additional merge blocker: this database needs to improve the simulation path, not only ship as an unused SDK/data asset.

The existing DeepEP-LL design already describes collector-driven routing as Stage 4: accept measured expert probabilities directly and keep the synthetic Power Law model as fallback. Please include that consumer integration in this PR, with the following contract:

1. Make the default selection `auto`: use a validated measured popularity profile when the requested model/checkpoint identity, routing dimensions, and inference phase match a packaged bundle. If no compatible bundle exists, automatically use the model family's current Power Law distribution.
2. Treat absence as a normal Power Law fallback, but treat a present corrupt, invalid, or identity/dimension-mismatched bundle as an error rather than silently hiding it. Preserve an explicit user override for Power Law/balanced behavior.
3. Feed the measured per-layer expert probabilities into the actual prediction path (including the placement/load calculation used by MoE communication and expert compute); do not merely load the metadata or relabel the vector as a Power Law string. Record the selected source and fallback reason in prediction provenance.
4. Respect phase coverage. The current bundles declare `measurement.phase: prefill`, so they may drive context/prefill predictions; generation/decode should continue to fall back until decode-specific measurements exist.
5. Add selection tests for measured-hit, missing-bundle fallback, explicit override, phase mismatch, and invalid-bundle failure. Also provide held-out accuracy evidence comparing the current Power Law baseline against measured popularity for covered models, with an end-to-end or MoE communication/compute error metric showing improvement or at least no regression.

Without a production consumer and before/after accuracy evidence, the PR establishes useful data infrastructure but does not yet deliver the accuracy value that justifies supporting and maintaining this database.

PR #96, opened 2026-09-03
ci: restore migrated AIC validation gates

jasonqinzhou:
## Why and what changed

Migrated AIConfigurator validation assets lived below the package tree, where GitHub could not execute them. This restores the applicable gates at the repository root and integrates them into Fast CI and Full CI: platform wheels, collector data, prediction regression, Python compatibility, Rust feature modes, engine goldens, and recommendation E2Es.

The change also verifies collector/FPM wheel payloads, fixes reduced Docker build context, centralizes Python/Rust setup, pins actions, disables persisted checkout credentials, preserves the optional wheel-builder base-image default, and makes aggregate checks fail closed. Lifecycle comparisons use the previous push commit; PR/manual checks resolve one immutable main SHA. Concurrency preserves lifecycle/manual runs while replacing stale PR validation.

## Review map

- Risk level: high
- Start with: `.github/workflows/ci.yml`, the three restored reusable workflows, `.github/actions/build-platform-wheel/action.yml`, and `tests/test_ci_workflow_contracts.py`
- Public or serialized contract changed: yes — reusable workflow inputs and stable `Fast CI Success` / `Full CI Success` checks; no product API, schema, or model formula changes
- Compatibility or rollback concern: callers must propagate the exact target SHA; revert these CI commits to roll back. Branch rules are unchanged.

## Evidence

- Tests: `d400b55fe94227ae347e9cc2176c5ec57b9cee9a` passes 134 CODEOWNERS/policy/workflow tests, the declared Ruff check/format scope, Python compilation, Cargo formatting, actionlint 1.7.12, whitespace, generated CODEOWNERS, and source/legal checks
- Fast CI: passed on `d400b55fe94227ae347e9cc2176c5ec57b9cee9a`; DCO and CODEOWNERS checks also passed
- Full CI: [exact-head run 34550530761](https://github.com/ai-dynamo/aisimulate/actions/runs/34550530761) passed on `d400b55fe94227ae347e9cc2176c5ec57b9cee9a`: 31 successful jobs and only the expected manual staging skip. [Post-merge run 34550573649](https://github.com/ai-dynamo/aisimulate/actions/runs/34550573649) passed all 30 test/build jobs on merge commit `dba4ae2777ca338a7c703ef3f97dcc8b480e4d7d` after retrying one runner-container failure. Wheel staging and the final aggregate remain pending the protected `automated-release` environment approval from `pvijayakrish`; the current account cannot approve that environment.
- CodeRabbit reviewed commit: `d400b55fe94227ae347e9cc2176c5ec57b9cee9a`; no actionable findings and all eight contract checks passed
- Codex reviewed commit: `d400b55fe94227ae347e9cc2176c5ec57b9cee9a`
- Negative or boundary cases: failed/cancelled/empty required results; incorrect/empty target SHA; copied-PR mismatch/API failure; pre-harness revisions; previous-commit baselines for main/release pushes; PR/manual/new-release baselines; malformed comparison SHA and API failure; omitted wheel-base input
- Expected-value derivation (for exact numerical assertions): recommendation topology assertions use the existing `num_total_gpus` field; no numerical prediction baselines change
- Before/after output, trace, benchmark, or golden diff: N/A — CI restoration and validation fixes only; no model constants, runtime scheduling, datasets, or goldens are changed by this PR

N/A for modeling. Adapted CI assets identify AIConfigurator commit `77fd0773407b3683d8a671fe24a30a7110651b64`; root and wheel-local third-party notices record the same provenance.

## Tracking

- Closes: N/A
- Related PRs or issues: [AIC-1931](https://linear.app/nvidia/issue/AIC-1931/ci-rebucket-pr-96-validation-gates-into-fast-ci-and-full-ci)


.github/workflows/ci.yml
```diff
@@ -107,6 +110,38 @@ jobs:
...
+      - name: Test without Python embedding
+        run: cargo test --workspace --no-default-features
+      - name: Install the unified Python package
+        run: python -m pip install './python/aisimulate[dev]'
+      - name: Test with every Rust feature enabled
+        run: |
+          export AIC_REQUIRE_EMBEDDED_ROUND_TRIP=1
+          export PYO3_PYTHON="${pythonLocation}/bin/python"
+          export PYTHONPATH="${GITHUB_WORKSPACE}/python/aisimulate/src"
+          cargo test --workspace --all-features
```
Arsene12358: **B1 + B2 (blocking):** `--all-features` enables `extension-module` (`crates/core/Cargo.toml:47`), so pyo3 drops the libpython link line → `ld: symbol(s) not found … _Py_InitializeEx`, `could not compile aisimulate-core (lib test)`. Independently, `PYTHONPATH=…/src` shadows the non-editable install → `ImportError: cannot import name '_runtime' from 'aisimulate' (…/src/aisimulate/__init__.py)` → `embedded_round_trip.rs:84` panics because `AIC_REQUIRE_EMBEDDED_ROUND_TRIP=1`. With the change below all 10 test binaries pass locally (1119 passed, 0 failed) with the round trip enforced.

```suggestion
          export AIC_REQUIRE_EMBEDDED_ROUND_TRIP=1
          export PYO3_PYTHON="${pythonLocation}/bin/python"
          cargo test --workspace --features embed-python,replay-bench
```

.github/actions/build-platform-wheel/action.yml
```diff
@@ -0,0 +1,97 @@
...
+      shell: bash
+      env:
+        UV_PYTHON_PREFERENCE: only-managed
+        UV_PYTHON_DOWNLOADS: automatic
+      run: |
+        uv_cache_dir="${RUNNER_TEMP}/aisimulate-uv-cache-${GITHUB_RUN_ID}-${GITHUB_RUN_ATTEMPT}"
+        export UV_CACHE_DIR="${uv_cache_dir}"
+        echo "UV_CACHE_DIR=${uv_cache_dir}" >> "${GITHUB_ENV}"
+
+        uv venv --python 3.13 "${RUNNER_TEMP}/aisimulate-wheel-venv"
```
Arsene12358: **B3 (blocking, macOS leg):** the shared `Verify installed unified wheel` step (line 96) runs `python -m pip install …`, but on the `native-macos` path `python` is this uv venv, and `uv venv` does not install pip (`python -m pip --version` → `No module named pip`). Upstream used `uv pip install` for the macOS verify step. Seeding pip is the smallest fix:

```suggestion
        uv venv --seed --python 3.13 "${RUNNER_TEMP}/aisimulate-wheel-venv"
```

Arsene12358 (changes requested):
## Review: REQUEST_CHANGES — three deterministic failures in the restored gates

I checked out `b0d76fff`, installed the package into a clean Python 3.13 environment exactly as the new jobs do (`pip install './python/aisimulate[dev]'`, non-editable) and re-ran every step of the new jobs locally. The validation numbers in the description reproduce (455 root, 121 contract on 3.13 / 99 on 3.11, 340 parity, 2 recommend E2E, collector and backend-facts checks, actionlint, ruff, cargo fmt, SPDX/legal), the FPE container fix matches the real failure in run 33640723513, and the facade update repairs a verifier that currently fails on `main`. Three steps that are not in the validation list fail deterministically; suggested fixes are inline and repeated below.

### B1 — `rust-feature-modes`: `cargo test --workspace --all-features` fails to link (`.github/workflows/ci.yml:143`)

`crates/core/Cargo.toml:47` declares `extension-module = ["python", "pyo3/extension-module"]` as a named feature, so `--all-features` enables it together with `embed-python`. pyo3 only emits the libpython link line when the `extension-module` feature is off (`pyo3-build-config-0.23.5/src/impl_.rs:877-884`, `pyo3-ffi-0.23.5/build.rs:205`), so every test binary that touches the C-API fails to link:

```
"_Py_InitializeEx", referenced from: pyo3::gil::prepare_freethreaded_python …
ld: symbol(s) not found for architecture arm64
error: could not compile `aisimulate-core` (lib test) due to 1 previous error
```

The AIConfigurator crate this job was adapted from has no `extension-module` feature at all (its `Cargo.toml` comment: "It must NEVER be a default feature: `cargo build`/`cargo test` link libpython (via `auto-initialize`)"), which is why `--all-features` was safe in upstream `build-test.yml:82` and is not safe here. Because `fff16846` adds `rust-feature-modes` to `application-wheel.needs`, this also skips Artifactory wheel staging on every `main`/`release/*` push once merged.

**Fix:** name the features explicitly. Verified locally: `cargo test -p aisimulate-core --features embed-python,replay-bench` → 1119 passed, 0 failed across all 10 test binaries with `AIC_REQUIRE_EMBEDDED_ROUND_TRIP=1` enforced.

```yaml
          cargo test --workspace --features embed-python,replay-bench
```

### B2 — `rust-feature-modes`: `PYTHONPATH=…/python/aisimulate/src` shadows the installed native module (`.github/workflows/ci.yml:142`)

Line 137 installs the package non-editably, so `_runtime.abi3.so` exists only in site-packages. With the source tree first on `PYTHONPATH`, the embedded interpreter resolves `aiconfigurator_core` from `src/`, whose shim (`src/aiconfigurator_core/_aiconfigurator_core.py:14`) runs `from aisimulate import _runtime` against the source-tree `aisimulate` package, which has no `_runtime`. Under `AIC_REQUIRE_EMBEDDED_ROUND_TRIP=1` that is a hard panic by design:

```
embedded_round_trip: import failed … ImportError: cannot import name '_runtime' from 'aisimulate' (…/python/aisimulate/src/aisimulate/__init__.py)
thread 'embedded_round_trip::embedded_builder_round_trip' panicked at crates/core/tests/perfmodel/embedded_round_trip.rs:84:9:
embedded_round_trip: AIC_REQUIRE_EMBEDDED_ROUND_TRIP is set but `aiconfigurator_core.sdk.engine` is not importable
test result: FAILED. 0 passed; 1 failed
```

The same command with `src` removed from `PYTHONPATH` passes (`test result: ok. 1 passed`). Upstream needed a `PYTHONPATH` because it ran in a uv venv (`build-test.yml:81` lists the venv site-packages explicitly); here the wheel is installed into the `setup-python` interpreter that `PYO3_PYTHON` points pyo3 at, so the embedded `sys.path` already contains it.

**Fix:** drop the `PYTHONPATH` export. Combined B1+B2 step:

```yaml
      - name: Test with Python embedding enabled
        run: |
          export AIC_REQUIRE_EMBEDDED_ROUND_TRIP=1
          export PYO3_PYTHON="${pythonLocation}/bin/python"
          cargo test --workspace --features embed-python,replay-bench
```

### B3 — composite action: `python -m pip` inside a `uv venv` has no `pip` (`.github/actions/build-platform-wheel/action.yml:96`)

For `build_method: native-macos`, `python` resolves to `${RUNNER_TEMP}/aisimulate-wheel-venv/bin/python` (lines 68-70), and `uv venv` does not seed pip (`python -m pip --version` → `No module named pip`). Upstream kept two verify steps (`uv pip install` for macOS, `python -m pip` for the docker host); merging them into one unconditional step breaks the `macosx_arm64` job of `validate-platform-wheels.yml` at its last step, after the wheel has already been built.

**Fix:** seed pip when creating the venv (or split the verify step back into the two conditional variants):

```yaml
        uv venv --seed --python 3.13 "${RUNNER_TEMP}/aisimulate-wheel-venv"
```

.github/workflows/ci.yml
jasonqinzhou: Fixed in 7b7731e: the embedding job now removes the source-tree PYTHONPATH and tests only embed-python,replay-bench instead of --all-features. I also added workflow-contract assertions for both constraints. Local validation passed all 10 Rust test binaries with AIC_REQUIRE_EMBEDDED_ROUND_TRIP=1 (1,153 passed, 1 expected data-dependent ignore).

.github/actions/build-platform-wheel/action.yml
jasonqinzhou: Fixed in 7b7731e: the macOS uv environment is now created with --seed, and the workflow-contract test asserts that pip is present for the shared installed-wheel verification step.

tianhaox (commented):
Reviewed against the AIC `.github` snapshot under `python/aisimulate/.github/` and the AIC upstream workflows. This PR closes the biggest CI gap the migration left (collector, prediction gate, feature modes, Python 3.11–3.13, wheel validation), and it correctly moves everything onto the `pull-request/*` push model. **I think we should prioritize landing this as soon as possible**; none of the findings below need to block the merge, and several are cheaper to fix on `main` once the workflows have completed a real run on GitHub.

Inline comments, roughly by severity:

1. Docker-built wheels in `validate-platform-wheels` silently omit `collector/**` (Dockerfile never copies it; verifiers do not check).
2. `rust-python-parity` runs under the root `pytest.ini` (`filterwarnings = error`, no timeout) instead of `python/aisimulate/pytest.ini`.
3. That job is really a frozen-golden regression, not Rust/Python parity — the Python engine step is gone. Suggest renaming (or folding into `application-tests`); the PR description's "Rust/Python parity" wording should follow.
4. `prediction-regression-gate` `report` job has no `setup-python`/`pip install`; it only works because `ubuntu-latest` happens to ship PyYAML.
5. Concurrency groups keyed on `github.ref` cancel concurrent manual dispatches (gate + collector-check).
6. `build-platform-wheels-copied-pr` path filter misses README / THIRD_PARTY_NOTICES / `collector/**` / `.dockerignore`.
7. `ci-success` keeps the gate list in three places and is silent about which job failed.
8. `collector-check` `--depth=1` base fetch grafts a shallow history (reproduced locally).
9. Nested-workflow contract test does not cover `./python/aisimulate/.github/actions/`.
10. `collect(old)` dropped the NO_HARNESS fallback (narrow; only pre-`40bc989a` `old-ref`).
11. `fpe-support-matrix` drops `lfs: true` while `.gitattributes` still LFS-tracks `systems/**/*.txt` (latent, zero files today).

Cleanup-only, not inline: the 4-step Python+Rust toolchain block is repeated 11 times and could be a composite action under the `.github/actions/` directory this PR creates; `cargo test --workspace --no-default-features` duplicates the `rust` job since `crates/core` has `default = []`; the six `cross_package` files are hand-listed twice and equal the whole directory.

Checked and fine: `test_cli_recommend.py` timing under `--timeout=180` (6–8 s locally), CODEOWNERS regeneration (byte-identical), duplicate THIRD_PARTY_NOTICES edit (required by `test_source_compliance`), dropping `--all-features` (correct, `extension-module` would break `cargo test`).

.github/actions/build-platform-wheel/action.yml
```diff
@@ -0,0 +1,97 @@
...
+    - name: Set up Docker Buildx
+      if: inputs.build_method == 'docker'
+      uses: docker/setup-buildx-action@d7f5e7f509e45cec5c76c4d5afdd7de93d0b3df5 # v4.1.0
+
+    - name: Build Linux wheel
+      if: inputs.build_method == 'docker'
+      uses: docker/build-push-action@f9f3042f7e2789586610d6e8b85c8f03e5195baf # v7.2.0
+      with:
+        context: .
+        file: python/aisimulate/docker/Dockerfile
```
tianhaox, line 38: The Docker path builds from `python/aisimulate/docker/Dockerfile`, whose build stage never `COPY`s `python/aisimulate/collector/`, but `[tool.maturin].include` in `pyproject.toml` bundles `collector/**` into the wheel. maturin skips unmatched include globs silently, so the manylinux wheels this action produces lack `collector/cases` and `collector/fpm_forward`, and neither `verify_release_wheels.py` nor `verify_installed_package_layers.py` (default `--expect unified`) looks for them. Result: the gate is green while validating an artifact that differs from the natively built release wheel in `ci.yml` `application-wheel`. Either add the `COPY` to the Dockerfile or have `verify_installed_package_layers.py` assert the collector payload is importable.

.github/workflows/ci.yml
```diff
@@ -353,6 +395,81 @@ jobs:
...
+        run: >-
+          python -m pytest -q
+          python/aisimulate/tests/cross_package/test_core_public_api.py
+          python/aisimulate/tests/cross_package/test_core_version_contract.py
+          python/aisimulate/tests/cross_package/test_import_contract.py
+          python/aisimulate/tests/cross_package/test_single_distribution.py
+          python/aisimulate/tests/cross_package/test_single_oracle_contract.py
+          python/aisimulate/tests/cross_package/test_table_view_registry.py
+
+  rust-python-parity:
```
tianhaox: Naming: the Python engine step was removed in AIC (`d62d3f7`), so these suites no longer compare Rust against Python. They compare the live Rust engine against the frozen goldens in `crates/core/parity_tests/perfmodel/goldens/` (the directory's own README now calls itself "Rust Engine Regression Tests (frozen goldens)"). Suggest renaming the job to something like `engine-golden-regression` / "Engine golden regression", or folding the two files into `application-tests` since it overlaps with `prediction-regression-gate`. The PR description's "Rust/Python parity suites" line should change accordingly. The file paths themselves are an AIC mirror boundary in `scripts/aic_sync.toml`, so I would leave those alone here.

.github/workflows/ci.yml
```diff
@@ -353,6 +395,81 @@ jobs:
...
+      - name: Install native build toolchain
+        run: |
+          sudo apt-get update
+          sudo apt-get install --yes --no-install-recommends build-essential
+      - uses: dtolnay/rust-toolchain@4be7066ada62dd38de10e7b70166bc74ed198c30 # stable
+      - name: Install AISimulate and development dependencies
+        run: python -m pip install './python/aisimulate[dev]'
+      - name: Run engine-step parity suite
+        run: >-
+          python -m pytest -q -rx -n 4
```
tianhaox, line 526: These suites are invoked from the repository root without `-c`, so pytest picks up the root `pytest.ini` (`filterwarnings = error`, `--strict-config`, no timeout) rather than `python/aisimulate/pytest.ini` (`--disable-warnings`, `ignore::DeprecationWarning`, `timeout = 120`) that the AIC original ran under. Any `DeprecationWarning`/`FutureWarning` from pandas/numpy/jax or the SDK itself (`aiconfigurator_core/sdk/models/helpers.py:133`, `aiconfigurator/sdk/task_v2.py:113` emit `DeprecationWarning`) across the 363 cases becomes a test error, so a dependency bump with zero prediction drift turns this job red and blocks `application-wheel`/`ci-success`; a hung case has no per-test timeout. Suggest `-c python/aisimulate/pytest.ini` on both invocations.

.github/workflows/ci.yml
```diff
@@ -421,3 +546,52 @@ jobs:
...
+          APPLICATION_WHEEL_RESULT: ${{ needs.application-wheel.result }}
+          APPLICATION_TESTS_RESULT: ${{ needs.application-tests.result }}
+          PYTHON_COMPATIBILITY_RESULT: ${{ needs.python-compatibility.result }}
+          RUST_PYTHON_PARITY_RESULT: ${{ needs.rust-python-parity.result }}
+          LINT_RESULT: ${{ needs.lint.result }}
+          RELEASE_ARTIFACT_CONTRACT_RESULT: ${{ needs.release-artifact-contract.result }}
+          EXPECTED_APPLICATION_WHEEL_RESULT: ${{ github.event_name == 'workflow_dispatch' && 'skipped' || 'success' }}
+        run: |
+          test "${APPLICATION_WHEEL_RESULT}" = "${EXPECTED_APPLICATION_WHEEL_RESULT}"
+          for result in \
```
tianhaox: The gate-job list is now maintained in three places (`needs:`, the `*_RESULT` env block, and this loop), plus a near-copy in `application-wheel.needs`, and `tests/test_ci_workflow_contracts.py` only checks a 3-name subset. A job added to `needs:` but not to the env/loop is never checked and "CI Success" goes green on a red job. Also, when an upstream job fails, `application-wheel` is skipped and the first `test` fails with only `exit code 1`, never naming the failing job. Simpler and self-maintaining: `if: ${{ contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled') }}` with a step that prints `toJSON(needs)`, keeping only the `application-wheel` skip-vs-success special case explicit.

.github/workflows/prediction-regression-gate.yml
```diff
@@ -0,0 +1,150 @@
...
+      - uses: actions/checkout@v7
+      - name: Download prediction snapshots
+        uses: actions/download-artifact@v8
+        with:
+          pattern: prediction-gate-*
+          path: /tmp/prediction-snapshots
+      - name: Compare predictions and enforce the gate
+        working-directory: python/aisimulate
+        run: |
+          python tools/prediction_regression_gate/report.py \
```
tianhaox, line 167: The `report` job checks out and immediately runs `python tools/prediction_regression_gate/report.py` with no `setup-python` and no `pip install`. The module-level import chain `report.py` → `tools/accuracy_tracking/run_silicon.py` → `make_silicon_refs.py` → `tools/prediction_regression_gate/grid.py` does `import yaml`, so this only works because `ubuntu-latest`'s system Python happens to ship PyYAML. Inherited from the nested original, but since this is the required check, worth adding `actions/setup-python` + `pip install pyyaml` so it does not depend on the runner image.

.github/workflows/prediction-regression-gate.yml
```diff
@@ -0,0 +1,150 @@
...
+        description: "Old revision to compare against (default: main)"
+        required: false
+        default: ""
+
+permissions:
+  contents: read
+  pull-requests: read
+
+concurrency:
+  group: prediction-regression-${{ github.ref }}
```
tianhaox: The AIC original keyed this as `${{ github.event_name == 'workflow_dispatch' && github.run_id || github.ref }}` precisely so two concurrent manual validations never cancel each other. Keyed only on `github.ref` with `cancel-in-progress: true`, a second `workflow_dispatch` on `main` (e.g. a different `old-ref`) cancels the first mid-collect, and the cancelled run's `if: always()` report job then fails closed with "collection result: cancelled". `collector-check.yml` adds the same ref-keyed cancel group (the original had none).

.github/workflows/prediction-regression-gate.yml
```diff
@@ -0,0 +1,150 @@
...
+          sudo apt-get update
+          sudo apt-get install --yes --no-install-recommends build-essential
+      - uses: dtolnay/rust-toolchain@4be7066ada62dd38de10e7b70166bc74ed198c30 # stable
+      - name: Install AISimulate and development dependencies
+        run: python -m pip install './python/aisimulate[dev]'
+      - name: Collect static, scheduling, and silicon snapshots
+        working-directory: python/aisimulate
+        run: |
+          mkdir -p /tmp/gate-snapshot
+          python tools/prediction_regression_gate/collect_static.py \
```
tianhaox, line 111: Minor: the original's `NO_HARNESS.txt` empty-snapshot fallback (`if [ ! -f tools/prediction_regression_gate/collect_static.py ]`) was dropped, so the degrade path documented in `report.py:213-222` is now unreachable in CI. Narrow in practice: every `main`/`release/*` base since `40bc989a` has the harness, so this only fires for a `workflow_dispatch` `old-ref` predating it, and there `pip install './python/aisimulate[dev]'` fails first anyway. Either restore the fallback or drop the dead branch in `report.py` so the two agree.

.github/workflows/build-platform-wheels-copied-pr.yml
```diff
@@ -0,0 +1,34 @@
...
+      # copy-pr-bot is the approval boundary for privileged pull-request builds.
+      - "pull-request/*"
+    paths:
+      - "Cargo.lock"
+      - "Cargo.toml"
+      - "crates/**"
+      - ".github/actions/build-platform-wheel/**"
+      - ".github/workflows/build-platform-wheels-copied-pr.yml"
+      - ".github/workflows/validate-platform-wheels.yml"
+      - "python/aisimulate/docker/Dockerfile"
```
tianhaox: The path filter omits inputs the Docker/maturin build consumes: `python/aisimulate/README.md` (pyproject `readme` and a Dockerfile `COPY`), `python/aisimulate/THIRD_PARTY_NOTICES.md` (`license-files`, byte-compared by `verify_release_wheels.py`), `python/aisimulate/collector/**` (maturin include), and `.dockerignore`. A copied-PR push that touches only those merges green and the next `main` wheel build fails at the `COPY` or the legal-file comparison. Inherited from the nested original.

.github/workflows/collector-check.yml
```diff
@@ -0,0 +1,161 @@
...
+            const {data: pull} = await github.rest.pulls.get({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              pull_number: Number(match[1]),
+            });
+            core.setOutput("sha", pull.base.sha);
+      - name: Fetch comparison base
+        env:
+          BASE_SHA: ${{ steps.base.outputs.sha }}
+        run: git fetch --no-tags --depth=1 origin "${BASE_SHA}"
```
tianhaox: `git fetch --no-tags --depth=1 origin "${BASE_SHA}"` after a `fetch-depth: 0` checkout writes `.git/shallow` and grafts `BASE_SHA` as parentless (reproduced locally: `git log BASE_SHA` shows a single commit afterwards). The original fetched `${BASE_REF}:refs/remotes/origin/${BASE_REF}` without `--depth`. `changed_ops.py` only uses `rev-parse`/`cat-file`/`show`/`ls-tree` today so nothing breaks yet, but any later `merge-base`/`log`/`A...B` use in this job would silently see truncated history. Either drop `--depth=1`, or make the checkout shallow too and rely solely on the explicit base fetch (`fetch-depth: 0` is otherwise ~280 MiB of historical parquet for nothing).

tests/test_ci_workflow_contracts.py
```diff
@@ -0,0 +1,126 @@
...
+
+    for workflow_name, job_name in jobs:
+        job = _workflow(workflow_name)["jobs"][job_name]
+        checkout = next(step for step in job["steps"] if step.get("uses", "").startswith("actions/checkout@"))
+        assert checkout.get("with", {}).get("lfs") != "true"
+
+
+def test_active_workflows_do_not_call_nested_inert_workflows() -> None:
+    for path in WORKFLOW_ROOT.glob("*.yml"):
+        assert "./python/aisimulate/.github/workflows/" not in path.read_text(encoding="utf-8")
```
tianhaox: This guard only forbids the substring `./python/aisimulate/.github/workflows/`, so `uses: ./python/aisimulate/.github/actions/build-platform-wheel` (the exact form the nested `validate-platform-wheels.yml` uses) would pass. That nested action is a near-identical twin of the new root action minus the `uv venv --seed` fix, so a workflow that picks it up would reproduce the macOS `pip` failure this PR fixes. Suggest widening to `./python/aisimulate/.github/` (or parsing each step's `uses:`).

.github/workflows/fpe-support-matrix.yml
```diff
@@ -60,13 +65,16 @@ jobs:
       - uses: actions/checkout@v7
```
tianhaox, line 65: Latent: `lfs: true` is removed here and the new contract test forbids re-adding it, but `.gitattributes` still LFS-tracks `python/aisimulate/src/aiconfigurator_core/systems/**/*.txt`, and `pyproject.toml` ships `systems/**/*.txt` in the wheel with no pointer-stub check on the `pip install` path (only the Dockerfile has one). Zero files match today (checked with `git check-attr`), so nothing breaks now; the first perf-data PR that adds a `.txt` under `systems/` would make this job read a 130-byte LFS pointer as data with no error, and the contract test would block the obvious fix. Consider dropping the stale `.gitattributes` rule in the same PR, or pointing the contract test's error message at it.

.github/actions/build-platform-wheel/action.yml
jasonqinzhou: Fixed in 3b64fa69. The Docker wheel build now copies the declared collector tree, the source-tree verifier requires every collector path declared by pyproject.toml, and installed-wheel verification checks representative collector files plus imports collector and collector.fpm_forward.

.github/workflows/ci.yml
jasonqinzhou: Fixed in 3b64fa69. Both golden-regression pytest invocations now select python/aisimulate/pytest.ini explicitly.

.github/workflows/ci.yml
jasonqinzhou: Fixed in 3b64fa69. The job and PR copy now call this Engine golden regression, matching the suites it actually runs.

.github/workflows/ci.yml
jasonqinzhou: Fixed in 3b64fa69. CI Success now consumes toJSON(needs), derives the expected result for every declared dependency, prints every observed result, and fails with the exact mismatches. The hand-maintained shell list is gone.

.github/workflows/prediction-regression-gate.yml
jasonqinzhou: Fixed in 3b64fa69. Harness detection now happens immediately after each old/new checkout, setup and collection are conditional on its presence, and a pre-harness revision emits NO_HARNESS.txt for the report path.

.github/workflows/prediction-regression-gate.yml
jasonqinzhou: Fixed in 3b64fa69. Manual collector and prediction runs key concurrency by github.run_id, while copied-PR pushes continue to key by ref.

.github/workflows/prediction-regression-gate.yml
jasonqinzhou: Fixed in 3b64fa69. The report job now sets up pinned Python 3.13 and installs pinned PyYAML 6.0.3 before invoking report.py.

.github/workflows/collector-check.yml
jasonqinzhou: Fixed in 3b64fa69. The collector base fetch no longer uses depth=1; checkout is full-history and the requested base SHA is fetched without creating a shallow graft.

.github/workflows/fpe-support-matrix.yml
jasonqinzhou: Addressed in 3b64fa69 without reintroducing the known runner failure. The job keeps LFS disabled because the required container does not provide git-lfs and there are currently no legacy systems/**/*.txt payloads. A fail-fast contract now enforces both assumptions and directs a future contributor to update the CI image or remove the stale .gitattributes rule before adding such a file.

tests/test_ci_workflow_contracts.py
jasonqinzhou: Fixed in 3b64fa69. The contract now rejects any active-workflow reference under ./python/aisimulate/.github/, covering nested actions as well as workflows.

.github/workflows/build-platform-wheels-copied-pr.yml
jasonqinzhou: Fixed in 3b64fa69. The copied-PR wheel trigger now includes .dockerignore, both notice inputs, the package README, and collector/** in addition to the prior build inputs.

jasonqinzhou:
Cleanup follow-up at `3b64fa69`: same-revision Python/Rust setup is now centralized in `.github/actions/setup-python-rust`; the redundant no-default-features workspace run is removed; and both callers now run the `cross_package` directory instead of duplicating six paths. The prediction snapshot job intentionally keeps inline setup because it checks out old and new revisions, and an older checkout may not contain this local action.

jasonqinzhou:
Ready for re-review at `5a650c16`. All 11 inline findings are addressed and replied to. [Exact-head Full CI](https://github.com/ai-dynamo/aisimulate/actions/runs/34278521290) passed with 31 successful jobs, zero failures, and only the intentional manual staging job skipped. Re-review has been requested from @Arsene12358 and @tianhaox; reviewer-owned threads remain open for their verification.

.github/actions/build-platform-wheel/action.yml
jasonqinzhou: Follow-up at `5a650c1`: the exact-head Full CI run exposed one more reduced-Docker-context gap after the collector copy—the `aiconfigurator/{model_configs,systems}` compatibility symlinks require `aic-core/src`. The image now recreates that link and fails fast if either target is unresolved. All three Platform Wheels jobs (Linux amd64/arm64 and macOS arm64), the source-tree verifier, and installed-wheel checks passed in https://github.com/ai-dynamo/aisimulate/actions/runs/34278521290.

.github/workflows/prediction-regression-gate.yml
```diff
@@ -0,0 +1,181 @@
...
+      old: ${{ steps.resolve.outputs.old }}
+    steps:
+      - name: Resolve comparison base
+        id: resolve
+        uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd # v8
+        env:
+          OLD_REF_INPUT: ${{ inputs.old-ref }}
+        with:
+          script: |
+            if (context.eventName === "workflow_dispatch" || context.eventName === "workflow_call") {
```
Arsene12358: [P1] Accept the caller's push event in this reusable gate

`ci.yml` calls this workflow on pushes to `main`, `release/*`, and `pull-request/*`. A reusable workflow retains the caller's GitHub context, so `context.eventName` is `push`, not `workflow_call` ([GitHub documentation](https://docs.github.com/en/actions/reference/workflows-and-actions/reusing-workflow-configurations#github-context)). This condition rejects every automatic Full CI invocation: `refs` fails, `collect` is skipped, the report fails closed, and `Full CI Success` fails. This is introduced by the new root gate and dependency; the reviewed base has neither.

Executing the exact resolver script from this YAML with `OLD_REF_INPUT=main` produced:

```text
workflow_dispatch: outputs={"old":"main"}, errors=[]
push:              outputs={}, errors=["unsupported event: push"]
```

The actual aggregate script then exits 1 with `Full CI did not pass: prediction-regression=failure, expected success`. The green run linked in the PR used `workflow_dispatch` and does not cover this path.

| Current behavior | Required behavior |
| --- | --- |
| Reject the supported caller event `push` | Resolve the supplied comparison ref for both automatic and manual callers |

The smallest fix is to remove the event restriction and retain `core.setOutput("old", process.env.OLD_REF_INPUT || "main");`, or explicitly support `push`. Add a behavioral regression test that executes this script for `push` and `workflow_dispatch`; the current string-based workflow contracts pass despite this failure.

Arsene12358 (changes requested):
Reviewed head `5a650c16bc0da09895b4e8fd060eb5d7d9e06ec1` against base `ffcb6576b3a60077ea1200788f1b784213f979cf`. One P1 merge blocker remains: the new reusable prediction gate rejects the caller's `push` event, so the automatic Full CI paths on `main`, `release/*`, and `pull-request/*` fail before collecting predictions.

The [successful exact-head Full CI run](https://github.com/ai-dynamo/aisimulate/actions/runs/34278521290) used `workflow_dispatch`, which explains why it did not expose this defect. I reproduced it by executing the resolver JavaScript extracted from the reviewed YAML with `context.eventName = "push"`, then executing the actual Full CI aggregate script with the resulting failed dependency. Independent standards and spec reviews confirmed the same root cause.

The 21 workflow/CODEOWNERS policy tests, actionlint 1.7.12, pinned Ruff 0.14.1 checks, and source/legal checks passed. No other hard blockers were identified. Please fix the event handling and cover both push and manual callers before merging.

.github/workflows/prediction-regression-gate.yml
jasonqinzhou: Fixed in `0db3a5cc`. The comparison-base resolver no longer branches on `context.eventName`; it now emits `${OLD_REF_INPUT:-main}` from a Bash step, so a reusable invocation retains the caller's `push` context without being rejected. The workflow-contract suite executes the exact production step for `push` (explicit and fallback base) and `workflow_dispatch` callers. The focused local matrix is green (155 tests, including the new behavioral cases), and actionlint 1.7.12 passes all root workflows. I am leaving this reviewer-owned thread open for verification.

jasonqinzhou:
@coderabbitai review

jasonqinzhou:
Ready for re-review at `0db3a5cc`.

- merged current `main` (`10613496`) and resolved the CI aggregate conflicts without regressing main's fail-closed Fast CI behavior
- fixed the P1 reusable-workflow caller bug by making comparison-base resolution event-agnostic
- added executable coverage for `push`, its default-base fallback, and `workflow_dispatch`
- exact-head Fast CI, DCO, and CODEOWNERS checks are green
- local validation: 155 focused tests plus 2 recommendation E2Es, Ruff 0.14.1, actionlint 1.7.12, compileall, cargo fmt, SPDX/legal, and generated CODEOWNERS

Re-review is requested from @Arsene12358, @tianhaox, and the infra CODEOWNERS team. Reviewer-owned threads remain open for verification. The advisory CodeRabbit pass is still in progress. Per `REVIEW.md`, exact-head Full CI will follow after the repaired P1 is verified.

.github/workflows/collector-check.yml
```diff
@@ -0,0 +1,156 @@
...
+    runs-on: prod-aisimulate-default-amd64-v1
+    timeout-minutes: 20
+    container:
+      image: ${{ vars.CI_JOB_CONTAINER_IMAGE }}
+      env:
+        HOME: /__w/aisimulate/aisimulate
+        USER: runner
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v7
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in 9e545196. Both checkout steps and setup-python are pinned to immutable 40-character commit SHAs.

.github/workflows/collector-check.yml
```diff
@@ -0,0 +1,156 @@
...
+        with:
+          ref: ${{ inputs.expected_sha || github.sha }}
+          fetch-depth: 0
+      - name: Verify exact target
+        run: test "$(git rev-parse HEAD)" = "${{ inputs.expected_sha || github.sha }}"
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in 9e545196. Both exact-target checks bind the expression through a TARGET_SHA step environment variable and quote it in the shell command.

.github/workflows/collector-check.yml
```diff
@@ -0,0 +1,156 @@
...
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  pull-requests: read
+
+concurrency:
+  # Manual and reusable invocations must not cancel another comparison run.
+  group: collector-data-${{ github.run_id }}
+  cancel-in-progress: true
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in 9e545196. I removed cancel-in-progress from the run-id-isolated collector concurrency group, so manual and reusable comparison runs cannot cancel one another.

.github/actions/build-platform-wheel/action.yml
```diff
@@ -0,0 +1,97 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+# Adapted from ai-dynamo/AIConfigurator at commit
+# 77fd0773407b3683d8a671fe24a30a7110651b64; modified for AISimulate packaging.
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in 9e545196. The platform-wheel action and validation workflow now include the canonical AIConfigurator source URL at the pinned upstream commit; I also added the same traceability to the other adapted workflow/action headers.

.github/workflows/collector-check.yml
```diff
@@ -0,0 +1,156 @@
...
+      image: ${{ vars.CI_JOB_CONTAINER_IMAGE }}
+      env:
+        HOME: /__w/aisimulate/aisimulate
+        USER: runner
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v7
+        with:
+          ref: ${{ inputs.expected_sha || github.sha }}
+          fetch-depth: 0
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in 9e545196. Both checkouts now set persist-credentials: false. Because this repository is private, the comparison-base fetch uses github.token only through a one-command HTTP authorization header, without writing credentials to git config.

.github/workflows/prediction-regression-gate.yml
```diff
@@ -0,0 +1,176 @@
...
+      - name: Detect prediction-regression harness
+        id: harness
+        shell: bash
+        run: |
+          if [[ -f python/aisimulate/tools/prediction_regression_gate/collect_static.py ]]; then
+            echo "present=true" >> "${GITHUB_OUTPUT}"
+          else
+            echo "present=false" >> "${GITHUB_OUTPUT}"
+          fi
+      - uses: actions/setup-python@v7
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in 9e545196. setup-python is pinned to the immutable v7 commit SHA; the adjacent checkout and download-artifact invocations are pinned as well.

.github/workflows/prediction-regression-gate.yml
```diff
@@ -0,0 +1,176 @@
...
+        default: ""
+
+permissions:
+  contents: read
+  pull-requests: read
+
+concurrency:
+  # Manual and reusable invocations must not cancel another comparison run.
+  group: prediction-regression-${{ github.run_id }}
+  cancel-in-progress: true
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in 9e545196. Prediction-regression concurrency now isolates workflow_dispatch runs by run_id while deduplicating automatic trusted-ref runs by github.ref.

.github/workflows/prediction-regression-gate.yml
```diff
@@ -0,0 +1,176 @@
...
+      - name: Fail closed if snapshot collection failed
+        if: needs.collect.result != 'success'
+        run: |
+          echo "::error::prediction snapshot collection result: ${{ needs.collect.result }}"
+          exit 1
+      - uses: actions/checkout@v7
+        with:
+          ref: ${{ inputs.expected_sha || github.sha }}
+      - name: Verify exact target
+        run: test "$(git rev-parse HEAD)" = "${{ inputs.expected_sha || github.sha }}"
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in 9e545196. The report job now passes the expected revision through a TARGET_SHA environment variable and resolves it as a commit before comparison.

.github/codeowners/test_aisimulate_policy.py
```diff
@@ -367,39 +385,40 @@ def test_full_ci_readiness_fails_closed(tmp_path: Path) -> None:
...
+    canceled_job = {**passing_pr, "NEEDS_JSON": json.dumps(canceled_results)}
     assert _run_readiness_script(script, tmp_path, canceled_job).returncode != 0
 
-    missing_job = {**passing_pr, "APPLICATION_TESTS_RESULT": ""}
+    missing_results = {
+        **passing_results,
+        "application-tests": {"result": ""},
+    }
+    missing_job = {**passing_pr, "NEEDS_JSON": json.dumps(missing_results)}
     assert _run_readiness_script(script, tmp_path, missing_job).returncode != 0
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in 9e545196. The fail-closed aggregate test now explicitly injects a normal failure result and verifies that Full CI exits nonzero.

python/aisimulate/tests/e2e/cli/test_cli_recommend.py
```diff
@@ -0,0 +1,62 @@
...
+    parallel_dims = [
+        top.get("tp", 1),
+        top.get("moe_tp", 1),
+        top.get("moe_ep", 1),
+        top.get("(p)tp", 1),
+        top.get("(d)tp", 1),
+        top.get("(p)moe_ep", 1),
+        top.get("(d)moe_ep", 1),
+    ]
+    assert max(parallel_dims) > 8
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in 9e545196. The incomplete parallel-dimension assertion was removed; the E2E still proves the recommendation is multi-node via num_total_gpus > 8 and retains positive latency checks.

.github/workflows/fast-ci.yml
```diff
@@ -97,9 +103,22 @@ jobs:
...
+          python/aisimulate/tests/e2e/cli/test_cli_recommend.py
           tests
+      - name: Ruff format
+        run: >-
+          ruff format --check
+          --config python/aisimulate/pyproject.toml
+          python/aisimulate/src/aisimulate
+          python/aisimulate/tests/e2e/cli/test_cli_experiments.py
+          python/aisimulate/tests/e2e/cli/test_cli_recommend.py
+          tests/test_ci_workflow_contracts.py
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in 9e545196 and 200ddf56. Ruff format now covers the full tests tree, matching Ruff lint, and the previously drifting tests were mechanically formatted so the widened check passes.

.github/workflows/prediction-regression-gate.yml
```diff
@@ -0,0 +1,176 @@
...
+  workflow_dispatch:
+    inputs:
+      old-ref:
+        description: "Old revision to compare against (default: main)"
+        required: false
+        default: ""
+
+permissions:
+  contents: read
+  pull-requests: read
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in 9e545196. The unused pull-requests: read permission was removed; the workflow now requests contents: read only.

.github/workflows/validate-platform-wheels.yml
```diff
@@ -0,0 +1,85 @@
...
+  contents: read
+
+jobs:
+  linux-amd64:
+    name: Build wheels (manylinux_2_28_x86_64)
+    runs-on: ubuntu-24.04
+    timeout-minutes: 90
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v7
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in 9e545196. All three checkout invocations are pinned to the immutable v7 commit SHA, and each exact-target shell check now uses a quoted TARGET_SHA environment variable.

.github/workflows/ci.yml
```diff
@@ -375,36 +404,93 @@ jobs:
         HOME: /__w/aisimulate/aisimulate
         USER: runner
     steps:
-      - uses: actions/checkout@v7
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in 8d1c6c2d. All nine direct Full CI checkouts now set persist-credentials: false. None of those jobs performs a later authenticated Git operation, and a workflow-contract test enforces the setting across every direct Full CI checkout.

jasonqinzhou:
Ready for re-review at `8d1c6c2d`.

- current `main` (`82de117b`) is included, and the remote branch exactly matches this head
- all 14 CodeRabbit findings are fixed and individually answered; the final incremental review covered `8d1c6c2d` with no additional actionable comments
- [exact-head Fast CI](https://github.com/ai-dynamo/aisimulate/actions/runs/34315643516), DCO, CODEOWNERS, and CodeRabbit are green
- local validation: 105 CODEOWNERS/policy tests, 19 workflow-contract tests, 569 root tests (5 intentional integration skips), 2 recommendation E2Es, Ruff 0.14.1, actionlint 1.7.12, compileall, cargo fmt, SPDX/legal, and workflow security checks
- the PR description now follows the repository template and the CodeRabbit description check passes

The 12 remaining conversations are human-reviewer-owned and stay open for verification. @Arsene12358, @tianhaox, and the applicable CODEOWNERS teams remain requested. Per `REVIEW.md`, exact-head Full CI follows after the repaired P1 is verified; this is ready for human re-review, not yet claimed merge-ready.

.github/workflows/ci.yml
```diff
@@ -67,6 +72,28 @@ jobs:
...
+    with:
+      expected_sha: ${{ github.sha }}
+
+  prediction-regression:
+    name: Prediction Regression
+    needs: fast-ci
+    uses: ./.github/workflows/prediction-regression-gate.yml
+    with:
+      expected_sha: ${{ github.sha }}
+      old-ref: main
```
tianhaox: Non-blocking. For `push` to `main`, `old-ref: main` resolves to the branch head, which is `github.sha` itself (or a newer commit if another push landed), so the old and new snapshots are identical and the gate cannot fail. `release/*` pushes compare against `main` rather than the previous release commit, and the collector's default base (default-branch head) has the same shape, so `changed_ops` is empty on `main` pushes. For copied-PR pushes `main` head is a reasonable base. Follow-up options: pass `github.event.before` for `main`/`release/*` pushes (skip when all zeros), or gate these two reusable jobs on `startsWith(github.ref, 'refs/heads/pull-request/')` and save the two 45-minute collects on every merge.

.github/workflows/prediction-regression-gate.yml
```diff
@@ -0,0 +1,184 @@
...
+
+permissions:
+  contents: read
+
+concurrency:
+  # Isolate manual runs, but replace stale automatic runs for the same trusted ref.
+  group: >-
+    prediction-regression-${{ github.event_name == 'workflow_dispatch'
+      && github.run_id || github.ref }}
+  cancel-in-progress: true
```
tianhaox, line 37: Non-blocking, unverified against GitHub docs. This group is keyed on `github.ref` for non-dispatch events with `cancel-in-progress: true`, and `validate-platform-wheels.yml:27` has the same shape. If GitHub honours a called workflow's top-level `concurrency` in the caller's context (which is my understanding), then two merges to `main` in quick succession make the second run cancel the first run's `Prediction Regression` / `Platform Wheels` jobs, so the first commit's `Full CI Success` goes red and its post-merge wheel is never staged (`fast-ci.yml` on `main` already has this shape). A safe variant is to cancel only on trusted PR copies: `${{ startsWith(github.ref, 'refs/heads/pull-request/') && github.ref || github.run_id }}`.

tianhaox (commented):
Re-reviewed at `8d1c6c2d`. All 11 items from my previous review are addressed in the head files (collector payload in the Docker wheel + verifiers, `-c python/aisimulate/pytest.ini`, the `engine-golden-regression` rename, report-job deps, run_id-keyed manual concurrency, NO_HARNESS fallback, full-history collector fetch, nested-asset guard, `toJSON(needs)` aggregate, FPE LFS contract). The reusable-gate push-event fix for the earlier P1 also looks right, and the `setup-python-rust` composite action is a nice consolidation.

**From my side this is good to merge as-is.** Nothing below is blocking; items 2 and 3 are fine as follow-ups on `main`.

1. **Pre-merge evidence (cheap, recommended):** the last Full CI run on GitHub is the `workflow_dispatch` at `5a650c16`; the five commits since (including the push-event resolver fix) have no run. Because this is a same-repo branch, copy-pr-bot never created `pull-request/96`, so the automatic `push` path through the reusable gates, which is exactly where the P1 lived, has not executed on GitHub yet; only the contract test exercised the script locally. Suggest one `workflow_dispatch` with `expected_sha=8d1c6c2d`, and if a maintainer can push the head to `refs/heads/pull-request/96` once, that covers the push path too.
2. On `push` events the prediction gate and collector compare against the `main` branch head, so for pushes to `main` the old side is the same (or a newer) commit and the gate is vacuous while still spending two 45-minute self-hosted collects. See inline.
3. `validate-platform-wheels.yml` and `prediction-regression-gate.yml` carry ref-keyed `cancel-in-progress: true` groups inside called workflows (same shape as `fast-ci.yml` on `main`). If GitHub applies called-workflow concurrency in the caller's context, two back-to-back merges to `main` cancel the earlier run's gates and its post-merge wheel is never staged. Not verified against docs; see inline.

tianhaox (approved):

tianhaox:
approved and added some comment, let's merge this asap.

Arsene12358 (approved):
Re-reviewed current head `8d1c6c2d9cdb6d15c2b8f3e8b735d42419a83a53`. The P1 from my previous review is resolved: the comparison-base resolver no longer rejects the caller's `push` event, and the behavioral regression test executes the production step for push with an explicit base, push with the default base, and manual dispatch.

Validation: 124 workflow/CODEOWNERS policy tests passed locally. Independent standards and spec follow-up reviews found no remaining merge-blocking code issue; the 25 existing root test files changed for formatting have identical Python ASTs to the current base. Exact-head Fast CI is green.

Approving the code changes. Full CI on this exact head is still required before merge, following the repository's review-before-dispatch process. The existing non-blocking comparison-baseline and automatic-concurrency observations remain follow-ups.

dreamtalen (approved):

.github/workflows/ci.yml
jasonqinzhou: Confirmed in the current head. I am addressing this during the authorized merge closeout: resolve one immutable comparison SHA, use the push's previous commit for main/release lifecycle runs, and pass it to both collector and prediction checks. PR/manual runs and new release branches will resolve main once. Behavioral tests will cover these event paths before the final Full CI run.

.github/workflows/prediction-regression-gate.yml
jasonqinzhou: Confirmed the concurrency semantics against GitHub's documentation: https://docs.github.com/en/actions/reference/workflows-and-actions/reusing-workflow-configurations . I am fixing this during merge closeout by isolating lifecycle/manual runs by run ID and retaining cancellation only for newer PR validation. This must cover reusable Fast CI too; otherwise it can still cancel a prior lifecycle run before its downstream gates start.

.github/workflows/ci.yml
jasonqinzhou: Fixed in ad88f8c304e845bbb959e43538f684a940f42927. Verify Exact Target now resolves one full comparison SHA and passes it to both reusable gates. Existing main/release pushes use github.event.before; manual/trusted-PR/new-release runs resolve main once. Six executable event cases and three failure cases pass, including malformed SHA and GitHub API failure. Full CI will validate the final head after the fresh review finishes.

.github/workflows/prediction-regression-gate.yml
jasonqinzhou: Fixed in ad88f8c304e845bbb959e43538f684a940f42927 across Fast CI, Prediction Regression, and Platform Wheels. Lifecycle and manual invocations use the run ID; only newer direct-PR or trusted-copy PR validation can replace an earlier run. Updated workflow contracts, actionlint 1.7.12, and all 133 policy/workflow tests pass.

jasonqinzhou (commented):
Codex closeout review at ad88f8c304e845bbb959e43538f684a940f42927 against main b9aa2a05e57e7a5cefac20a5b498a6cfb17fee8e: no remaining actionable code finding. Rechecked the earlier runtime/setup, wheel-payload, workflow-composition, exact-SHA, security, and aggregation corrections. The two final lifecycle observations are now fixed in this head, with executable comparison-base boundary tests. All 133 policy/workflow tests, actionlint 1.7.12, declared Ruff check/format scope, Cargo formatting, source/legal checks, generated CODEOWNERS, and whitespace checks pass. The 11 earlier discussions were resolved following Tianhao's explicit re-review acceptance; the two additional discussions now have fix evidence and are resolved. Fresh CodeRabbit review and Full CI on this exact head remain prerequisites for the authorized merge.

.github/actions/build-platform-wheel/action.yml
```diff
@@ -0,0 +1,98 @@
...
+    description: Python architecture passed to setup-python
+    required: true
+  platform:
+    description: Docker target platform for Linux builds
+    required: false
+    default: ""
+  wheel_base:
+    description: Manylinux wheel-builder base image
+    required: false
+    default: ""
```
jasonqinzhou: [P2] Preserve the Dockerfile base image when the optional input is omitted. This default is empty, but the Docker step always supplies WHEEL_BUILD_BASE=${{ inputs.wheel_base }}, overriding Dockerfile's valid manylinux default and producing an invalid FROM for a caller that omits wheel_base. Current platform workflows supply the input explicitly. Align the action default with the Dockerfile and cover the omitted-input contract.

jasonqinzhou (commented):
Validated the optional wheel_base contract warning in CodeRabbit's final summary; recording the concrete correction before applying it.

.github/actions/build-platform-wheel/action.yml
jasonqinzhou: Fixed in d400b55fe94227ae347e9cc2176c5ec57b9cee9a. The optional action input now defaults to the Dockerfile's manylinux x86_64 image. The new contract checks both default alignment and the Docker build argument produced when the caller omits wheel_base. All 134 policy/workflow tests, Ruff, actionlint, and whitespace checks pass.

jasonqinzhou (commented):
Codex follow-up review at d400b55fe94227ae347e9cc2176c5ec57b9cee9a: the optional wheel-base default now matches the Dockerfile, and the omitted-input path is covered. No remaining actionable finding. All 134 policy/workflow tests, Ruff, actionlint 1.7.12, and whitespace checks pass on this head. Additional integration evidence at its parent ad88f8c3: a fresh native package rebuild, 597 root tests passed with five expected cross-repository skips, and both recommendation E2Es passed. Only the action default and its regression test changed afterward. Fresh CodeRabbit review is pending; Full CI will follow on this final head.

jasonqinzhou:
Merge closeout verified: merge commit `dba4ae2777ca338a7c703ef3f97dcc8b480e4d7d` is on main and has exactly the same tree as final reviewed head `d400b55fe94227ae347e9cc2176c5ec57b9cee9a`. All 31 review conversations are resolved. The lifecycle-baseline, concurrency, and optional wheel-base follow-ups are fixed, and CodeRabbit's final review reported no actionable findings with all eight contract checks passing.

[Full CI on the final PR head](https://github.com/ai-dynamo/aisimulate/actions/runs/34550530761) passed: 31 successful jobs, with only the expected manual-run staging skip. Local validation also passed 134 policy/workflow tests; the parent integration build passed 597 root tests (five expected cross-repository skips) and both recommendation E2Es.

[Post-merge CI](https://github.com/ai-dynamo/aisimulate/actions/runs/34550573649) now has 30 successful test/build jobs. Its ARM release-artifact job initially failed in the runner's custom container implementation before building or testing; the retry passed. The remaining wheel-staging jobs and final aggregate are waiting for the protected `automated-release` environment approval from `pvijayakrish`. GitHub reports that the current account cannot approve this environment. No release approval was bypassed.

PR #98, opened 2026-09-03
feat(replay): model vLLM prefill schedule interval

simone-chen:
## Summary

- Add `EngineConfig::prefill_schedule_interval` with a backward-compatible default of 1.
- Drive cadence from a wave-local step shared by every attention-DP rank.
- On non-aligned steps, defer local prefill work with more than one token remaining while allowing connector loads and materialized or near-total prefix hits.
- Preserve the capacity-bound release only on non-preempting passes.
- Reset cadence across every immediate drain transition and document the difference from the upstream 32-step collective tail.
- Expose `engine.workers.<role>.scheduler.prefill_schedule_interval` through the Python prediction config and lower it into the native rank used by `aisimulate predict`.

## Accuracy impact

Kimi-K2.5 NVFP4, ISL 8192, OSL 1024, interval 4:

| GPU / topology | Concurrency | TPOT gap before | TPOT gap after |
| --- | ---: | ---: | ---: |
| B200, DP 8 / EP 8 | 512 | +94.1% | +7.4% |
| B200, DP 8 / EP 8 | 1024 | +73.4% | +2.9% |
| B300, DP 4 / EP 4 | 128 | +38.7% | +16.8% |
| B300, DP 4 / EP 4 | 256 | +45.6% | +10.1% |
| B300, DP 4 / EP 4 | 512 | +40.9% | +5.5% |

The interval-1 control exactly reproduces the prior AISimulate results.

## Validation

- `cargo test -p aisimulate-core`: 1,088 unit tests passed, 1 ignored; all 50 integration tests passed.
- Rebuilt the Python wheel from the PR head and ran `pytest tests -q -p no:timeout`: 451 passed, 5 expected skips.
- `ruff check` and `ruff format --check` pass for the changed Python files.
- Focused regressions cover aligned-step preemption, one-token prefix remainders, cancel-to-idle wave reset, positive interval validation, and YAML-to-native-rank lowering.
- CODEOWNERS strict changed-file coverage: 100%.
- The original feature commit rebuilt the release wheel and reproduced the B200/B300 values above.


Arsene12358 (approved):
Approving. Verified locally at ec7c1680: `cargo fmt` clean, core crate 1085 unit + 50 integration tests green. Cross-checked the model against upstream vLLM at the pinned commit e2fa2859 (`DPEngineCoreProc._should_throttle_prefills` + `Scheduler.schedule(throttle_prefills)`): the `defer_prefills` conjunction, the running-loop chunk skip, and the capacity-bound release all line up, and interval 1 is statically inert so the default path is unchanged.

Two non-blocking fidelity notes you may want to pick up here or in a follow-up:

1. `prefill_capacity_bound` is updated even on passes that preempted (core.rs, the `if !defer_prefills` block after the waiting loop). Upstream only reaches that assignment inside `if not preempted_reqs` (scheduler.py L684 → L1105). Since `preempt_uuid` prepends to waiting, a preempting pass flips the flag to true and releases the cadence one step early under KV pressure. Adding `&& !preempted_any` to the guard restores parity.
2. The waiting loop is gated wholesale by `!defer_prefills`. Upstream's deferral is a `break` after the `load_kv_async` branch and only when `num_computed_tokens < num_tokens - 1` (scheduler.py L866–873), so connector async loads and ≤1-token-remaining requests still get admitted on throttled steps. Only matters with native host offload or near-total prefix hits under DP, so fine to document as a simplification instead.

Also worth noting for users: the knob is reachable through the Runner API / nested `rank` JSON as in the doc, but not yet through the `SchedulerPredictionConfig` spec used by the CLI and Sweeper (StrictModel forbids extra keys). Follow-up material.

crates/core/src/engine/scheduler/vllm/core.rs
```diff
@@ -1882,6 +1919,10 @@ impl VllmCore {
             }
         }
 
+        if !defer_prefills {
+            self.prefill_capacity_bound = !self.state.waiting_members.is_empty();
```
jasonqinzhou, line 1930: Please preserve vLLM's preemption guard when updating this flag. Upstream assigns `prefill_capacity_bound = bool(waiting)` only inside its `if not preempted_reqs` waiting-scheduling branch. Here, a KV allocation failure can set `preempted_any`, requeue the victim into `waiting_members`, skip the waiting loop, and still set this flag to `true`. The next non-aligned pass then bypasses cadence throttling even though vLLM would keep prefills deferred. This can distort the memory-pressure workloads where preemption is active. Gate the update on `!preempted_any` and add a regression test that forces KV preemption on an aligned pass before checking the next non-aligned pass.

crates/core/src/engine/generalized/engine.rs
```diff
@@ -355,6 +365,9 @@ impl<C: RankEngine> GeneralizedMockerEngine<C> {
             }
         }
         debug_assert!(pending_by_rank.next().is_none());
+        if self.ranks.iter().all(RankEngine::is_drained) {
```
jasonqinzhou: The wave counter resets only from `complete_pass`, but this engine can also become fully drained after `apply_command_effects` (for example, canceling the last request or releasing/canceling the last handoff) or `process_internal_work`. A later request then starts with the old `next_wave_step`, shifting which passes are cadence-aligned despite the documented fresh-wave reset. Please centralize the drained transition/reset across every mutating boundary, or reset lazily before the first pass after an idle period, and cover a cancel-to-idle-then-resubmit case.

jasonqinzhou (approved):
**APPROVE · 7.8/10 · high confidence**

## What this PR does

This PR adds a serialized vLLM prefill scheduling interval and drives it from a shared, wave-local attention-DP step counter so non-aligned passes prioritize decode work. It also models capacity-bound release, documents the new control and Kimi-K2.5 accuracy results, and adds focused config, group-step, and scheduler tests.

The backward-compatible default, rank-shared counter, explicit vLLM/attention-DP gating, and focused tests for chunked prefills, capacity release, and single-rank behavior make the intended fidelity change unusually easy to trace.

## Why this score

The 7.8 quality score reflects generally sound work with two non-blocking correctness issues in edge paths. The canonical recommendation is BLOCKED—not because of a confirmed P0/P1 issue, but because the required Fable lane requested claude-fable-5 and observed only claude-sonnet-4-6 and claude-haiku-4-5, so the dual-review contract cannot be satisfied. Independently, Codex found that KV preemption can incorrectly enable the capacity escape hatch and that cancellation/internal-work drain paths can retain a stale cadence phase.

crates/core/src/engine/scheduler/vllm/core.rs
```diff
@@ -1882,6 +1919,10 @@ impl VllmCore {
             }
         }
 
+        if !defer_prefills {
```
tedzhouhk: Upstream updates `prefill_capacity_bound` only when the waiting-scheduling block ran without preemption. Here an aligned pass that preempts can leave a request in `waiting_members`, flip this flag to true, and release prefills one step early on the next non-aligned pass. Please also gate this update on `!preempted_any` and cover that transition with a regression test.

crates/core/src/engine/scheduler/vllm/core.rs
```diff
@@ -1618,6 +1654,7 @@ impl VllmCore {
         };
         while !preempted_any
             && self.state.running.len() < max_num_running
+            && !defer_prefills
```
tedzhouhk: Gating the entire waiting loop diverges from upstream vLLM. On throttled steps vLLM still starts async KV loads and admits fully materialized/full-prefix-hit resumes; it only stops requests that still require local prefill compute. This can add artificial delay to P/D decode handoffs and full cache hits. Please move the cadence check into the admission path or document this as an intentional limitation.

crates/core/src/engine/generalized/engine.rs
tedzhouhk: This reset is earlier than upstream vLLM: `DPEngineCoreProc` advances real/dummy passes and checks global unfinished state only every 32 steps, resetting only after that collective confirms the wave is complete. A request arriving during the dummy tail can therefore observe a different cadence phase here. Please model that tail or explicitly document and test this approximation.

crates/core/src/engine/scheduler/vllm/core.rs
simone-chen: Addressed in 3f42fb6f. The update now requires both !defer_prefills and !preempted_any. The aligned-preemption to non-aligned regression covers this transition.

crates/core/src/engine/scheduler/vllm/core.rs
simone-chen: Addressed in 3f42fb6f. The cadence gate moved from the whole waiting loop to the local-compute admission point after connector load handling. Materialized and at-most-one-token prefix remainders can advance; a prefix-hit regression covers the latter.

crates/core/src/engine/generalized/engine.rs
simone-chen: Addressed in 3f42fb6f. The immediate-drain approximation and its difference from the 32-step upstream collective tail are now explicit in the doc. Existing drain coverage plus the new cancel-to-idle test pin the immediate reset behavior.

simone-chen: Addressed in 3f42fb6f. Wave reset is centralized and called after command effects, pass completion, and internal work. A cancel-to-idle-then-resubmit regression verifies the new wave restarts at step zero.

crates/core/src/engine/scheduler/vllm/core.rs
simone-chen: Addressed in 3f42fb6f. The capacity-bound flag update is gated on !preempted_any, with a regression that forces aligned-step KV preemption and verifies the victim remains deferred on the next non-aligned step.

dreamtalen (approved):
Approving the Rust logic. We also need to expose `prefill_schedule_interval` through the Python config so the standard `aisimulate predict` CLI can use this new field.

simone-chen:
Addressed the Python config gap in `09166e6e`:

- added positive `engine.workers.<role>.scheduler.prefill_schedule_interval` with default `1`;
- lowered it into the native rank used by the engine-backed `aisimulate predict` path;
- documented the CLI YAML form and added validation plus YAML-to-runtime regression coverage.

Validation: rebuilt the wheel from the PR head; 451 Python tests passed with 5 expected skips; Ruff checks pass.

PR #100, opened 2026-09-03
feat(perfmodel): add GB200 multi-node communication data

Arsene12358:
## Summary

- Continue [AIConfigurator #1593](https://github.com/ai-dynamo/aiconfigurator/pull/1593) at source head `aa4848d98d3ce14dba095a9a347ea7e877a6889c` in AISimulate, now the canonical repository.
- Add measured GB200 multi-node NCCL world-size 8/16 data and SGLang custom-all-reduce rows.
- Select the greatest recorded communication fan-out that is compatible with the requested topology, preserving exact TP8/TP16 behavior while making sparse TP6/TP12 requests deterministic.
- Add numeric TP16 coverage across static, mixed, aggregate, and disaggregated engine-step surfaces.

## Migration notes

AISimulate already contains the generalized recorded-slice selection introduced after the AIC branch point, so this port carries the measured data and adapts the selection rule instead of copying the superseded AIC implementation. Both parquet blobs are byte-identical to the source PR. The AIC synchronization ledger is intentionally unchanged because this is an isolated PR migration rather than a contiguous upstream sync.

## Tracking

Ignore AIC-1593

The branch name references archived AIConfigurator PR #1593; Linear issue AIC-1593 is unrelated to this migration.

## Validation

- Communication database tests: 15 passed
- Communication operator tests: 9 passed
- Engine-step parity: 280 passed
- Collector data invariants R1-R7: passed
- NCCL all-reduce/reduce-scatter/all-gather/all-to-all coverage includes TP6→4, TP12→8, exact TP8/TP16, and TP32→16 selection
- SGLang custom-all-reduce coverage includes TP6→4, TP12→8, exact TP8/TP16, and TP32→16 selection
- Parquet schema, duplicate-key, formatting, legal, and diff checks: passed
- [Trusted AISimulate CI](https://github.com/ai-dynamo/aisimulate/actions/runs/33766390288): 13/13 executable jobs passed across amd64 and arm64; the workflow-dispatch Application Wheel placeholder was skipped as expected

## Stack

This is the first PR in the migration stack. AISimulate merge order: [#100](https://github.com/ai-dynamo/aisimulate/pull/100) → [#103](https://github.com/ai-dynamo/aisimulate/pull/103) → [#101](https://github.com/ai-dynamo/aisimulate/pull/101) → [#102](https://github.com/ai-dynamo/aisimulate/pull/102).

The repository permits squash merges only. After each PR lands, rebase the next branch onto the updated `main` and retarget that PR to `main` before merging it.

PR #103, opened 2026-09-03
feat(models): add Muse Glimmer support

Arsene12358:
## Summary

- Continue [AIConfigurator #1516](https://github.com/ai-dynamo/aiconfigurator/pull/1516) at source head `de5388914532a68e7195dc15cdde7e9dfa557c79` in AISimulate.
- Add first-class `meta-models/Muse-Glimmer-30B` parsing, modeling, offline configuration, support-matrix coverage, and measured sliding-window attention data.
- Model Muse attention gates, row-parallel reductions, TP-sharded embeddings, and Q/K normalization in both Python and Rust engine paths.
- Preserve B300 as empirical fallback rather than mislabeling GB300 measurements as B300 silicon.

## Review concerns addressed

- Attention gate GEMMs and elementwise gating are modeled in both phases.
- All required row-parallel collectives are emitted and counted exactly.
- Context and generation Q/K normalization both contribute latency through the serialized schema-v16 Rust path.
- Embedding weights are sharded by tensor parallelism, preventing per-GPU memory overcounting.
- All 12 B300 support-matrix rows evaluate successfully through the documented empirical fallback; no GB300 rows are published as B300 data.

## Attribution

The bundled configuration is byte-identical to `meta-models/Muse-Glimmer-30B@f84ecc3a0ea984a4c04542a84269e3d065350a6e` and is covered by Apache-2.0 attribution in both packaged notice copies.

## Tracking

Ignore AIC-1516

The branch name references archived AIConfigurator PR #1516; Linear issue AIC-1516 is unrelated to this migration.

## Validation

- Full unit suite on the feature branch: 5,041 passed, 11 skipped
- Combined engine-step parity after stacking: 280 passed
- Compile/per-op parity: 64 passed
- Focused Muse, sweep, support-matrix, collector, and import suites: 211 passed
- Original post-stack integration command: 342 passed
- Rust Q/K-normalization regression, Ruff, formatting, SPDX/legal, JSON, and diff checks: passed
- [Trusted AISimulate CI](https://github.com/ai-dynamo/aisimulate/actions/runs/33768846902): 13/13 executable jobs passed across amd64 and arm64; the workflow-dispatch Application Wheel placeholder was skipped as expected

## Stack

Base: [AISimulate #100](https://github.com/ai-dynamo/aisimulate/pull/100). Merge order: [#100](https://github.com/ai-dynamo/aisimulate/pull/100) → [#103](https://github.com/ai-dynamo/aisimulate/pull/103) → [#101](https://github.com/ai-dynamo/aisimulate/pull/101) → [#102](https://github.com/ai-dynamo/aisimulate/pull/102).

The repository permits squash merges only. After each PR lands, rebase the next branch onto the updated `main` and retarget that PR to `main` before merging it.


python/aisimulate/src/aiconfigurator_core/sdk/models/muse_glimmer.py
```diff
@@ -0,0 +1,282 @@
...
+        d = self._resolve_dims(tp)
+
+        self.context_ops = [
+            ops.Embedding("context_embedding", 1, self._vocab_size // tp, h, 0.3),
+            ops.CustomAllReduce("context_embedding_ar", 1, h, tp),
+        ]
+        if counts["swa"] > 0:
+            self.context_ops.extend(self._attention_block("context", "swa", counts["swa"], cfg.sliding_window_size, d))
+        if counts["global"] > 0:
+            self.context_ops.extend(self._attention_block("context", "global", counts["global"], 0, d))
```
jasonqinzhou, line 185: The bundled config marks all 13 `full_attention` layers with `layer_rope_theta = 0`, but this call constructs the same `ContextAttention` used by the RoPE-enabled sliding layers. In `ContextAttentionOp::query`, `apply_rope` is added unconditionally; `window_size=0` only selects full attention and does not suppress that fused cost. Prefill predictions therefore charge RoPE for every global layer even though those layers are NoPE. Could we carry an explicit RoPE flag (or split the NoPE context path), set it false here, and add a regression showing that global layers omit the RoPE extra while sliding layers retain it?

jasonqinzhou (approved):
**APPROVE · 8.6/10 · medium confidence**

## What this PR does

Adds end-to-end Muse Glimmer 30B support: pinned config parsing, a dedicated dense sliding/global-attention model, support-matrix and measured-attention data, plus Rust/Python QK-normalization parity.

The hybrid layer accounting is unusually well documented and tested, with explicit operation-shape, KV-cache, context-parallel, schema-round-trip, and shipped-data coverage that makes this large model addition straightforward to audit.

## Why this score

Score 8.6/10 reflects the completed Codex lane, but the recommendation is BLOCKED because the required independent Claude Fable review timed out before returning a result or verifiable model-usage evidence. Codex identified one non-blocking modeling-accuracy issue: the 13 global NoPE layers reuse ContextAttention, whose runtime query still charges mandatory RoPE latency during prefill.

python/aisimulate/src/aiconfigurator_core/sdk/models/muse_glimmer.py
jasonqinzhou: Fixed in a0e265e: `ContextAttention` now carries an explicit `apply_rope` policy (defaulting to true), Muse Glimmer global NoPE layers disable it, and both analytic query paths omit the RoPE cost when disabled. The engine-spec schema is bumped to 17, with Rust SOL/query and Python model/wire/pickle regressions. Local Rust, Python, parity, wheel-install, source, and CODEOWNERS validation is green.

jasonqinzhou (approved):
Approved at exact head 8d1ceaa. The NoPE modeling issue is fixed with an explicit ContextAttention RoPE policy, schema/binding parity, and Rust/Python regressions; the review thread is resolved. Full manually dispatched CI run 33808615724 passed all 13 executable jobs across amd64 and arm64, with DCO and CODEOWNERS checks also green.

PR #105, opened 2026-09-03
feat(replay): support attention DP in disaggregated replay

dreamtalen:
## Summary

- enable request-level attention-DP routing in disaggregated replay
- support independent and asymmetric prefill/decode DP sizes across vLLM, SGLang, and TensorRT-LLM
- align authored rank hints with Dynamo: `dp_rank` selects decode/aggregated rank and `prefill_dp_rank` overrides prefill rank
- preserve both rank hints through Python configured inline traffic
- apply SGLang DP-attention launch normalization per rank exactly once
- retain the fail-closed native host-offload restriction
- document aggregate KV-transfer semantics and add nonzero-rank cleanup coverage

## Scope

Each request transfers from one concrete prefill `(worker, dp_rank)` to one concrete decode `(worker, dp_rank)`. KV transfer remains a single aggregate request-level event; rank-wise KV layout conversion and network contention are not modeled.

## Review follow-ups

- authored `P1 -> D3` routing is preserved, with independent P/D bounds validation and legacy `dp_rank` fallback
- Python inline traffic accepts and validates both rank hints as uint32
- SGLang divides `chunked_prefill_size` by ADP size and scales `schedule_conservativeness` by `0.3` before rank construction
- DP1-vs-DP4 long-prompt chunking and memory-pressure admission are covered

## Testing

- `cargo +1.93.1 test --workspace -q`
  - 1086 passed, 1 ignored
  - all integration tests passed
- `cargo +1.93.1 test --manifest-path crates/tests/public-api/Cargo.toml`
  - 4 passed
- targeted Python runner/search/traffic tests
  - 84 passed
- AISimulate replay-parity skill, pinned 5,000-row Mooncake corpus, baseline `d25f25de` vs candidate `0bae79b6`, two fresh processes per revision
  - vLLM disaggregated canonical digest matched: `267516ef9e9582270a10fab01c706a54e50a8082acda00d5613fd91de195bbcf`
  - SGLang disaggregated canonical digest matched: `130cad9ee779bc4456d3e2b0d7d84bf8c2675345565704d5e627dbc4e91d2733`
  - 5,000 completed requests and complete handoffs per row; no short outputs
- Rust fmt, Ruff 0.14.1 lint/format, and `git diff --check` passed

Performance was not measured; the replay-parity skill is a semantic qualification campaign.


python/aisimulate/src/aisimulate/runner.py
```diff
@@ -106,7 +106,7 @@ def capabilities(self) -> RunnerCapabilities:
         return RunnerCapabilities(
             replay_spec_api_version=1,
             supported_backend_topologies=_SUPPORTED_BACKEND_TOPOLOGIES,
-            supports_disaggregated_attention_dp=False,
+            supports_disaggregated_attention_dp=True,
```
tedzhouhk, line 109: [P1] Preserve authored rank identity in disaggregated replay

This advertises disaggregated attention-DP as supported, but both pools use `PoolRoundRobinPlacement`, whose `place()` reads only the request ID and never `preferred_dp_rank()`. A `ReplayRequest.dp_rank` captured from a real route is therefore silently replaced by round-robin, unlike aggregated replay and the explicit-rank APIs in vLLM, SGLang, and TensorRT-LLM. With asymmetric pools, one rank field is also insufficient to represent a concrete P-rank to D-rank pair. Please add stage-specific `prefill_dp_rank` / `decode_dp_rank` fields with independent range validation, or fail closed when an authored rank cannot be represented. A regression test should pin P1 to D3 and cover each side being out of range.

crates/core/src/replay/engine.rs
```diff
@@ -152,18 +152,6 @@ impl ReplayEngineConfig {
                                 .to_string(),
                         ));
                     }
-                    if role.dp_size != 1 {
```
tedzhouhk, line 155: [P1] Apply SGLang DP-attention normalization before removing this guard

For a SGLang role with `dp_size > 1`, the rank-local `SglangConfig::from_args` never receives the group DP size. Real SGLang DP-attention scales `schedule_conservativeness` by 0.3 and divides `chunked_prefill_size` by `dp_size`; replay therefore keeps DP=1 admission and chunking parameters on every rank, changing batching, retraction, and TTFT. Please thread the role DP size into SGLang configuration and apply the normalization exactly once, with a DP1-versus-DP4 long-prompt/memory-pressure test.

tedzhouhk (approved):

python/aisimulate/src/aisimulate/runner.py
dreamtalen: Addressed in 0bae79b6. ReplayRequest now adds Dynamo-compatible prefill_dp_rank while existing dp_rank remains the decode/aggregated rank and the prefill fallback. Disaggregated placement switches to the stage-specific hint, PoolRoundRobinPlacement honors pinned ranks across workers, and validation checks each hint against its role DP size. Python configured inline traffic also accepts, validates, and preserves both hints. Added regressions for explicit P1 -> D3 routing, each side out of range, legacy dp_rank fallback, aggregated rejection of prefill_dp_rank, and invalid Python rank values.

crates/core/src/replay/engine.rs
dreamtalen: Addressed in 0bae79b6. SGLang rank construction now mirrors handle_data_parallelism exactly once: chunked_prefill_size is divided by attention DP size and schedule_conservativeness is scaled by 0.3 before constructing each rank scheduler. Added DP1-vs-DP4 tests for the resolved controls, long-prompt chunking, and memory-pressure admission. The 5,000-row replay-parity campaign also produced byte-identical SGLang DP1 disaggregated canonical output against baseline.

PR #107, opened 2026-09-03
feat(afd): add A/F performance-model measurements [AIC-1905]

jasonqinzhou:
## Summary

- add a versioned `AFDMeasurementRequest` / `AFDPerformanceModel` boundary
- use AIC's public estimate API to produce full-precision A-pool, F-pool, A-to-F, and F-to-A per-layer measurements
- attach validated measurements and their units, model/hardware/backend identity, workload point, and source to `BackendDeploymentSpec`
- measure transfer latency without calibration and let the backend-neutral evaluator apply `comm_overhead_factor` exactly once
- reject missing, duplicate, or unexpected phase measurements and fail closed when AIC cannot measure a candidate
- document that this layer requires concrete synthetic `isl`/`osl`; trace-aware measurement remains part of the replay layer

## Stack

- layer 3 of 7 for AFD CLI and runtime support
- stacks on #21 through `jasonzho/afd-layer-2-sweeper-search`
- tracked by AIC-1905

This layer supplies the measurements required by AFD evaluation. It does not yet advertise a production AFD runner or execute the A/F lifecycle; those remain in #108 and #109.

## Validation

Exact head: `1a3db8815a8fe2e650b138993e87ad8dcce2af1c`

- Sweeper plus AFD phase-completion suites: 332 passed, 1 expected public-CLI xfail
- focused Ruff check and format check: passed
- `git diff --check`: clean
- live AIC-backed smoke: Qwen/Qwen3-32B on h200_sxm/vLLM 0.24.0 produced a finite decode measurement with 64 layers

## Known limitation

A single measurement cannot faithfully represent a trace containing different sequence lengths, so trace-only AFD sweeps fail closed in this layer. The staged replay work will own trace-aware execution semantics.

PR #21, opened 2026-08-19
feat(sweeper): integrate AFD search domains [AIC-1904]

jasonqinzhou:
## Summary

- add `afd` and `afd+pd` branches to the existing `SmartSearchConfig` and generic Sweeper path
- enumerate complete finite A/F topology domains and legal opposite-phase companion configurations
- carry typed AFD candidates through projection, sampling, GPU accounting, unrolling, and `BackendDeploymentSpec`
- enforce `afd_max_candidates` while generating the AFD x companion product instead of materializing an unbounded list
- mark AFD performance metadata as unresolved until the performance-model layer attaches measurements
- fail closed unless a runner advertises the exact backend plus `afd` or `afd+pd`
- document the internal Sweeper configuration and update the AIC migration guide without claiming public CLI support

## Stack

- layer 2 of 7 for AFD CLI and runtime support
- stacks on #17 through `jasonzho/afd-layer-1-sweeper-core`
- does **not** depend on #18; the required parallel-search abstractions are already on current `main`
- tracked by AIC-1904

This layer intentionally excludes public recommendation lowering, performance-model A/F measurements, Mocker/Replayer execution, production runner capability advertisement, and deployment artifacts.

The built-in engine runner therefore continues to reject AFD. A test-only runner that explicitly advertises `vllm/afd` verifies the complete Sweeper-to-`ReplaySpec` path.

## Validation

Exact head: `b6a89672a571c97becbbc0bcd0cfd720946a8689`

- `python -m ruff check` on every changed Python file: passed
- `python -m ruff format --check` on every changed Python file: passed
- `PYTHONPATH=python/aisimulate/src python -m pytest -q -p no:timeout tests/sweeper --ignore=tests/sweeper/test_packaging.py`: 303 passed, 1 expected xfail
- the expected xfail is the public `aisimulate recommend` AFD lowering reserved for a later layer
- `git diff --check jasonzho/afd-layer-1-sweeper-core..HEAD`: clean

## Next layer

AIC-1905 adds the performance-model measurements needed to produce `AFDLayerTimes` for real candidates.

PR #108, opened 2026-09-03
feat(afd): execute staged A/F full passes [AIC-1906]

jasonqinzhou:
## Summary

- add `AFDForegroundEngine` for deterministic, non-preemptive A/F full passes
- expand every layer and microbatch into explicit attention, A-to-F, FFN, and F-to-A intervals
- use the exact layer-1 optimistic, conservative, fallback, and serial cadence so staged execution and analytical evaluation share one pass boundary
- retain an in-flight pass until its modeled completion boundary and keep late caller wakeups from inflating modeled latency
- support separate prefill/decode passes on a shared both-phase A/F pool and the AFD side of `afd+pd`
- expose completion effects with exact sequence, token, latency, and pass identity accounting

## Stack

- layer 4 of 7 for AFD CLI and runtime support
- stacks on #107 through `jasonzho/afd-layer-3-perfmodel`
- tracked by AIC-1906

This layer owns analytical foreground-pass semantics only. PR #109 integrates the engine with workload replay, the companion phase, reporting, and production runner capability advertisement.

## Validation

Exact head: `53be191ed5ac6dc659d2ea5339a15e0cf0f1dea6`

- full Sweeper suite excluding packaging: 312 passed, 1 expected public-CLI xfail
- focused foreground-engine tests cover stage ordering, exact pass boundaries, optimistic fallback, shared both-phase pools, `afd+pd`, and invalid completion attempts
- focused Ruff check and format check: passed
- `git diff --check`: clean

## Review boundary

This is an analytical Mocker foreground engine. It does not claim physical backend serving support; the later runner layer consumes its deterministic schedule and measurements.


jasonqinzhou:
Self-review blocker: this PR currently has one signed empty commit and zero changed files. It does not execute staged attention → A-to-F → FFN → F-to-A work or verify pipeline lifecycle semantics. I am replacing the placeholder with the scoped Mocker/foreground-engine implementation and deterministic tests.

jasonqinzhou:
Follow-up: the zero-file blocker is resolved in 7df0d72. This PR now implements deterministic staged A/F foreground passes, exact agreement with the existing pipeline formula, completion-barrier semantics, both-phase pool reuse, `afd+pd` AFD-side handling, and focused tests. The full Sweeper suite is 312 passed with 1 expected later-layer CLI xfail. I kept the PR draft for your manual review and explicitly scoped physical runner support to #109.

python/aisimulate/src/aisimulate/sweeper/afd_engine.py
```diff
@@ -0,0 +1,361 @@
...
+            (AFDStage.ATTENTION, float(measurement.attention_ms)),
+            (AFDStage.A_TO_F, float(measurement.a_to_f_ms) * self._topology.comm_overhead_factor),
+            (AFDStage.FFN, float(measurement.ffn_ms)),
+            (AFDStage.F_TO_A, float(measurement.f_to_a_ms) * self._topology.comm_overhead_factor),
+        )
+        intervals: list[AFDStageInterval] = []
+        unit_index = 0
+        for layer in range(measurement.num_layers):
+            for microbatch in range(self._topology.num_microbatches):
+                cursor = now + unit_index * evaluation.cycle_ms * latency_correction
```
Arsene12358, line 296: **[P1] Enforce the previous-layer dependency before starting this microbatch**

With one microbatch, this cadence can start the next layer before the same microbatch's previous F→A transfer returns. For two layers with 1 ms in each of attention, A→F, FFN, and F→A, the conservative schedule starts layer 1 at 2 ms while layer 0 returns at 4 ms, and completes the pass at 6 ms despite an 8 ms dependency minimum. Requested optimistic execution falls back to the same invalid schedule; serial execution correctly takes 8 ms. This directly affects the completion boundary consumed by #109 and the latency/throughput reported by the public CLI.

| Current | Required |
| --- | --- |
| Every layer/microbatch starts at `unit_index * cycle_ms`; 6 ms completion in this case. | Gate attention on the same microbatch's previous-layer return and resource availability; completion must be at least 8 ms here. |

Please derive completion from a causal schedule, or explicitly reject/force serial execution for unsupported microbatch counts. This finding concerns the new staged engine, not a separate request to fix the pre-existing legacy estimator.

Reproduction from a checkout with this PR applied:

```python
from aisimulate.sweeper.afd_engine import AFDForegroundEngine
from aisimulate.sweeper.afd_parallel import AFDTopology
from aisimulate.sweeper.afd_perfmodel import AFDLayerTimes

topology = AFDTopology(
    n_a_nodes=1, n_f_nodes=1, gpus_per_node=1, tp_a=1,
    a_batch_size=1, num_microbatches=1, phase="decode",
    combined_with_pd=False, pipeline_model="conservative",
)
times = AFDLayerTimes(
    phase="decode", attention_ms=1.0, ffn_ms=1.0,
    a_to_f_ms=1.0, f_to_a_ms=1.0, num_layers=2,
)
planned = AFDForegroundEngine(topology, (times,)).execute_pass(
    phase="decode", now_ms=0, input_length=8, output_length=3,
)
assert planned.intervals[4].start_ms >= planned.intervals[3].end_ms
# FAIL: 2.0 >= 4.0
assert planned.end_ms >= 8.0
# Actual: 6.0
```

Arsene12358 (changes requested):
Request changes based on a holistic review of the seven-PR AFD stack (#17 → #21 → #107 → #108 → #109 → #110 → #111).

This PR introduces a causal dependency violation in the staged foreground engine: a valid one-microbatch configuration starts the next layer before the previous layer returns and reports a physically impossible pass completion time. The attached inline finding contains a standalone reproduction, observed/required behavior, and a suggested fix. Please fix this before the downstream runner and CLI layers are merged.

Validation on combined stack head `5e5d105abb45b17d7cd73ca971fcd62ce26fe7d0`: root suite **579 passed, 5 skipped**; generator plus AFD phase-completion suites **369 passed, 4 skipped**. Five independent expectation-based checks expose four distinct defects and fail again in the freshly installed locked environment. Passing existing tests does not cover these invariants.

The three independent runner blockers introduced in #109 are being reported on that PR. The findings are attached to their introducing PRs; disclosed lack of physical serving is not being filed as an additional code blocker.

python/aisimulate/src/aisimulate/sweeper/afd_engine.py
Arsene12358: Fixed at `40026249de897576a5527cdc81de51ced3eada6d`. One microbatch now uses an effective serial schedule while retaining the requested pipeline model in provenance. The two-layer, four-stage 1 ms reproduction completes at 8 ms and never starts attention before the prior F-to-A return. Independent review reproduced the original failure, verified the new causal invariants, and compared 4,536 multi-microbatch cases without behavior changes.

Arsene12358 (approved):
The stage-causality blocker is fixed and independently verified. A single microbatch with two layers and four 1 ms stages now takes 8 ms; the next attention stage waits for returned FFN output. Requested scheduling provenance is retained, and 4,536 independent multi-microbatch comparisons preserve prior valid behavior. The complete stack passed Full CI: https://github.com/ai-dynamo/aisimulate/actions/runs/34564878459. The refreshed head preserves the reviewed source tree. This approval supersedes my earlier Request Changes; fresh PR checks must pass before merge.

PR #109, opened 2026-09-03
feat(afd): add Replay lifecycle and analytical runner [AIC-1907]

jasonqinzhou:
## Summary

- teach `EngineReplayRunner` to execute measured AFD workloads through the staged foreground engine instead of flattening them to aggregate or disaggregated replay
- advertise `afd` and `afd+pd` for vLLM, SGLang, and TRT-LLM only after that executable path exists
- require pure AFD to cover both phases and require single-phase AFD to carry the opposite P/D companion
- schedule fixed-length synthetic arrivals into bounded batches, preserve queueing in TTFT/E2E, and overlap independent AFD/P-D pools
- measure the regular companion from an explicit fixed model or AIC static estimate
- report throughput, SLA goodput, latency, GPU-hours, exact GPU accounting, lifecycle counts, and optional per-request records
- expose the canonical `native_report` required by public CLI raw/per-request output
- fail closed for unresolved measurements, trace or random-length reuse, simulation deadlines, wrong companion phases, and inconsistent GPU provenance
- explicitly route all `tests/test_afd_*.py` files through replay, Sweeper, FPE, and maintainer code owners

## Stack

- layer 5 of 7 for AFD CLI and runtime support
- stacks on #108 through `jasonzho/afd-layer-4-mocker`
- tracked by AIC-1907

This is the production **analytical** AISimulate runner path. It deliberately does not claim physical backend AFD launch support and does not route AFD through the existing aggregate/disaggregate native runtime.

## Validation

Exact head: `2f28bf52d1c518fa6229a7041c4b7e0287aa1b4e`

- runner plus full Sweeper suites excluding packaging: 361 passed, 1 expected public-CLI xfail
- focused native-report regression: 18 passed
- end-to-end Sweeper -> measured ReplaySpec -> EngineReplayRunner test: passed
- focused tests cover pure both-phase AFD, overlapping `afd+pd`, fixed companion timing, SLA goodput, capability gates, unresolved input, deadlines, GPU-accounting mismatches, and public CLI report compatibility
- CODEOWNERS full-tree coverage: 100%; ownership tests: 99 passed
- focused Ruff check and format check: passed
- `git diff --check`: clean

## Current replay limits

- fixed synthetic positive `isl` and `osl` only
- `random_range_ratio` must be 1.0
- trace replay and `max_sim_time_ms` fail closed
- physical serving configuration remains an artifact/backend concern rather than an analytical-runner claim


jasonqinzhou:
Self-review blocker: this PR currently has one signed empty commit and zero changed files. It does not add an AFD replay topology, runner capability, lifecycle, or production execution path. I am replacing the placeholder with the scoped Replay/runner implementation and end-to-end tests.

jasonqinzhou:
Follow-up: the zero-file blocker is resolved in 6944f05. The engine runner now has an explicit analytical AFD path with staged foreground execution, pure-both and AFD+P/D lifecycle semantics, companion timing, arrivals/queueing, standard metrics, SLA goodput, exact GPU accounting, capability advertisement, and fail-closed unsupported cases. Runner plus Sweeper validation is 361 passed with 1 expected later-layer CLI xfail. I kept the PR draft and made the analytical-vs-physical support boundary explicit for your manual review.

jasonqinzhou:
Follow-up self-review found one public-CLI integration defect: analytical AFD reports exposed only internal `afd_report`/`per_request` metadata, so `aisimulate predict --capture-per-request` could not consume the runner’s canonical `native_report`. Fixed in signed commit `48fff86`: AFD now preserves its internal metadata and also emits the same `native_report.summary` and `native_report.per_request` contract used by the CLI. Focused validation: `18 passed` across `tests/test_afd_runner.py` and `tests/test_unified_cli.py`.

jasonqinzhou:
CI follow-up: the live codeowners check found `tests/test_afd_runner.py` falling through to the catch-all. Fixed at exact head `d15fbe6` by adding one generated ownership rule for `tests/test_afd_*.py`, shared by Replay, Sweeper, FPE, and maintainers. Full-tree coverage is now 100%, all 99 ownership tests pass, and the fix is merged forward through PRs #110 and #111.

python/aisimulate/src/aisimulate/runner.py
```diff
@@ -196,6 +349,342 @@ def close(self) -> None:
...
+            raise ValueError("afd+pd replay requires one concrete AFD phase")
+        companion = (companion_model or AICAFDCompanionPerformanceModel()).measure(spec)
+        expected_companion = AFDPhase.DECODE if topology.phase is AFDPhase.PREFILL else AFDPhase.PREFILL
+        if companion.phase is not expected_companion:
+            raise ValueError(
+                f"AFD companion phase {companion.phase.value!r} does not complement {topology.phase.value!r}"
+            )
+        batch_capacity = min(topology.total_batch_size, companion.total_batch_capacity)
+    if max_in_flight is not None:
+        batch_capacity = min(batch_capacity, max_in_flight)
```
Arsene12358, line 596: **[P1] Enforce closed-loop admission globally across the AFD and companion pools**

This limits only each batch's size. The next batch can enter prefill while the prior batch remains in decode, exceeding the configured deployment-wide concurrency. Synthetic closed-loop requests also retain arrival-time-zero placeholders instead of receiving replacement arrivals when an active request finishes.

Using the existing runner fixtures with concurrency 2 and four requests: the first batch needs 3 ms prefill plus 4 ms decode and completes at 7 ms. The next two requests must start at 7 ms, yielding 14 ms total. This runner instead starts their prefill at 3 ms and finishes everything at 11 ms, with four requests active simultaneously. Pure AFD also inflates mean TTFT from 3 ms to 7.5 ms by treating replacement requests as queued since time zero.

| Current | Required |
| --- | --- |
| Per-batch concurrency cap; AFD+P/D duration 11 ms; pure-AFD replacement arrivals remain zero. | One global in-flight count; completion-driven replacement arrivals; AFD+P/D duration 14 ms in this fixture. |

Please share admission state across both pools and release a slot only when the request completes. Preserve independent-pool overlap when the global admission limit permits it. This is the documented [closed-loop traffic contract](https://github.com/ai-dynamo/aisimulate/blob/5e5d105abb45b17d7cd73ca971fcd62ce26fe7d0/docs/cli/migrate-from-aiconfigurator.md#L239).

Reproduction using this PR's existing fixtures:

```python
import runpy
from aisimulate.runner import EngineReplayRunnerFactory

h = runpy.run_path("tests/test_afd_runner.py")
spec = h["_spec"](
    h["_topology"](phase="prefill", combined_with_pd=True),
    companion_role="decode",
)
spec.workload.update(
    source_type="synthetic", load_type="concurrency", request_count=4,
)
report = EngineReplayRunnerFactory().create(0).run(spec)
assert report.metrics["duration_ms"] == 14.0
# FAIL: actual duration is 11.0 ms.
```

python/aisimulate/src/aisimulate/runner.py
```diff
@@ -196,6 +349,342 @@ def close(self) -> None:
...
+            afd_intervals += interval_count
+
+        for request in batch:
+            arrival = float(request["arrival_time_ms"])
+            request_records.append(
+                {
+                    "id": str(request["id"]),
+                    "arrival_time_ms": arrival,
+                    "ttft_ms": prefill_end - arrival,
+                    "tpot_ms": decode_step_ms,
```
Arsene12358: **[P1] Include destination queueing in TPOT before applying the ITL SLA**

`decode_start = max(prefill_end, companion_available_ms)` (and the opposite-phase equivalent) can add a wait after the first output token, but this line reports only one decode step's service time as TPOT. The resulting request metric, summary, and SLA goodput exclude that queueing.

An independent open-loop reproduction reports TPOT `[10, 10, 10, 10]` ms while the emitted first/last-token timings require `[10, 18.5, 18.5, 27]` ms. An 11 ms ITL SLA consequently qualifies all four requests instead of one, inflating the goodput used to rank recommendations. Open-loop traffic isolates this from the concurrency defect.

| Current | Required |
| --- | --- |
| `tpot_ms = decode_step_ms`; 4/4 requests pass the 11 ms SLA. | Derive average TPOT from first/last-token timestamps including decode queueing; only 1/4 passes. |

Please use `(decode_end - prefill_end) / (output_length - 1)` for multi-token output, or equivalent actual token timestamps, and feed the same semantics into summaries and SLA checks. Preserve the single-output-token behavior. This matches the existing [native report/SLA contract](https://github.com/ai-dynamo/aisimulate/blob/5e5d105abb45b17d7cd73ca971fcd62ce26fe7d0/crates/core/src/replay/report.rs#L711).

Reproduction using this PR's fixtures:

```python
from dataclasses import replace
import runpy
from aisimulate.runner import EngineReplayRunnerFactory
from aisimulate.sweeper import ReplayOutputRequirements

h = runpy.run_path("tests/test_afd_runner.py")
spec = h["_spec"](
    h["_topology"](phase="prefill", combined_with_pd=True),
    companion_role="decode",
)
spec.workload.pop("concurrency")
spec.workload.update(
    source_type="synthetic", load_type="constant_rate",
    arrival_interval_ms=0.01, request_count=4,
)
spec = replace(spec, concurrency=None)
spec.backend_deployment.decode_engine_args["timing_model"]["decode_ms"] = 10.0
spec.goal["sla"] = {"itl_ms": 11.0}
report = EngineReplayRunnerFactory().create(0).run(
    spec, output_requirements=ReplayOutputRequirements(capture_per_request=True),
)
assert report.metrics["goodput_completed_requests"] == 1.0
# FAIL: actual value is 4.0.
```

python/aisimulate/src/aisimulate/runner.py
```diff
@@ -196,6 +349,342 @@ def close(self) -> None:
...
+        batch_capacity = min(topology.total_batch_size, companion.total_batch_capacity)
+    if max_in_flight is not None:
+        batch_capacity = min(batch_capacity, max_in_flight)
+    batch_capacity = _positive_int(batch_capacity, "AFD replay batch capacity")
+
+    ordered_requests = sorted(requests, key=lambda request: float(request["arrival_time_ms"]))
+    request_records: list[dict[str, JSONValue]] = []
+    next_request = 0
+    afd_available_ms = 0.0
+    companion_available_ms = 0.0
```
Arsene12358, line 606: **[P1] Honor or explicitly reject the companion's configured startup delay**

Both pools become available at zero here, irrespective of the companion's `startup_time`. The public configuration added by #110 accepts `workers.prefill.startup_seconds: 10` and compilation preserves it as `startup_time: 10.0`; the new AFD runner silently ignores it. On the combined stack, a valid fixed-timing prefill companion with that 10-second startup completes all four requests in 107.6 ms, before it can start serving. TTFT, duration, and SLA results therefore describe a different configuration.

| Current | Required |
| --- | --- |
| `companion_available_ms = 0.0`; accepted 10-second startup has no effect. | No companion service before 10,000 ms, or an explicit unsupported-option error. |

Please initialize availability using the companion's startup delay with consistent units, or reject nonzero startup if it is outside the analytical runner's supported lifecycle. AIC-1907 explicitly includes pool startup in its acceptance criteria.

Public-boundary reproduction from the combined stack checkout (including #110):

```python
import runpy
from aisimulate.compiler import prediction_to_replay_spec
from aisimulate.config import CorePredictionConfig
from aisimulate.runner import EngineReplayRunnerFactory

h = runpy.run_path("tests/test_afd_cli.py")
raw = h["_pure_prediction"]()
raw["engine"]["afd"].update(phase="decode", combined_with_pd=True)
raw["engine"]["workers"] = {
    "prefill": {
        "timing": {"type": "fixed", "prefill_ms": 2.0, "decode_ms": 2.0},
        "startup_seconds": 10,
    },
}
spec = prediction_to_replay_spec(
    CorePredictionConfig.model_validate(raw),
    afd_performance_model=h["_AFDPerformanceModel"](),
)
assert spec.backend_deployment.prefill_engine_args["startup_time"] == 10.0
report = EngineReplayRunnerFactory().create(0).run(spec)
assert report.metrics["duration_ms"] >= 10_000
# FAIL: actual duration is 107.6 ms.
```

Arsene12358 (changes requested):
Request changes based on a holistic review of the seven-PR AFD stack (#17 → #21 → #107 → #108 → #109 → #110 → #111).

Three reproduced P1 blockers are introduced by this runner and remain observable through #110/#111:

1. Closed-loop concurrency is enforced per batch instead of globally across both pools, and replacement arrival times are incorrect.
2. Decode queueing is excluded from TPOT and the ITL SLA, causing incorrect goodput and recommendation ranking.
3. Accepted companion startup delays are ignored.

Each inline finding includes reproduction code, observed versus required behavior, and a concrete fix direction. Please add independent contract tests for these cases and propagate the fixes through the stack before merging the public CLI/qualification layers.

Validation on combined stack head `5e5d105abb45b17d7cd73ca971fcd62ce26fe7d0`: root suite **579 passed, 5 skipped**; generator plus AFD phase-completion suites **369 passed, 4 skipped**. Five independent expectation-based checks expose four distinct defects and fail again in the freshly installed locked environment. Passing existing tests does not cover these invariants.

The separate staged-engine dependency blocker is reported on #108, where it originates. The broader Linear/native scope mismatch is distinct from these demonstrated analytical correctness defects.

python/aisimulate/src/aisimulate/runner.py
```diff
@@ -93,6 +102,135 @@ def run_replay_json(self, execution_spec_json: str) -> str:
...
+            "isl": isl,
+            "osl": osl,
+            "batch_size": batch_capacity,
+            "tp_size": _positive_int(parallel.get(f"{prefix}tp"), f"parallel_config.{prefix}tp"),
+            "pp_size": _positive_int(parallel.get(f"{prefix}pp", 1), f"parallel_config.{prefix}pp"),
+            "attention_dp_size": _positive_int(
+                parallel.get(f"{prefix}attention_dp", 1),
+                f"parallel_config.{prefix}attention_dp",
+            ),
+        }
```
Arsene12358, line 210: [P1] Preserve the companion forward-model selector

The public worker setting `timing.forward_model: fpm` is retained as `aic_forward_model` in the companion engine arguments, but these estimator kwargs never forward it to `cli_estimate`. That API defaults to `op_level`, so the prediction can succeed using a different model from the one requested. I reproduced the missing keyword for a decode companion, and independent review reproduced a real prefill prediction succeeding with op-level data where explicitly requesting FPM raises `PerfDataNotAvailableError`.

Current behavior: an FPM request can silently produce op-level timing. Required behavior: preserve and validate the requested selector through companion measurement, report it in provenance, and fail explicitly when its data is unavailable. Add regression coverage for both companion phases and the public lowering path.

Arsene12358 (changes requested):
Additional P1 found during the final holistic review: the AFD companion estimator drops an explicitly requested forward model. This is introduced by the new AICAFDCompanionPerformanceModel in this PR and is exposed by the public CLI layer. Please address it along with the previously requested replay fixes.

python/aisimulate/src/aisimulate/runner.py
Arsene12358: Fixed at `d7f7a546d39ac3b11bee5a7935744af2b1bf58b0`. Closed-loop client slots are released at end-to-end completion across both pools, and replacement arrival times use those completions. The c=2 four-request reproduction now takes 14 ms; the pure-AFD replacement arrivals are `[0, 0, 9, 9]`. Independent review checked 840 scheduling cases, including partial batches, both companion roles, and preserved open-loop overlap.

python/aisimulate/src/aisimulate/runner.py
Arsene12358: Fixed at `d7f7a546d39ac3b11bee5a7935744af2b1bf58b0`. TPOT now includes the decode queue after the first output token: `(completion - prefill_completion) / (OSL - 1)`, with zero for OSL=1. The deterministic queued reproduction now reports `[10, 18.5, 18.5, 27]` ms and only 1 of 4 requests meets the 11 ms ITL bound. Unit tests, independent oracles, and the real public CLI scenario passed.

python/aisimulate/src/aisimulate/runner.py
Arsene12358: Fixed at `d7f7a546d39ac3b11bee5a7935744af2b1bf58b0`. Companion availability now uses the existing startup-seconds conversion. Both prefill and decode companion startup scenarios passed through the real public CLI; initial requests wait for the configured 10 seconds, while closed-loop replacements arrive after completion. OSL=1 completes at prefill and does not wait for an unused decode pool.

python/aisimulate/src/aisimulate/runner.py
Arsene12358: Fixed at `d7f7a546d39ac3b11bee5a7935744af2b1bf58b0`, with public-boundary coverage in #110. The companion validates and forwards `aic_forward_model`, defaults to `op_level` only when absent, and records the chosen model in measurement provenance. Independent real-AIC checks for both roles now raise `InvalidRunnerError` caused by missing FPM data, with one estimator call and no fallback. Explicit and default op-level runs retain identical metrics. Both original selector-propagation oracles are green.

Arsene12358 (approved):
All four replay blockers from my Request Changes reviews are fixed: client slots are released at end-to-end completion across both pools; TPOT includes waiting after prefill; configured companion startup is honored in seconds; and explicit FPM selection reaches AIC without silent op_level fallback. Independent red/green regressions, 840 scheduling oracle cases, five public CLI scenarios, and six real AIC scenarios verified these contracts. The full stack passed Full CI: https://github.com/ai-dynamo/aisimulate/actions/runs/34564878459. This refreshed head preserves the reviewed source tree. This approval supersedes my earlier Request Changes reviews; fresh PR checks must pass before merge.

PR #115, opened 2026-09-04
feat(replay): add direct Weka AgentX ingestion and lane controls

Harrilee:
#### Overview:

Make AISimulate the direct entry point for published Weka AgentX traces. Weka JSON or JSONL is lowered inside AISimulate to the existing typed Agentic Mooncake v2 representation, validated as a ValidatedAgenticGraph, and executed through the existing agentic workload driver.

This PR completes the AISimulate scope of AIC-1810 and AIC-1814. It deliberately keeps qualification at functional_only: scheduler semantics, failure propagation, and the final AgentX runtime checkpoint remain follow-up work under AIC-1890, AIC-1816, and AIC-1815.

#### Details:

- Add a Dynamo-neutral WekaImporter and load_weka_agentic_graph convenience API in aisimulate-core.
- Accept one Weka JSON/JSONL file or a recursively traversed directory, with deterministic corpus ordering and provenance.
- Accept one corpus-wide nested timestamp basis (`auto`, `absolute`, or `relative`). Auto scans every nested request in every deterministic JSON/JSONL row: if any `child.t < marker.t - epsilon`, the entire corpus is relative; otherwise it is absolute. Both lower uniformly to root-absolute canonical IR. This is explicitly reported as a heuristic because malformed absolute input or unusually large relative offsets can be misclassified; ambiguous producers should select a basis explicitly. The resolved basis is logged, reported, and included in source identity.
- Fail closed on symlinks, empty corpora, invalid or mixed block sizes, malformed timing, unsupported hash scope/status, ambiguous ownership, and source changes after preflight.
- Preserve main-agent chains, explicit and flattened subagents, fan-out/fan-in, replay barriers, typed dependency triggers, delays, and local hash namespaces.
- Preserve optional recorded_api_time_ms in Mooncake v2 rows and AgenticNode without changing old fixture digests.
- Expose format: weka and agentic_lanes through YAML, Python schema/compiler, runner capabilities, and the native runtime.
- Keep source and engine block sizes separate and enforce an explicitly configured source block size only as an assertion.
- Restrict Weka and Agentic Mooncake v2 to trace_timestamps and aggregated topology for this checkpoint.
- Add functional_only input-format and lane audit fields to reports.
- Document AISimulate as the Weka ingestion owner and Dynamo as an optional downstream integration.

#### Where should the reviewer start?

- crates/core/src/replay/loadgen/weka.rs contains corpus preflight, parser, lowering, provenance, and parity tests.
- crates/core/src/python.rs contains the native runtime dispatch and validation boundary.
- python/aisimulate/src/aisimulate/config/traffic.py and sweeper/replay.py contain the public schema and capability contract.
- tests/e2e/configs/unified_cli/predict/engine/11-trace-weka-agentic-lane.yaml and 12-trace-weka-jsonl-agentic-lane.yaml cover JSON-directory and multi-row JSONL entry points.

#### Related Issues:

- AIC-1810
- AIC-1814
- Follow-up checkpoint: AIC-1815
- Follow-up runtime semantics: AIC-1890, AIC-1816

#### Validation:

- Latest-rebase Rust validation: `cargo check -p aisimulate-core --lib` passed; replay/loadgen tests: 100 passed; focused Weka importer/lowering tests: 25 passed.
- Latest-rebase Python schema, compiler, runner, and native runtime tests: 129 passed; default Weka CLI JSON/JSONL cases: 2 passed.
- Public two-row qualification: passed with corpus-wide timestamp basis resolved as `absolute`.
- CODEOWNERS policy tests: 99 passed; strict changed-tree coverage is 100%.
- Packaged legal-file consistency, Ruff format/lint, cargo fmt, and git diff checks: passed.
- Two pinned rows from semianalysisai/cc-traces-weka-062126-256k: 2 plays, 162 requests, resolved basis `absolute`, source digest c602efa3abcb3ad29811d916de54a0bc1e62b8ce30c4b8d2f9a20b8628df9bf2, graph digest 44c30a0d65d647a4abd3fd06fe0a843d599b064849053801d9a4b4be46924c3d.
- The public-sample qualification fetches only two rows into a temporary directory; no full dataset is downloaded or retained.
- No Dynamo dependency was added to AISimulate.


crates/core/src/replay/loadgen/weka.rs
```diff
@@ -0,0 +1,2102 @@
...
+
+struct LoweredTrace {
+    rows: Vec<AgenticMooncakeRow>,
+    raw_zero_outputs: usize,
+}
+
+fn lower_trace(trace: &WekaTrace, relative_path: &str) -> Result<LoweredTrace> {
+    validate_trace_header(trace, relative_path)?;
+    let namespace = namespace(relative_path);
+    let play_id = format!("{namespace}:play:{}", trace.id);
```
jasonqinzhou, line 701: `AgenticTraceBuilder::finish` sorts plays by `play_id` before `WorkloadDriver` assigns them round-robin to lanes, but this ID begins with `namespace(relative_path)`, which is a BLAKE3 digest. That discards the sorted file/JSONL order: an eight-row JSONL probe was scheduled as rows 3, 0, 6, 5, 7, 1, 4, 2. With `agentic_lanes`, this changes both per-lane membership and serial execution order, so replay no longer follows the authored corpus sequence. Could we carry an explicit source ordinal through graph construction (or otherwise make the canonical play order preserve traversal/row order) and assert the original play IDs in the lane test rather than sorting the emitted hashed IDs?

crates/core/src/replay/loadgen/weka.rs
```diff
@@ -0,0 +1,2102 @@
...
+            if digest != file.digest {
+                bail!(
+                    "Weka source {} changed after preflight: expected {}, found {}",
+                    file.relative_path,
+                    file.digest.to_hex(),
+                    digest.to_hex()
+                );
+            }
+        }
+        for file in &self.files {
```
jasonqinzhou: The digest pass above is separate from this later reopen-and-parse pass, so the bytes being emitted are not actually bound to the preflight digest. A two-file probe changed `z.json` from the callback after the first row was emitted; `for_each_row` then returned `Ok` with models `["original", "changed"]` while both `summary.header.source.digest` and the importer's digest remained the original value. The same window exists for any concurrent atomic replacement, violating the advertised fail-closed source-change/provenance contract. Please parse from a stable snapshot/handle whose bytes are covered by the digest, or otherwise finish validating the exact bytes before invoking the caller's sink, and add a regression that mutates a later file from the callback.

jasonqinzhou (dismissed):

## What this PR does

This PR makes AISimulate a direct entry point for published Weka AgentX JSON, JSONL, and directory corpora, lowers them into the typed Agentic Mooncake v2 graph, and exposes agentic lane controls through the unified Python and native replay paths.

The change does careful work around typed graph validation, direct-versus-materialized parity, public-sample qualification, deterministic hashing, and immutable upstream attribution, while honestly labeling the current prediction boundary as functional-only.

## Why this score

This packet is blocked because the helper did not verify that Claude Fable served the independent review; it observed claude-sonnet-4-6 and claude-haiku-4-5 instead. Independently, Codex scored the technical change 6.5 and would request changes for two confirmed P1 issues: lane assignment loses authored corpus order, and emitted rows can diverge from the preflighted source digest. No review should be submitted from this packet.

crates/core/src/replay/loadgen/weka.rs
Harrilee: Fixed in cfdbcf56. Weka now assigns a zero-based source_play_ordinal during deterministic file/JSONL traversal, carries it through Agentic Mooncake v2 rows and graph nodes, and validates that ordered graphs have one unique contiguous ordinal per play before sorting plays by that field. Because the ordinal is serialized, direct Weka and materialized v2 retain identical graph identity and lane order. I also changed the Python E2E to assert authored play IDs instead of sorting hashed IDs and added an eight-row JSONL regression that executes all plays through a single-lane WorkloadDriver in original order.

crates/core/src/replay/loadgen/weka.rs
Harrilee: Fixed in cfdbcf56. WekaImporter::open now copies each source into a stable snapshot while hashing those exact raw bytes, parses only that snapshot, validates the lowering, and writes canonical rows to a private disk-backed spool. for_each_row streams only the preflighted spool and never reopens an original source, so caller or concurrent source mutations cannot diverge emitted rows from header provenance while memory remains bounded to one play. The new regression mutates z.json from the callback after the first row and verifies that every emitted row still matches the preflight snapshot.

Harrilee:
@jasonqinzhou Rebased this PR onto the latest main (a95b3e59) and resolved the conflict with the new online prediction contract by retaining supported_execution_modes alongside the AgentX trace-format, lane, topology, and qualification capabilities. The two requested correctness fixes remain intact on the new head ad181bf3: source_play_ordinal preserves authored corpus order through lanes/materialized v2, and Weka emission reads only the preflighted stable spool. Post-rebase validation passed: 1,212 Rust tests (1 ignored), 497 Python tests (5 skipped), the 128 focused runner/config/runtime tests, and the two-row public AgentX qualification (2 plays / 162 requests). Could you please take another look when convenient?

crates/core/src/replay/loadgen/weka.rs
```diff
@@ -0,0 +1,2194 @@
...
+        bail!("Weka trace {} has invalid request timestamp", relative_path);
+    }
+    if request.model.trim().is_empty() {
+        bail!("Weka trace {} has an empty request model", relative_path);
+    }
+    if request.input_length == 0 {
+        bail!("Weka trace {} has a zero-length request", relative_path);
+    }
+    if request
+        .api_time
```
PeaBrane, line 1584: Could we preserve the fail-closed invariant from Dynamo #13586 here? `api_time` is graph-shaping: `request_end` feeds sequence, spawn, join, and seam classification, so absent/null is not equivalent to an explicit zero-duration request. This exact rewrite was reproduced in [the Dynamo review](https://github.com/ai-dynamo/dynamo/pull/13586#discussion_r3927795984) and fixed before merge. Please require presence while retaining explicit zero, cover absent/null/zero separately, and leave an invariant comment such as `// api_time determines dependency trigger and relation; absence is not a zero-duration request.` The current behavior matches the pre-review Dynamo implementation, not the cited final source revision.

crates/core/src/replay/loadgen/weka.rs
```diff
@@ -0,0 +1,2194 @@
...
+        let spawn_source_id = owner_request.source_id.clone();
+
+        let mut inner = Vec::with_capacity(subagent.requests.len());
+        for (inner_index, entry) in subagent.requests.iter().enumerate() {
+            let mut request = match entry {
+                WekaInnerEntry::Normal(request) | WekaInnerEntry::Streaming(request) => {
+                    request.clone()
+                }
+            };
+            if request.t + JOIN_EPSILON_SECONDS < subagent.t {
```
PeaBrane: Could we keep published Weka `t` values absolute trace-relative and reject an inner request before its subagent marker instead of guessing that it is relative? Dynamo removed this same heuristic after checking 129 sampled subagent groups; none preceded its marker ([resolution](https://github.com/ai-dynamo/dynamo/pull/13586#discussion_r3929022531)). Rebasing here changes `not_before_ms`, dependency triggers/delays, and graph identity. Please restore that rejection test and add `// Weka nested request timestamps are absolute trace-relative; do not rebase them on subagent.t.` so the deliberate contract is not generalized away again.

crates/core/src/python.rs
```diff
@@ -645,18 +661,61 @@ fn build_runtime_input(
...
+            ensure!(
+                traffic.max_sim_time_ms.is_none(),
+                "Weka agentic trace does not support max virtual time"
+            );
+            ensure!(paths.len() == 1, "weka requires exactly one path");
+            ensure!(
+                traffic.replay_concurrency.is_none(),
+                "Weka agentic trace does not support concurrency load"
+            );
+            let trace = load_weka_agentic_graph(&paths[0], traffic.trace_block_size)?
```
PeaBrane: Keeping multiple source models in the neutral importer is useful and matches the public corpus. At this execution boundary, however, the graph is handed to `WorkloadDriver`, which drops `AgenticNode.model`, while one configured timing model drives every request. Could this boundary either reject mixed-model graphs until model-aware execution exists, or explicitly define, test, and report projection onto the configured target model? I would also document the intended split: mixed source models are valid provenance in the importer, while execution policy is enforced here.

PeaBrane (commented):
Cross-repository parity review against merged Dynamo #13586. Moving raw Weka ownership into AISimulate, retaining source order, and binding output to a stable source snapshot are good directions. The inline notes cover two stale timing-contract regressions, the mixed-model execution policy, v2 ordering compatibility, and resource/provenance follow-ups. Current Dynamo remains pinned to dev2, so there is no live binary mismatch; the companion bump should remove its duplicate Weka parser/DTO adapter and explicitly update runner capabilities. Please also refresh the validation counts and graph digest in the PR description for the current head.

crates/core/src/replay/loadgen/types.rs
```diff
@@ -154,6 +154,11 @@ pub struct AgenticDependency {
 pub struct AgenticMooncakeRow {
     pub request_id: String,
     pub play_id: String,
+    /// Zero-based source corpus order for this play. When present, every play
+    /// must provide one unique contiguous ordinal. This keeps scheduling order
+    /// independent of producer-generated play IDs.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub source_play_ordinal: Option<usize>,
```
PeaBrane, line 163: `source_play_ordinal` is the right fix for the reviewer-demonstrated lane-order bug. The remaining concern is compatibility: it changes scheduling while the interchange still reports v2, so an older v2 reader can accept and ignore the unknown field and fall back to sorting generated `play_id`s. Could we version or capability-signal this ordering contract and add a multi-play materialize/reload compatibility test? I would keep the field; a comment alone cannot repair the old-reader direction.

crates/core/src/replay/loadgen/weka.rs
```diff
@@ -0,0 +1,2194 @@
...
+        corpus_hasher.update(b"aisimulate-weka-corpus-v1\0");
+        let mut block_size = None;
+        let mut plays = 0;
+        let mut requests = 0;
+        let mut raw_zero_outputs = 0;
+        let file_count = files.len();
+        let mut canonical_rows =
+            tempfile::NamedTempFile::new().context("creating Weka row spool")?;
+        for (file_path, relative_path) in files {
+            let snapshot = snapshot_source_file(&file_path, &relative_path, &mut corpus_hasher)?;
```
PeaBrane, line 155: The stable snapshot is the right provenance fix. The recommendation path currently invokes Weka import for every candidate, though, so each trial snapshots, lowers, spools, reparses, and hashes the complete corpus. With the documented 570 MB corpus and default parallelism 16, raw snapshots alone can exceed 9.1 GB before row spools. Could the compiled graph be frozen once per sweep, or at minimum cached once per worker by immutable corpus identity and block-size assertion?

crates/core/src/replay/loadgen/weka.rs
```diff
@@ -0,0 +1,2194 @@
...
+        let (root, files) = collect_source_files(path)?;
+        if files.is_empty() {
+            bail!(
+                "Weka source {} contains no JSON or JSONL files",
+                path.display()
+            );
+        }
+
+        let mut corpus_hasher = blake3::Hasher::new();
+        corpus_hasher.update(b"aisimulate-weka-corpus-v1\0");
```
PeaBrane: Is changing the corpus-hash domain from `dynamo-weka-corpus-v1` to `aisimulate-weka-corpus-v1` intentional? Identical raw bytes now receive a different source digest across an ownership-only migration. If continuity is intended, retaining the original domain seems preferable. If the identity break is intentional, could this become a named algorithm-version constant with a comment that changing it changes persisted provenance identity?

crates/core/src/replay/loadgen/weka.rs
```diff
@@ -0,0 +1,2194 @@
...
+                &mut used_hashes,
+            )?;
+            rows.push(AgenticMooncakeRow {
+                request_id,
+                play_id: play_id.clone(),
+                source_play_ordinal: Some(source_play_ordinal),
+                session_id: stream.session_id.clone(),
+                model: request.request.model.clone(),
+                input_length: Some(request.request.input_length),
+                output_length: Some(request.request.output_length.max(1)),
```
dreamtalen: Wondering if this >=1 OSL guard is intentional? Our replay supports OSL = 0, so it seems we could preserve the authored value here.

crates/core/src/replay/loadgen/types.rs
Harrilee: Thanks, agreed that the ordering contract needs an explicit multi-play materialize/reload test. There is no deployed old-reader/new-writer compatibility boundary here yet, though: Dynamo does not currently consume the AISimulate Weka trace builder, and AISimulate has not published this builder or its materialized output as a supported interchange contract. We are still finalizing v2 before its first rollout.

Introducing v3 now would imply that v2 had already been released as a meaningful public generation requiring migration. It would also create a confusing lifecycle in which the never-public transitional v2 is immediately deprecated while v1 and the proposed v3 remain the long-lived versions.

I therefore propose to finalize the ordering semantics in v2: keep `source_play_ordinal`, make it the normative ordering contract, and add the requested multi-play Weka → materialize → reload test to verify identical graph identity and single-lane dispatch order. Once this PR is merged, Dynamo can be seamlessly adapted to consume the finalized AISimulate v2 builder contract, allowing both repositories to adopt the ordering semantics together without creating a mixed-version compatibility window.

crates/core/src/replay/loadgen/weka.rs
Harrilee: Thanks for flagging the difference from the final Dynamo implementation. I checked this against AIPerf's Weka contract, which this public ingress must continue to accept.

AIPerf intentionally models `api_time` as optional and treats absent/null as a zero-width interval for dependency inference:
- optional schema: https://github.com/ai-dynamo/aiperf/blob/13ae4f6b6b5363007ad52ee2470c3b49c9403b34/src/aiperf/dataset/loader/weka_trace_models.py#L45-L47
- request-end fallback while preserving `None` vs explicit zero: https://github.com/ai-dynamo/aiperf/blob/13ae4f6b6b5363007ad52ee2470c3b49c9403b34/src/aiperf/dataset/loader/weka_trace.py#L114-L128
- the same policy in cross-stream interval inference: https://github.com/ai-dynamo/aiperf/blob/13ae4f6b6b5363007ad52ee2470c3b49c9403b34/src/aiperf/timing/replay_dependencies.py#L57-L83
- explicit regression coverage for `0.0` and `None`: https://github.com/ai-dynamo/aiperf/blob/13ae4f6b6b5363007ad52ee2470c3b49c9403b34/tests/unit/dataset/loader/test_weka_agent_chains_adv_phase1.py#L104-L118

Requiring presence here would make AISimulate reject AIPerf-compatible Weka inputs and break the shared Weka ingestion contract used by downstream integrations. In 5cc9ace8 I therefore kept absent/null accepted as the graph-shaping zero-width fallback, added separate absent/null/explicit-zero tests, and documented the invariant. The provenance remains distinct: absent/null stays `recorded_api_time_ms=None`, while an authored zero is `Some(0.0)` and participates distinctly in graph identity. AISimulate continues to reject negative and non-finite durations.

crates/core/src/replay/loadgen/weka.rs
Harrilee: Agreed. 5cc9ace8 removes the relative-time guess and now rejects any nested request whose timestamp precedes its subagent marker, with the published-Weka absolute trace-relative invariant stated in the error and covered by a regression test.

I also opened AIPerf PR https://github.com/ai-dynamo/aiperf/pull/1396 to remove the same legacy heuristic there. That PR:
- references this AISimulate discussion and the Dynamo resolution;
- adds fail-closed coverage for an inner request before its marker;
- validates the absolute-timestamp contract against AIPerf's checked-in real SemiAnalysis AgentX trace (marker `33.161`, nested requests `33.161` and `33.22`);
- migrates old synthetic fixtures from relative to absolute timestamps.

No corpus download was needed.

crates/core/src/replay/loadgen/weka.rs
Harrilee: Yes, the identity break is intentional. AISimulate now owns the public Weka ingestion/provenance contract, and the follow-up Dynamo adapter will call this implementation rather than retain a second parser/lowering copy, so the persisted identity is intentionally in the AISimulate namespace.

I kept `aisimulate-weka-corpus-v1\0`, but 5cc9ace8 promotes it to the named `WEKA_CORPUS_DIGEST_DOMAIN_V1` constant with an explicit comment that changing the bytes changes every persisted Weka source digest and graph identity and must be treated as a versioned compatibility break.

crates/core/src/replay/loadgen/weka.rs
Harrilee: Agreed. 5cc9ace8 removes the `.max(1)` rewrite and preserves the authored zero in canonical Agentic Mooncake rows and `ValidatedAgenticGraph`.

AISimulate's native runtime supports a zero-output request as a prefill-only/KV-cache-warmup operation: it runs prompt prefill and produces cache state without entering autoregressive decode. AIPerf's `0 -> 1` behavior exists because its real OpenAI-compatible HTTP endpoint path rejects `max_tokens: 0`; that transport limitation should not alter AISimulate's canonical IR.

The JSON and JSONL E2E fixtures now include OSL=0, and the native E2E asserts that the request completes with requested/output length 0 and no first-token event.

crates/core/src/replay/loadgen/weka.rs
Harrilee: DCO-only follow-up: the final head SHA is 5cc9ace8 (the implementation is unchanged from fe791281).

Harrilee: Agreed. fe791281 implements the minimum per-worker option you suggested: a process-local compiled graph cache keyed by the immutable corpus digest.

The first candidate snapshots, validates, lowers, spools, and builds the graph. Subsequent candidates re-hash the current source bytes and reuse the compiled `AgenticTrace`, avoiding repeated parse/lower/spool/build work. The source path is only a lookup hint: an edit or replacement changes the digest and forces a miss, and the source block-size assertion is checked on every call. Tests cover cache hits, content-change invalidation, graph-identity changes, and block-size mismatch on the cached path.

This still means each sweep worker compiles once on its first candidate; it removes repeated work within that worker without introducing cross-process mutable graph state..

crates/core/src/python.rs
Harrilee: Agreed — I implemented the explicit projection option in e7f04a04.

The boundary is now defined as follows:
- the neutral importer and `ValidatedAgenticGraph` continue to accept mixed source models;
- `AgenticGraphIdentity` retains the sorted source-model set as provenance;
- the Python runner resolves the configured deployment model independently of the timing implementation and passes it as `execution_model`;
- the native agentic boundary fails closed if no configured target model is available before handing the graph to the model-neutral `WorkloadDriver`;
- every agentic request is explicitly projected onto that one target model for timing in v1.

The native report now includes:
```json
"agentic_model_projection": {
  "policy": "project_to_configured_target",
  "source_models": ["model", "other-model"],
  "target_model": "example/model"
}
```

I changed the checked-in Weka E2E corpus to contain two source models, and the public engine-stack test verifies both successful mixed-model execution and the exact projection audit above. A separate runner test rejects agentic execution when no target model can be resolved, and the Rust importer test verifies deterministic source-model provenance. The CLI design doc now documents the importer/execution split and states that per-node heterogeneous timing is not supported yet.

Validation: 93 Rust loadgen tests and 61 Python runner/runtime tests passed.

crates/core/src/replay/loadgen/weka.rs
Harrilee: Thanks again for insisting that we must not infer or rebase timestamps per request. New upstream evidence changes only the absolute-only assumption: Anthony verified that the raw kv-cache-tester generator defines nested `t` as marker-relative, while the registered SemiAnalysis republications are root-absolute ([evidence and generator excerpt](https://github.com/ai-dynamo/aiperf/pull/1396#discussion_r3963470880)).

Commit a2ca06d6 therefore recognizes both producer conventions without restoring the old per-request heuristic:
- `traffic.source.nested_timestamp_basis` accepts `auto`, `absolute`, or `relative` for the single public `weka` format;
- `auto` preflights every deterministic JSON/JSONL row and resolves one basis for the entire corpus from decisive first-inner anchors; mixed evidence or a non-equivalent inconclusive corpus fails closed with source/trace/agent context;
- explicit `absolute` rejects a child materially before its marker and clamps only sub-epsilon producer rounding to the marker; explicit `relative` performs the marker offset;
- both paths lower to root-absolute canonical IR before graph construction;
- the resolved basis is included in source identity and the native report, and the default CLI prints requested/resolved basis on stderr.

Tests cover raw kv-cache-tester-style relative input, published-style absolute input, mixed and ambiguous corpora, exact-zero equivalence, epsilon behavior, semantic-digest separation, direct/materialized-v2 parity, and the user-visible CLI audit. This preserves your core invariant—one corpus-level interpretation, never request-local guessing—while keeping both actual Weka producers loadable.

PeaBrane (approved):
✅

crates/core/src/replay/loadgen/weka.rs
Harrilee: Follow-up with the simpler corpus rule approved in the AIPerf review ([review](https://github.com/ai-dynamo/aiperf/pull/1396#pullrequestreview-5157785465)): commit 5dbe9221 replaces the first-child anchor voting described above.

`auto` now scans every nested request in every deterministic JSON/JSONL row. If any `child.t < marker.t - epsilon`, it resolves the entire corpus as `relative`; otherwise it resolves it as `absolute`. The importer completes validation of all rows even after finding a relative witness, then applies exactly one resolved basis uniformly during lowering. Explicit `absolute` and `relative` remain authoritative; the same literal pre-marker predicate is shared by auto detection and the explicit-absolute guard, with within-epsilon values clamped to the marker.

The CLI INFO message now calls auto resolution a heuristic, and the documentation states its limits: this rule cannot certify a homogeneous corpus, a malformed absolute child can select relative globally, and relative offsets all at or above their markers can select absolute. Tests cover delayed absolute children, a later corpus row overriding a coincidental absolute-looking anchor, evidence after the first child, exact/near-zero markers, the epsilon boundary, continued validation after a witness, uniform lowering, and direct/materialized parity. The published two-row golden remains unchanged and resolves `absolute`.

docs/cli/design.md
```diff
@@ -489,13 +491,44 @@ first-arrival pacing, and inter-turn or dependency delays remain unscaled.
...
+Dynamo is an optional integration and is not required to parse, convert, or predict a Weka corpus.
+Two producer timestamp conventions exist: raw kv-cache-tester nested request timestamps are relative
+to their subagent marker, while SemiAnalysis-published AgentX timestamps are root-trace absolute.
+`nested_timestamp_basis` may select either convention explicitly. When omitted (or set to `auto`),
+AISimulate scans every nested request in every JSON/JSONL row before lowering. If any child timestamp
+is earlier than its subagent marker by more than the join epsilon, the complete corpus is interpreted
+as relative; otherwise it is interpreted as absolute. This is one corpus-wide heuristic, never a
+per-request rewrite. It cannot prove that a corpus is homogeneous: a malformed absolute request can
+select relative for the entire corpus, while relative offsets that are all at or above their markers
+can select absolute. Producers with ambiguous data should set the basis explicitly. Both conventions
```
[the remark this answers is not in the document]
Harrilee: This is the deliberately approved auto contract, not an unhandled fallback. Rudy requested this exact corpus-wide ordering heuristic in https://github.com/ai-dynamo/aiperf/pull/1396#pullrequestreview-5157785465, and the implementation/limitations are recorded in https://github.com/ai-dynamo/aisimulate/pull/115#discussion_r3972983549. Without producer metadata, numeric timestamps cannot reliably distinguish mixed conventions from delayed relative offsets; restoring anchor voting would reintroduce the false rejections discussed there. Explicit absolute/relative modes remain authoritative, full-input numeric/structural validation still runs, and default CLI INFO labels auto as heuristic. Resolving this as an intentional reviewed contract; no change to that agreed rule is needed.

crates/core/src/python.rs
```diff
@@ -628,14 +662,37 @@ fn build_runtime_input(
...
+        ensure!(
+            traffic.agentic_lanes != Some(0),
+            "agentic_lanes must be greater than 0"
+        );
+        if traffic.agentic_lanes.is_some() {
+            ensure!(
+                matches!(format, "weka" | "agentic_mooncake" | "dynamo"),
+                "agentic_lanes requires weka, agentic_mooncake, or agentic Dynamo input"
+            );
+        }
```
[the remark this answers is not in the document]
Harrilee: Fixed in d71203e489541bc98c5f4a616dcec3b225b8fb72. The native source-level guard now rejects any supplied agentic_lanes before entering the synthetic/trace branches; existing positive-value and supported-trace-format checks remain intact.

Verified against a freshly rebuilt native extension: synthetic and synthetic-session each reject lanes=0 and lanes=1, while omitting lanes still runs successfully. Also checked that Weka lanes=0, ordinary Mooncake with lanes, and standard Dynamo with lanes are rejected. These were local ad hoc checks; no new test files were added.

python/aisimulate/src/aisimulate/runner.py
```diff
@@ -318,12 +344,36 @@ def _materialize_engine_execution_spec(
         traffic = {
             key: value for key, value in spec.workload.items() if key in _RUNTIME_TRAFFIC_FIELDS and value is not None
         }
-        if traffic.get("trace_format") != "dynamo":
+        if traffic.get("trace_format") not in {"dynamo", "weka"}:
             traffic.setdefault("trace_block_size", trace_block_size)
+        if traffic.get("trace_format") in {"agentic_mooncake", "dynamo", "weka"}:
+            if execution_model is None:
+                raise ValueError("agentic execution requires a configured target model")
+            traffic["execution_model"] = execution_model
```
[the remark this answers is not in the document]
Harrilee: Fixed in d71203e489541bc98c5f4a616dcec3b225b8fb72. Python now requires the target model only for known agentic inputs (Weka/v2, or Dynamo with explicit lanes), so ordinary disaggregated Dynamo is no longer rejected.

One important distinction from the suggested patch: we still forward an available execution_model for aggregated Dynamo even when lanes are omitted. Such a file may contain an agentic graph, and the native loader validates the actual standard/agentic kind.

Verified with the rebuilt native runtime: standard Dynamo disaggregated completed 2 requests without execution_model; agentic Dynamo aggregated with omitted lanes still received the model and completed 4 requests. Weka/v2 disaggregated remains rejected. The existing engine CLI matrix also passes.

Harrilee:
Addressed the outside-diff recommendation finding from [this review](https://github.com/ai-dynamo/aisimulate/pull/115#pullrequestreview-5152765463) in d71203e489541bc98c5f4a616dcec3b225b8fb72.

- `_recommendation_workload` now forwards optional `agentic_lanes`; the downstream strict `Workload` schema also accepts and validates it, so forwarding does not merely turn a dropped field into a schema error.
- Fixed the related downstream schema gap for explicit `weka_nested_timestamp_basis`. `auto`, `absolute`, and `relative` now survive recommendation compilation, candidate workload serialization, native execution, and exported prediction configuration. Invalid values and non-Weka uses remain rejected; omission stays optional.
- Preserved the aggregated-Dynamo model path while restoring ordinary disaggregated Dynamo replay, and rejected lanes on synthetic inputs at the native boundary (details in the corresponding inline replies).

Validation on these changes: 215 existing tests passed (182 runner/traffic/CLI/sweeper tests plus 33 runtime/engine CLI tests). Local ad hoc checks additionally ran all three explicit timestamp modes through real CLI recommend -> exported predict, with 3 requests completed per mode; native report checks confirmed the expected resolved basis and lanes=1. Eight invalid timestamp-basis/source combinations were rejected. Native lane guards were checked against a freshly rebuilt extension.

No test files were added or modified in this pass. The separate requests for stronger checked-in ordering/CLI/model-fallback coverage and revision-pinned sample fetching remain open; this update does not claim to have implemented them.

crates/core/src/replay/loadgen/types.rs
Harrilee: Maintainer scope decision: old v2 readers consuming the finalized new output are not a supported compatibility target for this PR. We are keeping source_play_ordinal as the ordering contract in v2 and are not adding v3 or an old-reader compatibility layer. The additional multi-play reload regression coverage is accepted as non-blocking for this merge, rather than claimed implemented. Resolving on that scope decision.

crates/core/src/replay/loadgen/tests.rs
```diff
@@ -558,6 +558,57 @@ fn agentic_v2_rejects_invalid_schema_and_graph_contracts() {
     }
 }
 
+#[test]
+fn agentic_v2_rejects_ambiguous_source_play_order() {
```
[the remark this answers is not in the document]
Harrilee: Agreed that inverse-ID, multi-play materialize/reload coverage would be stronger than the existing tests. The maintainer has explicitly accepted this coverage improvement as non-blocking and requested no additional tests in this pass. This is deferred, not claimed implemented; current source-ordinal ordering remains unchanged. Resolving on that decision.

scripts/qualify_weka_samples.py
```diff
@@ -0,0 +1,132 @@
...
+        text=True,
+    )
+    revision = json.loads(metadata.stdout).get("sha")
+    if revision != DATASET_REVISION:
+        raise RuntimeError(f"dataset revision changed: expected {DATASET_REVISION}, found {revision}")
+
+    url = (
+        "https://datasets-server.huggingface.co/rows"
+        f"?dataset={DATASET.replace('/', '%2F')}&config=default&split=train&offset=0&length=2"
+    )
```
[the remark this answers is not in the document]
Harrilee: The maintainer has accepted this qualification-tool reproducibility hardening as non-blocking for this merge. The current script still checks revision, expected sample IDs, source digest and graph digest, but the rows request is not atomically revision-pinned; we are not claiming otherwise. A future change must verify that the selected API actually honors immutable revisions and continue fetching only two rows. Resolving as accepted/deferred, not as a code fix.

tests/e2e/test_unified_cli_engine.py
```diff
@@ -103,10 +109,15 @@ def test_engine_predict_cli_cases(config_path: Path, tmp_path: Path) -> None:
     saved_summary = report.get("summary", report)
     assert summary["completed_requests"] > 0
     assert saved_summary["completed_requests"] == summary["completed_requests"]
+    if config_path.name.startswith(("11-", "12-")):
+        assert "validated the complete Weka corpus" in result.stderr
+        assert "requested='auto', resolved='absolute'" in result.stderr
```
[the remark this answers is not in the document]
Harrilee: Agreed this would strengthen the checked-in CLI regression assertions. Existing engine-stack coverage verifies per-play timing, and the recent local CLI/native checks verified lane plumbing, but these do not replace the requested persisted CLI timing assertions. The maintainer has accepted the additional test coverage as non-blocking and requested no new tests in this pass. Resolving as deferred, not claimed implemented.

tests/test_runner.py
```diff
@@ -127,6 +127,157 @@ def test_factory_is_pickleable_and_advertises_engine_only_capabilities():
...
+            _spec(
+                deployment=deployment,
+                workload={
+                    "source_type": "trace",
+                    "load_type": "trace_timestamps",
+                    "trace_path": "weka-corpus",
+                    "trace_format": "weka",
+                },
+            )
+        )
```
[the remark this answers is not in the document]
Harrilee: The standard-Dynamo disaggregated defect is fixed in d71203e4 and was verified against the rebuilt native runtime (2 requests completed). The metadata-first model resolver remains in place. The additional metadata-only and dedicated regression test coverage requested here has been accepted by the maintainer as non-blocking; no new tests were added in this pass. Resolving on that coverage-scope decision.

python/aisimulate/src/aisimulate/sweeper/config.py
```diff
@@ -290,6 +292,23 @@ def synthetic_arrival_interval_ms(self) -> float | None:
...
     def _validate_workload(self) -> Workload:
+        if self.weka_nested_timestamp_basis is not None and (
+            self.trace_path is None or self.source_type != "trace" or self.trace_format != "weka"
+        ):
+            raise ValueError("weka_nested_timestamp_basis requires Weka trace input")
+        if self.agentic_lanes is not None:
+            if (
+                self.trace_path is None
+                or self.source_type != "trace"
+                or self.trace_format not in {"weka", "agentic_mooncake", "dynamo"}
```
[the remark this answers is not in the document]
Harrilee: This guard is intentional after following the real consumer. In runner.py, source_type=None selects the legacy _materialize_requests path, whose materialize_configured_traffic payload does not carry agentic_lanes or weka_nested_timestamp_basis. Allowing the new controls through that path would not implement their semantics. The public recommendation compiler already emits source_type='trace' and load_type='trace_timestamps'; all three explicit Weka basis modes and lanes=1 were verified through recommendation, native execution, and exported prediction. Before this change, these fields were rejected as extras by Workload, so this does not remove a previously supported combination. Implicitly normalizing legacy workloads would be a separate API behavior change. Keeping the explicit native workload-driver boundary and resolving this as intentional fail-closed validation.

PR #120, opened 2026-09-06
fix(engine): align the SGLang scheduler with SGLang: prefill-first passes, retraction handling, chunked-prefill prefix reuse

tianhaox:
## Summary

Three SGLang-fidelity fixes for the scheduler emulation, found by comparing against HiSim (the real SGLang scheduler running on CPU).

### 1. Prefill-first pass selection

The SGLang scheduler emulation ran every engine pass as "admit a prefill batch, then one decode step for **every** running request" — the pass shape inherited from the Dynamo mocker (#12525), which matches vLLM's single-step scheduling but not SGLang's. Real SGLang (`Scheduler.get_next_batch_to_run`, unchanged from v0.5.6.post2 to current `main`) is prefill-first: whenever a prefill batch can be formed it runs alone and the running decode batch waits; prefill and decode only share a forward with `--enable-mixed-chunk`.

This makes the SGLang scheduler prefill-first (no switch: SGLang has none either):

- In a pass that formed a prefill batch, requests that were already running are set aside and do not decode; only the freshly prefilled requests go through the decode bookkeeping step, which is not charged any time because the first token is produced by the prefill forward itself (`simulate_decode_step_with_sampler(.., charge_time)`).
- FPM snapshot of such a pass reports a pure prefill forward (`num_decode_requests = 0`).
- Tests: the two FPM tests that encoded the old assumption ("after prefill the request joins the decode step of the same pass") are updated; a new test asserts running requests emit no token during a prefill pass, keep FIFO order, and resume on the next pass.

No Python-side changes and no new configuration.

### 2. Retraction re-queue and `new_token_ratio` under KV pressure

Aligns with `Scheduler.update_running_batch` / `ScheduleBatch.retract_decode` (v0.5.6.post2, unchanged on `main`):

- **Retracted requests re-queue at the back.** SGLang re-adds them through `_add_request_to_queue` (`waiting_queue.append`); the mocker pushed them to the front, ahead of requests that had never run.
- **`new_token_ratio` is re-estimated after a retraction**, `min(1, (decoded + 20 * n) / (max_new + 1))` over the remaining batch (`SGLANG_RETRACT_DECODE_STEPS = 20`), instead of jumping back to the initial 0.7, which is far more conservative than SGLang's post-retraction estimate and made the mocker admit fewer requests after every retraction.
- **No ratio reset when prefill admission runs out of KV** (SGLang only marks the batch full on `AddReqResult.NO_TOKEN`), and **the ratio decays only on decode passes** (`update_running_batch`).

Test: `test_retraction_requeues_at_back_and_reestimates_new_token_ratio` drives a 56-token KV pool into retraction and asserts the queue order and the re-estimated ratio.

### 3. Prefill admission matches the whole prompt once (SGLang `Req.init_next_round_input(tree_cache)`)

SGLang matches a waiting request's whole prompt against the radix tree once at admission, capped at `input_len - 1`, then chunks the remaining `extend_input_len`; chunked continuations are re-initialised **without** the tree cache and never re-match (`scheduler.py` `chunked_req.init_next_round_input()` vs `req.init_next_round_input(self.tree_cache)` for waiting requests), and `add_one_req` / `add_chunked_req` charge only the computed tokens to the chunk and input budgets. The mocker scoped the first-admission match to the first chunk (`alloc_tokens = req.sequence_prefix(chunk_end)`), so a shared prefix longer than one chunk was recomputed chunk by chunk, and it charged the cached prefix against the budgets. On the Mooncake conversation trace (46% of requests span several 8192-token chunks) this cost 7–13% extra prefill tokens.

Now: `prefill.rs` runs the read-only `prefix_match_hashes_len` over the whole prompt (the request already carries its page hashes), sizes the chunk from the matched prefix, charges only the uncached tokens, and locks the same path in `allocate_for_request_lease` (clamped to the SGLang cap so a fully cached prompt still computes its last token). `extend_allocation` keeps extending the request's own pages only; it returns the owned prefix length and does the eviction for new pages after the path is locked (the scheduler-side pre-eviction is gone). Zero-output prefix-only replay requests keep the full match since they have no first token to sample. vLLM is untouched (own scheduler and KV backend; vLLM v1 looks up the prefix cache once per request).

Tests: `test_first_admission_matches_whole_prompt_once` (shared 10-token prefix → one 4-token prefill pass, nothing chunked), `test_fully_cached_prompt_still_computes_its_last_token`, `test_small_request_is_co_admitted_behind_mostly_cached_long_request` (only the uncached tokens count against the chunk budget), `chunked_continuation_extends_own_pages_without_rematching` (KV manager).

## Why it matters — comparison against HiSim (real SGLang scheduler)

[HiSim](https://github.com/alibaba/tair-kvcache/tree/main/hisim) runs the real SGLang 0.5.6.post2 Python scheduler on CPU with the model forward replaced by an aiconfigurator latency predictor, so it is a scheduler-level reference for AISimulate. Setup used for the numbers below (all inputs aligned so that only scheduling differs):

- Qwen/Qwen3-32B-FP8, `h200_sxm`, SGLang perf data `0.5.14`, bf16 KV cache + bf16 FMHA on both sides (AIS defaults to fp8 KV for `*-FP8` checkpoints, which differs from SGLang's `--kv-cache-dtype auto`; pinned via engine args for this experiment), KV pool pinned to HiSim's estimate (371,587 tokens at TP1), `chunked_prefill_size 8192`, `max_prefill_tokens 16384`, FCFS, no overlap, single replica.
- Closed loop, concurrency 128, 1280 requests. HiSim is driven with all requests created at t=0 and `max_running_requests=128`; its TTFT is reconstructed for the closed-loop client (k-th admission arrives at the (k−128)-th completion).
- The op-level latency estimators agree to <1% for identical batch shapes, so the residual gap is scheduling.

| scenario | metric | AIS before (mixed pass) | AIS after (prefill-first) | HiSim (real SGLang) |
|---|---|---:|---:|---:|
| fixed 1024 in / 1024 out, TP1 | TTFT mean (ms) | 961 | 4,489 | 4,472 |
| | TPOT mean (ms) | 33.6 | 30.2 | 30.2 |
| | ITL p99 (ms) | 552 | 30 | 30 |
| | output tok/s | 3,704 | 3,708 | 3,708 |
| fixed 1024 / 1024, TP2 | TTFT mean (ms) | 511 | 2,378 | 2,371 |
| | TPOT mean (ms) | 19.1 | 17.3 | 17.3 |
| | ITL p99 (ms) | 294 | 17 | 17 |
| fixed 1024 / 1024, TP4 | TTFT mean (ms) | 325 | 1,506 | 1,500 |
| | TPOT mean (ms) | 12.8 | 11.6 | 11.6 |
| | ITL p99 (ms) | 187 | 11 | 11 |
| random in/out U[512,1536], TP1 | TTFT mean (ms) | 572 | 533 | 566 |
| | TPOT mean (ms) | 34.5 | 34.5 | 34.5 |
| | ITL p99 (ms) | 116 | 116 | 116 |
| | output tok/s | 3,518 | 3,522 | 3,519 |
| KV-bound, random in U[2000,6000] / out U[500,1500], TP1 | TTFT mean (s) | 34.8 | 34.6 | 34.1 |
| | TPOT mean (ms) | 57.6 | 57.7 | 58.6 |
| | ITL p99 (ms) | 445 | 496 | 483 |
| | output tok/s | 1,345 | 1,347 | 1,341 |

Throughput and E2E were already within 0.3% (the pass shape does not change the total work); the change fixes *where* the prefill stall lands. With the mixed pass, every prefill batch stalled all running decodes for one prefill duration and staggered same-length cohorts, so TTFT was underestimated by up to 4.6× and per-token ITL p99 was overestimated by 18× on synchronized bursts, while ITL tails on realistic random-length loads were already coincidentally close. After the change all five scenarios match the real scheduler within ~1% on TTFT mean, TPOT, ITL p99 and throughput (TTFT p50 within 6%, which is the closed-loop reconstruction noise on the HiSim side).

### KV-bound scenarios (fix 2)

Same setup, 371,587-token KV pool, concurrency 128:

| scenario | metric | prefill-first only | prefill-first + retraction fix | HiSim (real SGLang scheduler) |
|---|---|---:|---:|---:|
| fixed 4000 in / 1000 out | requests retracted at least once | 89 | 56 | 64 |
| | TTFT mean / p50 (s) | 45.2 / 62.0 | 42.0 / 44.2 | 25.1 / 16.2 ¹ |
| | TPOT mean (ms) | 49.3 | 51.7 | 70.0 (47.1 excluding retraction re-queue gaps) ¹ |
| | output tok/s | 1,308 | 1,331 | 1,313 |
| random in U[2000,6000] / out U[500,1500] | requests retracted | 3 | 2 | 2 |
| | TTFT mean (s) | 34.6 | 34.6 | 34.1 |
| | TPOT mean (ms) | 57.7 | 57.8 | 58.6 |
| | output tok/s | 1,347 | 1,347 | 1,341 |

¹ The HiSim reference is not trustworthy for the equal-length KV-bound case: HiSim (OFFLINE mode) had to be driven with all 1280 requests queued at t=0 plus `max_running_requests=128`, so a retracted request re-queues behind ~1,200 never-run requests and waits 38–880 s (median 456 s) before re-prefill. That inflates HiSim's TPOT (70 → 47 ms once those gaps are excluded, matching AIS's 52 ms) and keeps ~16 retracted requests "in flight", which shortens the queue seen by fresh arrivals and hence its TTFT. A real 128-concurrency client would only have ~50 requests queued behind a retraction. Retraction counts, throughput and the random-length scenario are unaffected by this artefact and agree.

### Mooncake trace (fix 3)

Public `kvcache-ai/Mooncake` FAST25 `conversation_trace.jsonl`, first 600 s (1629 requests, mean isl 9.9k / osl 349, 2.7 rps; 121 requests above the 40k context dropped), timestamp replay on both sides, prefix caching on, same aligned setup (bf16 KV, sglang 0.5.14 data, KV pool pinned to HiSim's estimate). HiSim materializes `hash_ids` into deterministic 512-token blocks so the radix cache sees the same sharing structure.

| metric | TP4: before / after / HiSim | TP8: before / after / HiSim |
|---|---|---|
| output tok/s | 930 / 935 / 934 | 944 / 944 / 944 |
| TTFT mean (ms) | 1,823 / 1,604 / 1,432 | 940 / 843 / 739 |
| TTFT p99 (ms) | 5,724 / 5,073 / 4,936 | 3,361 / 3,104 / 3,032 |
| TPOT mean (ms) | 151 / 96 / 94 | 30.2 / 23.4 / 23.4 |
| ITL p99 (ms) | 1,835 / 1,159 / 1,151 | 9.7 / 8.4 / 8.7 |
| E2E mean (ms) | 30,408 / 20,228 / 20,868 | 5,407 / 4,635 / 4,678 |
| first-admission prefix reuse | 0.08 / 0.13 / 0.14 | 0.11 / 0.21 / 0.21 |

"before" = fixes 1+2 only. TPOT, ITL p99, E2E and the prefix-hit ratio now agree within 3%. TTFT mean is still +12–14%; the budget over-charge is fixed and did not move it, so it has another cause. Next candidate, not in this PR: when a request's `extend_input` fits `chunked_prefill_size` but exceeds the pass's remaining chunk budget, the mocker rejects it and leaves the budget unused (`charged_input_tokens > rem_chunk_tokens`), whereas SGLang's `add_one_req` takes a partial chunk to fill the pass.

### Provenance

The scheduler modules are behavioural re-implementations of SGLang's scheduling semantics as of v0.5.6.post2 (`5c8bd8b5`; `python/sglang/srt/managers/scheduler.py`, `schedule_batch.py`, `schedule_policy.py`). No SGLang source is copied; the module docs now name the revision and files for traceability. The pre-existing SGLang and vLLM scheduler emulations in this tree carry the same NVIDIA-only headers, so whether behavioural ports need a `THIRD_PARTY_NOTICES.md` entry is a maintainer/OSRB call for the module as a whole rather than for this PR.

## Test plan

- `cargo test --workspace` (1149 pass; the two FPM tests and the MTP burst test updated as described, eight tests added)
- `cargo fmt --all -- --check`, `cargo clippy -p aisimulate-core --all-targets` (no new warnings)
- `cargo test --manifest-path crates/tests/public-api/Cargo.toml`


crates/core/src/engine/kv_manager/sglang_backend.rs
```diff
@@ -282,28 +282,85 @@ impl SglangKvManager {
     /// This is used by chunked-prefill continuation where the request still
     /// owns token slots for a prefix that may extend past the radix-tree's
     /// page-aligned cached prefix.
+    ///
+    /// SGLang re-runs `match_prefix` over the whole `fill_ids` for every chunk
```
Arsene12358: This is the inverted claim. `get_new_batch_prefill` calls `self.chunked_req.init_next_round_input()` with no tree cache ([v0.5.6.post2 scheduler.py:1782-1784](https://github.com/sgl-project/sglang/blob/v0.5.6.post2/python/sglang/srt/managers/scheduler.py#L1782-L1784); same on `main` at c99d906:3771), and `init_next_round_input` only matches when one is passed ([schedule_batch.py:812](https://github.com/sgl-project/sglang/blob/v0.5.6.post2/python/sglang/srt/managers/schedule_batch.py#L812)). Between chunks `cache_unfinished_req` pins `prefix_indices` to the request's own tokens ([radix_cache.py:503-533](https://github.com/sgl-project/sglang/blob/v0.5.6.post2/python/sglang/srt/mem_cache/radix_cache.py#L503-L533)).

If the re-match stays as an approximation, the honest wording is: "SGLang matches the whole prompt once at first admission (`Req.init_next_round_input(tree_cache)`) and never re-matches a chunked continuation; this mocker only matched within the first chunk, so re-matching on continuation approximates the one-time whole-prompt match." Preferably drop the re-match and match once at first admission (sketch in the review body).

crates/core/src/engine/scheduler/sglang/prefill.rs
```diff
@@ -116,9 +108,9 @@ pub(super) fn get_new_batch_prefill(
                     req.uuid, req.materialized_tokens
                 );
             }
-            kv_manager
-                .extend_allocation(alloc_tokens, &mut lease)
-                .then_some(req.materialized_tokens)
+            // Chunked-prefill continuation re-matches the radix tree over the whole target
```
Arsene12358: Same attribution issue, and line 102 just above is where the mocker actually diverges from SGLang: `alloc_tokens = req.sequence_prefix(chunk_end)` scopes the first-admission match to the first chunk, so `allocate_for_request_lease` never sees tokens beyond it. SGLang matches `fill_ids[:input_len-1]` once before sizing the chunk and charges only `extend_input_len` to the chunk and input budgets ([add_one_req](https://github.com/sgl-project/sglang/blob/v0.5.6.post2/python/sglang/srt/managers/schedule_policy.py#L581-L611), [add_chunked_req](https://github.com/sgl-project/sglang/blob/v0.5.6.post2/python/sglang/srt/managers/schedule_policy.py#L454-L471)). The `charged_input_tokens` over-charge at line 83 that the description leaves open is the same root cause. Sketch of the one-time match in the review body.

crates/core/src/engine/scheduler/sglang/tests.rs
```diff
@@ -1835,6 +1907,168 @@ mod forward_pass_metrics {
...
+            retracted_seen,
+            "expected a decode retraction under KV pressure"
+        );
+    }
+
+    #[test]
+    fn test_chunked_continuation_reuses_prefix_cached_by_another_request() {
+        // r1 (12 tokens) finishes and is cached. r2 shares r1's first 10 tokens and is prefilled
+        // in 6-token chunks: the second chunk must reuse tokens 6..10 from the cache and compute
+        // only tokens 10..12, as SGLang does by re-matching `fill_ids` on every chunk.
```
Arsene12358: SGLang would match the shared 10 tokens once and run one non-chunked 4-token prefill (`extend_input_len` 4 <= chunk 6); it does not produce this 0 / 2 / 2 three-pass sequence, and pass 1 is a zero-duration pass SGLang cannot have because the match is capped at `input_len - 1` ([schedule_batch.py:806](https://github.com/sgl-project/sglang/blob/v0.5.6.post2/python/sglang/srt/managers/schedule_batch.py#L806)). With the one-time match this test becomes a single 4-token prefill pass, and a companion test that a small request queued behind a mostly cached long one is co-admitted in the same batch would pin the behaviour that matters for TTFT.

Arsene12358 (changes requested):
**Verdict: requesting changes on fix 3 (chunked-prefill re-match). Fixes 1 and 2 check out.**

**What I verified.** Fix 1 (prefill-first passes) and fix 2 (retracted requests re-queue at the back, `new_token_ratio = min(1, (decoded + 20·n) / (max_new + 1))`, decay only on decode passes, no reset on prefill OOM) match SGLang v0.5.6.post2 line for line: the [prefill-first branch of `get_next_batch_to_run`](https://github.com/sgl-project/sglang/blob/v0.5.6.post2/python/sglang/srt/managers/scheduler.py#L1702-L1711), [`update_running_batch` retraction → `_add_request_to_queue` and the decay](https://github.com/sgl-project/sglang/blob/v0.5.6.post2/python/sglang/srt/managers/scheduler.py#L1926-L1953) → [`waiting_queue.append`](https://github.com/sgl-project/sglang/blob/v0.5.6.post2/python/sglang/srt/managers/scheduler.py#L1497-L1505), the [`retract_decode` estimate including the `+ 1`](https://github.com/sgl-project/sglang/blob/v0.5.6.post2/python/sglang/srt/managers/schedule_batch.py#L1665-L1677), and [`NO_TOKEN` only setting `batch_is_full`](https://github.com/sgl-project/sglang/blob/v0.5.6.post2/python/sglang/srt/managers/scheduler.py#L1828-L1837). Same on `main` at c99d906. The new tests cover each claim. Since CI has not run on this PR yet, locally at dff3395e: `cargo fmt --all -- --check` clean, `cargo test --workspace` 1200 passed / 0 failed, public-api crate 4/4, no new clippy warnings in the changed files. No regression versus `main` found.

**Blocking: the SGLang behaviour fix 3 cites does not exist.** The `extend_allocation` docstring, the continuation comment in `prefill.rs`, the new test and the PR body say SGLang re-runs `match_prefix` over the whole `fill_ids` for every chunk via `Req.init_next_round_input`. In SGLang the chunked request is re-initialised **without** the tree cache ([scheduler.py:1782-1784](https://github.com/sgl-project/sglang/blob/v0.5.6.post2/python/sglang/srt/managers/scheduler.py#L1782-L1784)), and `init_next_round_input` only matches when a tree cache is passed ([schedule_batch.py:798-812](https://github.com/sgl-project/sglang/blob/v0.5.6.post2/python/sglang/srt/managers/schedule_batch.py#L798-L812)); only waiting requests get one ([scheduler.py:1821](https://github.com/sgl-project/sglang/blob/v0.5.6.post2/python/sglang/srt/managers/scheduler.py#L1821)). Between chunks, `cache_unfinished_req` inserts the request's own tokens and re-matches exactly that key, asserting `len(new_indices) == len(keys)`, so `prefix_indices` stays the request's own range ([radix_cache.py:503-533](https://github.com/sgl-project/sglang/blob/v0.5.6.post2/python/sglang/srt/mem_cache/radix_cache.py#L503-L533)). `add_chunked_req` then charges only `extend_input_len` ([schedule_policy.py:454-471](https://github.com/sgl-project/sglang/blob/v0.5.6.post2/python/sglang/srt/managers/schedule_policy.py#L454-L471)). Unchanged on `main` ([scheduler.py:3771](https://github.com/sgl-project/sglang/blob/c99d906effa8bd05573995127f0d4a0984c5a96a/python/sglang/srt/managers/scheduler.py#L3771) vs [:3834](https://github.com/sgl-project/sglang/blob/c99d906effa8bd05573995127f0d4a0984c5a96a/python/sglang/srt/managers/scheduler.py#L3834)). SGLang's rule is therefore: match the whole sequence **once** at first admission, capped at `input_len - 1` ([schedule_batch.py:806](https://github.com/sgl-project/sglang/blob/v0.5.6.post2/python/sglang/srt/managers/schedule_batch.py#L806)), then chunk the remainder and never re-match.

**The actual divergence is one step earlier and pre-existing.** `prefill.rs:102` scopes the first-admission match to the first chunk (`alloc_tokens = req.sequence_prefix(chunk_end)`, and `allocate_for_request_lease` matches hashes only through `token_ids.len()`). Re-matching on every continuation compensates the compute total but not the pass structure. Probe (temporary test, code at the end; polynomial perf model, so durations are illustrative): block 64, `chunked_prefill_size` 8192, `max_prefill_tokens` 16384; r1 = 20480 tokens, finished and cached; then r2 = r1 + 4096 new tokens and r3 = 512 fresh tokens are queued together.

| | `main` a95b3e59 | this PR dff3395e | SGLang v0.5.6.post2 (from the source above) |
|---|---|---|---|
| tokens recomputed for r2's prefill | 16384 | 4096 | 4096 |
| prefill passes spent on r2 | 3 | 3 (passes 1 and 2 compute 0 tokens in 0.000 ms) | 1 (prefix 20480 matched once, `extend_input_len` 4096 ≤ chunk) |
| r3 admitted in the same batch as r2 | no | no (r2's 8192-token chunk charge exhausts `rem_chunk_tokens` three times) | yes (4096 + 512 ≤ 8192) |
| zero-duration passes with an all-zero FPM snapshot | 1 | 2 | 0 |

Per-pass trace at dff3395e: pass 1 admits r2 with reused 8192 / computed 0 / 0.000 ms; pass 2 reused 16384 / computed 0 / 0.000 ms; pass 3 reused 20480 / computed 4096 / 85.8 ms; pass 4 admits r3 alone (512 tokens); decode from pass 5. On `main`, passes 2 and 3 each compute 8192 tokens (169 ms and 180 ms).

Consequences of the current design: (a) the phantom passes emit an all-zero FPM snapshot with `wall_time_secs = 0` and re-fire `on_admit` for the same request with a growing reused count; (b) the chunk and input budgets are charged for tokens that turn out to be cached (`charged_input_tokens`, prefill.rs:83), which is the same root cause as the +10–14% Mooncake TTFT gap the description leaves open, not a separate issue; (c) `test_chunked_continuation_reuses_prefix_cached_by_another_request` asserts a 0 / 2 / 2-token three-pass sequence "as SGLang does", where SGLang would run a single non-chunked 4-token prefill. The phantom passes are harmless for the replay engine (each carries an admission event, so the zero-duration guard does not fire), which is why this is a fidelity claim that does not hold rather than a regression.

**Requested changes**

1. Required: correct the SGLang attribution in the `extend_allocation` docstring (sglang_backend.rs:286-293), the continuation comment (prefill.rs:111-112), the test comment (tests.rs:2009-2012) and the PR body. As written they send the next reader to look for a re-match in SGLang that is not there.
2. Recommended, and it closes the gap in (b) as well: implement SGLang's one-time match instead of the per-chunk re-match. Match the whole sequence at first admission with the read-only hash match the LPM policy already uses (`RadixCache::prefix_match_hashes_len`, see policy.rs:41-43; `SglangRequest::new` already hashes the whole prompt), capped at `input_len - 1`; size the chunk from the matched prefix; charge only the uncached tokens; keep `extend_allocation` own-pages-only. The `Option<usize>` return, the eviction moved inside after locking and the lock hygiene in this PR are good and can stay; only the re-match block (sglang_backend.rs:310-323) and the adoption block (348-361) go.

```rust
// now (prefill.rs:66-116)
let extend_input = req.extend_input_len();                 // whole prompt on first admission
...
let chunk_end = req.materialized_tokens + chunk_tokens;
let alloc_tokens = req.sequence_prefix(chunk_end);         // match scoped to this chunk
let prefix_len = if req.materialized_tokens > 0 {
    kv_manager.extend_allocation(alloc_tokens, &mut lease) // re-match + adopt
} else {
    kv_manager.allocate_for_request_lease(alloc_tokens, &mut lease)
};
```

```rust
// proposed: SGLang `Req.init_next_round_input(tree_cache)` semantics — match once, then chunk
let cached_prefix = if req.materialized_tokens == 0 && config.enable_prefix_caching {
    let match_len = req.current_sequence_len().saturating_sub(1);          // input_len - 1
    let hashes = &req.kv_lease.page_hashes()[..match_len / config.block_size];
    kv_manager.cache().prefix_match_hashes_len(hashes)                    // read-only, page-aligned
} else {
    0
};
let start = req.materialized_tokens.max(cached_prefix);
let extend_input = req.current_sequence_len() - start;                     // SGLang extend_input_len, >= 1
let chunk_tokens = if extend_input <= config.chunked_prefill_size { extend_input } else { /* unchanged */ };
let charged_input_tokens = ceil_to_block(chunk_tokens, config.block_size) as f64; // excludes the cached prefix
...
let chunk_end = start + chunk_tokens;
let alloc_tokens = req.sequence_prefix(chunk_end);
let prefix_len = if req.materialized_tokens > 0 {
    kv_manager.extend_allocation(alloc_tokens, &mut lease)          // own pages only, Some(owned)
} else {
    kv_manager.allocate_for_request_lease(alloc_tokens, &mut lease) // locks the same >= cached_prefix path
};
```

The read-only match and the locking allocation run back to back on one thread, and eviction inside `allocate_for_request_lease` happens after the path is locked, so the two agree. With the `input_len - 1` cap every prefill computes at least one token, which also removes the zero-duration pass that already exists on `main` for a fully cached first chunk. Tests: `test_chunked_continuation_reuses_prefix_cached_by_another_request` becomes a single 4-token prefill pass; `chunked_continuation_rematches_longer_cached_prefix` should expect the owned prefix or go; a test that a small request queued behind a mostly cached long request is co-admitted in the same batch would pin the co-admission behaviour.

<details><summary>Probe test used above (appended to <code>crates/core/src/engine/scheduler/sglang/tests.rs</code>, run with <code>cargo test -p aisimulate-core probe_chunked_prefix_reuse_pass_structure -- --nocapture</code>)</summary>

```rust
// ---- review probe (NOT part of the PR; appended locally by the reviewer) ----
mod review_probe {
    use super::*;

    fn probe_args() -> MockEngineArgs {
        MockEngineArgs::builder()
            .engine_type(EngineType::Sglang)
            .block_size(64)
            .num_gpu_blocks(4096)
            .max_num_batched_tokens(Some(16384))
            .max_num_seqs(Some(64))
            .speedup_ratio(0.0)
            .sglang(Some(SglangArgs {
                page_size: Some(64),
                chunked_prefill_size: Some(8192),
                max_prefill_tokens: Some(16384),
                ..Default::default()
            }))
            .build()
            .unwrap()
    }

    fn submit(core: &mut SglangCore, id: u128, tokens: Vec<u32>, max_out: usize) {
        core.receive(DirectRequest {
            tokens,
            max_output_tokens: max_out,
            output_token_ids: None,
            uuid: Some(Uuid::from_u128(id)),
            arrival_timestamp_ms: None,
        });
    }

    #[test]
    fn probe_chunked_prefix_reuse_pass_structure() {
        let mut core = SglangCore::new(probe_args());
        let mut collector = crate::engine::trace::TraceCollector::default();
        // r1: 20480 tokens, finishes after 1 output token -> its prompt is cached.
        let r1: Vec<u32> = (0..20480).collect();
        submit(&mut core, 1, r1.clone(), 1);
        let mut now = 0.0;
        for _ in 0..8 {
            let p = core.execute_pass(&mut collector, now);
            now = p.end_ms + 1.0;
            if core.running.is_empty() && core.waiting.is_empty() {
                break;
            }
        }
        assert!(core.running.is_empty() && core.waiting.is_empty(), "r1 must be finished+cached");
        println!("PROBE r1 cached; clock={now:.3} ms");

        // r2 = r1's 20480 tokens + 4096 new tokens (24576 total, 3 chunks of 8192).
        let mut r2 = r1.clone();
        r2.extend(1_000_000..1_004_096);
        submit(&mut core, 2, r2, 8);
        // r3 = small fresh request queued right behind r2.
        submit(&mut core, 3, (2_000_000..2_000_512).collect(), 8);

        for i in 1..=6 {
            let running_before = core.running.len();
            let waiting_before = core.waiting.len();
            let p = core.execute_pass(&mut collector, now);
            let fpm = p.fpm.as_ref().expect("fpm");
            let adm: Vec<(u128, usize)> = p
                .admissions
                .iter()
                .map(|a| (a.uuid.as_u128(), a.reused_input_tokens))
                .collect();
            let toks: Vec<u128> = p
                .output_signals
                .iter()
                .filter(|s| s.token_id.is_some())
                .map(|s| s.uuid.as_u128())
                .collect();
            println!(
                "PROBE pass {i}: dur_ms={:.3} admissions(uuid,reused)={:?} fpm[prefill_reqs={} prefill_tokens_computed={} prefill_kv={} decode_reqs={} wall_s={:.6}] tokens_emitted_for={:?} running {}->{} waiting {}->{}",
                p.end_ms - now,
                adm,
                fpm.num_prefill_requests,
                fpm.sum_prefill_tokens,
                fpm.sum_prefill_kv_tokens,
                fpm.num_decode_requests,
                fpm.wall_time_secs,
                toks,
                running_before,
                core.running.len(),
                waiting_before,
                core.waiting.len()
            );
            now = p.end_ms + 1.0;
        }
    }
}
```
</details>

tianhaox:
Thanks — you are right on all points about fix 3, and I re-verified them against the v0.5.6.post2 source: `chunked_req.init_next_round_input()` gets no tree cache, only waiting requests match (`fill_ids[:input_len-1]`), and `add_one_req` / `add_chunked_req` charge `extend_input_len` only. The re-match claim was mine and it was wrong.

Rewrote the third commit (d45c7bd3) along the lines of your sketch:

- `prefill.rs`: read-only `prefix_match_hashes_len` over the whole prompt at first admission (capped at `input_len - 1`), chunk sized from the matched prefix, `charged_input_tokens` and `rem_total_tokens` charged for computed tokens only, `allocate_for_request_lease` locks the same path (result clamped to the cap so a duplicate prompt still computes its last token). Zero-output prefix-only replay requests keep the full match — `fully_cached_zero_output_request_is_not_forward_pass_work` pins that contract and SGLang has no such requests.
- `extend_allocation`: back to own-pages-only; kept the `Option<usize>` return, the eviction-after-lock and removed the scheduler-side pre-eviction. The re-match and adoption blocks are gone.
- Attribution corrected in the docstring, the `prefill.rs` comment, the tests and the PR body.
- Tests: your three-pass 0/2/2 test became `test_first_admission_matches_whole_prompt_once` (single 4-token pass, nothing chunked); added `test_fully_cached_prompt_still_computes_its_last_token` and the co-admission test `test_small_request_is_co_admitted_behind_mostly_cached_long_request` (20 cached + 4 new, plus a 2-token request, one 6-token pass); the KV-manager test now asserts the continuation does not re-match.

Your probe scenario now runs as one pass computing 4096 tokens with r3 co-admitted. Mooncake TP8 vs HiSim: TPOT 23.43 / 23.40 ms, E2E mean 4635 / 4678 ms, first-admission prefix reuse 0.21 / 0.21 (was 0.11); TP4 E2E 20.2 / 20.9 s. TTFT mean is still +12–14% even with the budget over-charge fixed, so (b) was not the cause of that gap; the next candidate is that the mocker rejects a request whose `extend_input` exceeds the remaining chunk budget instead of taking a partial chunk like `add_one_req` — I left that out of this PR. `cargo test --workspace` 1147 pass, fmt/clippy clean.

crates/core/src/engine/kv_manager/sglang_backend.rs
tianhaox: Fixed in d45c7bd3: first admission now matches the whole prompt once (`prefix_match_hashes_len`, cap `input_len - 1`), chunk sized from the matched prefix, budgets charged for computed tokens only; `extend_allocation` is own-pages-only again and the attribution text is corrected. Details in the PR comment.

crates/core/src/engine/scheduler/sglang/prefill.rs
tianhaox: Fixed in d45c7bd3: first admission now matches the whole prompt once (`prefix_match_hashes_len`, cap `input_len - 1`), chunk sized from the matched prefix, budgets charged for computed tokens only; `extend_allocation` is own-pages-only again and the attribution text is corrected. Details in the PR comment.

crates/core/src/engine/scheduler/sglang/tests.rs
tianhaox: Fixed in d45c7bd3: first admission now matches the whole prompt once (`prefix_match_hashes_len`, cap `input_len - 1`), chunk sized from the matched prefix, budgets charged for computed tokens only; `extend_allocation` is own-pages-only again and the attribution text is corrected. Details in the PR comment.

crates/core/src/engine/scheduler/sglang/core.rs
```diff
@@ -750,16 +777,27 @@ impl SglangCore {
...
+        for req in decode.requests.drain(..) {
+            self.waiting.push_back(req);
         }
 
         if decode.retracted_any {
-            self.new_token_ratio = self.config.init_new_token_ratio;
+            // `ScheduleBatch.retract_decode` re-estimates the output reservation ratio from the
+            // remaining batch: min(1, (decoded + RETRACT_DECODE_STEPS * n) / (max_new + 1)).
+            let remaining = self.running.len();
+            let total_decoded: usize = self.running.iter().map(|req| req.output_len()).sum();
```
jasonqinzhou: Please compute or capture this estimate at the retraction point before the decode forward runs. SGLang's `update_running_batch` calls `retract_decode` before `prepare_for_decode`, so its numerator and denominator describe the surviving requests before the next forward (https://github.com/sgl-project/sglang/blob/5c8bd8b51b53b9b39eb1edec582ee43b21002106/python/sglang/srt/managers/scheduler.py#L1916-L1960). Here `simulate_decode_step_with_sampler` first retracts, then appends this pass's token or speculative burst and removes completed requests; only afterward do these sums run. That adds the just-produced outputs to every survivor, and if the last survivor completes it makes `remaining == 0` and resets the ratio to 0, causing later admissions to under-reserve output KV. Return the pre-forward estimate with `DecodeResult` (or compute it inside the retraction path), and add a regression where retraction and survivor completion occur in the same pass.

crates/core/src/engine/scheduler/sglang/core.rs
```diff
@@ -736,7 +757,13 @@ impl SglangCore {
             self.speculative_sampler.as_mut(),
             decode_start_ms,
             true,
+            !prefill_pass,
```
jasonqinzhou: When `prefill_pass` is true, this call still supplies `self.speculative_sampler`; `simulate_decode_step_with_sampler` therefore samples up to `aic_nextn + 1` outputs even though `charge_time` is false. The existing `test_mtp_batch_applies_request_bursts_in_stable_order` demonstrates this at the current head: its first prefill pass emits 9 signals for 3 requests with `aic_nextn=2` and acceptance rates `1,1`, so every request gets three free output tokens. That contradicts the new contract above that the prefill forward supplies only the first token and materially underestimates MTP latency. Gate the sampler/max burst off for this bookkeeping path (or separate first-token bookkeeping from decode simulation), and update the test to require exactly one first-pass signal per freshly prefilling request.

crates/core/src/engine/scheduler/sglang/prefill.rs
```diff
@@ -116,11 +131,15 @@ pub(super) fn get_new_batch_prefill(
...
+            // Continuation: extend the request's own pages; no re-match.
+            kv_manager.extend_allocation(alloc_tokens, &mut lease)
         } else {
-            kv_manager.allocate_for_request_lease(alloc_tokens, &mut lease)
+            // Locks the same path the read-only match above saw (same thread, eviction inside
+            // happens after locking), so the locked prefix is at least `cached_prefix`. Clamp to
+            // the SGLang cap so a fully cached prompt still computes its last token.
+            kv_manager
+                .allocate_for_request_lease(alloc_tokens, &mut lease)
+                .map(|matched| matched.min(start))
```
jasonqinzhou: This clamps only the value returned to prefill accounting. `allocate_for_request_lease` has already matched every complete hash in `alloc_tokens`, used that unbounded match to reduce `required_pages`, adopted those pages, and stored the unbounded length in `lease.cached_tokens`. For a fully cached five-token prompt with block size 1, `start` is correctly capped at 4, but this allocator still records all 5 tokens as cached and consumes no page for the token the simulator says it computed. Under tight KV this can admit work or avoid a retraction that SGLang would not. Please pass a maximum match length into the allocator (or lock/allocate from the already capped match) so physical ownership and `cached_tokens` honor the same cap, and extend the fully-cached regression to assert cache/available-token state as well as FPM counts.

crates/core/src/engine/scheduler/sglang/prefill.rs
```diff
@@ -63,8 +63,30 @@ pub(super) fn get_new_batch_prefill(
     while can_run.len() < available_running_slots
         && let Some(mut req) = waiting.pop_front()
     {
-        let extend_input = req.extend_input_len();
-        if extend_input == 0 {
+        // SGLang `Req.init_next_round_input(tree_cache)`: a waiting request matches its whole
```
jasonqinzhou, line 73: This implementation, the retraction formula, and their tests are explicitly adapted from SGLang v0.5.6.post2, but the modified Rust files still carry only the NVIDIA header, the module reference is an unversioned path, and neither canonical notice records these derived files. The repository's third-party gate requires the upstream repository, immutable revision, original paths, copyright owner, license, and modification status in a file header or adjacent README plus matching entries in both `THIRD_PARTY_NOTICES.md` files. Please record SGLang commit `5c8bd8b51b53b9b39eb1edec582ee43b21002106`, the relevant `scheduler.py` / `schedule_batch.py` / `schedule_policy.py` paths, the SGLang Team copyright and Apache-2.0 license, keep the two notice files byte-identical, and run `python3 scripts/check_packaged_legal_files.py`; if maintainers believe this is not substantially derived, please get the required maintainer/OSRB determination before merge.

jasonqinzhou (dismissed):
**REQUEST_CHANGES · 5/10 · high confidence**

## What this PR does

This PR makes the SGLang simulator schedule prefill-only passes ahead of decode work, aligns retraction queueing and reservation-ratio behavior, and now matches a waiting request's whole prompt once before chunking the uncached suffix.

The updated head responds directly to the earlier prefix-cache review with a focused one-time-match implementation and regression tests for single-pass reuse, last-token computation, co-admission, and continuation behavior.

## Why this score

The 5.0 technical score reflects that the incremental prefix-cache correction fixes the previous head's per-continuation re-match finding, but three P1 issues and one P2 issue remain: MTP bursts are still free during prefill-only passes, the retraction ratio is still computed from post-forward state, required SGLang attribution is missing, and the capped prefix is not enforced in allocation state. The canonical recommendation is BLOCKED rather than REQUEST_CHANGES because the required independent helper again observed Sonnet and Haiku instead of the requested claude-fable-5.

crates/core/src/engine/scheduler/sglang/core.rs
```diff
@@ -720,13 +725,29 @@ impl SglangCore {
...
+        } else {
+            self.running
+                .iter()
+                .filter(|req| req.remaining_output_tokens() > 0)
+                .map(|req| req.current_sequence_len() as u64)
+                .collect()
+        };
 
         let decode_start_ms = now_ms + prefill_time.as_secs_f64() * 1000.0;
         let mut decode = simulate_decode_step_with_sampler(
```
dreamtalen: It could be cleaner to separate first-token bookkeeping from `simulate_decode_step_with_sampler`. Prefill only needs to apply one token, while this function also includes decode-specific retraction, speculative sampling, and timing logic, as highlighted by the MTP issue @jasonqinzhou raised below.

crates/core/src/engine/scheduler/sglang/core.rs
```diff
@@ -691,9 +695,9 @@ impl SglangCore {
             ),
         };
 
-        if admit.oom {
-            self.new_token_ratio = self.config.init_new_token_ratio;
-        }
+        // SGLang's `get_new_batch_prefill` only marks the running batch full on
+        // `AddReqResult.NO_TOKEN`; it does not reset `new_token_ratio` there.
+        let _ = admit.oom;
```
dreamtalen: nit: we could remove `oom` here since it's no longer used

crates/core/src/engine/scheduler/sglang/core.rs
tianhaox: Done in cbd7b48f: the estimate is now computed in the step simulation right after `check_decode_mem_for_burst`, over the survivors before this pass's tokens are appended or completions removed, and returned as `DecodeResult::new_token_ratio_estimate`; core.rs just applies it. Regression `retraction_ratio_is_estimated_from_survivors_before_the_forward`: r2 retracted, r1 completes in the same step, estimate is the pre-forward (1 + 20)/(2 + 1) → 1.0 rather than 0.

crates/core/src/engine/scheduler/sglang/core.rs
tianhaox: Done in cbd7b48f: prefill passes now go through `simulate_prefill_first_tokens`, which shares the KV bookkeeping with the decode step but forces `max_burst = 1`, no sampler and zero time (`StepKind::PrefillFirstToken`). `test_mtp_batch_applies_request_bursts_in_stable_order` now expects exactly one first-pass signal per request (3, not 9) and the 3-token bursts only on the decode passes that follow.

crates/core/src/engine/scheduler/sglang/prefill.rs
tianhaox: Done in cbd7b48f: `allocate_for_request_lease_capped(token_ids, lease, max_prefix_tokens)` limits the locking match to the same page-aligned cap the admission match used, so `cached_tokens`, the adopted pages and the newly allocated slot all honour `input_len - 1`; the accounting-only clamp is gone. Unit test `allocation_match_is_capped_like_sglang` (fully cached 5-token prompt: prefix 4, cached_tokens 4, exactly one token allocated) and `test_fully_cached_prompt_still_computes_its_last_token` now also asserts the pool state.

crates/core/src/engine/scheduler/sglang/prefill.rs
tianhaox: The modules are behavioural re-implementations, not copied source: no SGLang function bodies, comments or test data are reproduced, the constants (0.7, 0.14, 600, 20) are the documented defaults and the ratio formula is a one-line behavioural fact. The pre-existing SGLang and vLLM scheduler emulations in this tree (from the Dynamo mocker) carry the same NVIDIA-only headers and no notice entry, so this PR follows the established treatment. For traceability cbd7b48f adds module docs naming the revision (v0.5.6.post2, 5c8bd8b5) and the files the semantics were read from. Whether behavioural ports should get a THIRD_PARTY_NOTICES entry is a module-wide maintainer/OSRB decision that I would rather not settle in this PR; happy to add the entry if maintainers decide it is required.

crates/core/src/engine/scheduler/sglang/core.rs
tianhaox: Agreed and done in cbd7b48f: `simulate_prefill_first_tokens` is the prefill-pass entry point; it shares the completion and KV-slot bookkeeping with the decode step via an internal `StepKind`, but has no sampler, no burst and no modeled time. This also resolves the MTP finding.

crates/core/src/engine/scheduler/sglang/core.rs
tianhaox: Removed in cbd7b48f (the `oom` field itself, plus the now-unused `retracted_any`, `extend_input_len` and the `charge_time` parameter).

tianhaox:
Pushed cbd7b48f addressing the second review round (@jasonqinzhou, @dreamtalen):

- **MTP bursts on prefill passes (P1)** — prefill passes now use `simulate_prefill_first_tokens` (`StepKind::PrefillFirstToken`: no sampler, `max_burst = 1`, zero time). The MTP burst test expects one first-pass signal per request and bursts only on decode passes. This is also the separation @dreamtalen suggested.
- **Retraction ratio from post-forward state (P1)** — the estimate is now computed at the retraction point over the survivors before the forward and carried in `DecodeResult::new_token_ratio_estimate`; regression `retraction_ratio_is_estimated_from_survivors_before_the_forward` covers retraction + survivor completion in one pass (pre-forward estimate instead of 0).
- **Cap not enforced in allocation state (P1)** — `allocate_for_request_lease_capped` limits the locking match to the admission cap so `cached_tokens`, adopted pages and the fresh slot all stop at `input_len - 1`; `allocation_match_is_capped_like_sglang` asserts the lease and pool state, `test_fully_cached_prompt_still_computes_its_last_token` asserts pool accounting.
- **Attribution (P2)** — module docs now name the SGLang revision and files the semantics were modelled on (v0.5.6.post2, 5c8bd8b5). No SGLang source is copied, and the pre-existing SGLang/vLLM scheduler emulations in this tree carry the same NVIDIA-only headers, so I have not added a `THIRD_PARTY_NOTICES.md` entry; if maintainers/OSRB decide behavioural ports need one, I will add it for the module as a whole.
- Nits: `oom`, `retracted_any`, `extend_input_len` and the `charge_time` parameter are gone.

`cargo test --workspace` 1149 pass, fmt/clippy clean. Mooncake numbers are unchanged from the previous head (no MTP, no retraction on that trace): TP8 TPOT 23.43 / 23.40 ms vs HiSim, E2E 4635 / 4678 ms, first-admission prefix reuse 0.21 / 0.21.

crates/core/src/engine/scheduler/sglang/decode.rs
```diff
@@ -279,11 +345,16 @@ pub(super) fn simulate_decode_step_with_sampler(
...
+    let max_burst = if kind == StepKind::PrefillFirstToken
+        || config.worker_type == crate::engine::common::protocols::WorkerType::Prefill
+    {
         1
     } else {
         config.speculative_max_tokens.unwrap_or(1)
     };
+    if kind == StepKind::PrefillFirstToken {
+        sampler = None;
+    }
```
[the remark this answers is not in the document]
tianhaox: Agreed, fixed in 9d6911ce: `StepKind::PrefillFirstToken` no longer calls `check_decode_mem_for_burst`; it only evicts cached pages to make room for the first-output slots and leaves `new_token_ratio_estimate` as `None`. If the slots still do not fit, the tokens are produced by the next decode step, which runs the normal retraction over the whole running batch (SGLang's `update_running_batch`). New test `test_prefill_pass_never_retracts_first_token_slots_wait_for_decode`: two prompts fill the pool in one prefill pass → no pressure event, no tokens; the next decode pass retracts one and emits the other's token.

crates/core/src/engine/scheduler/sglang/prefill.rs
```diff
@@ -116,17 +135,18 @@ pub(super) fn get_new_batch_prefill(
                     req.uuid, req.materialized_tokens
                 );
             }
-            kv_manager
-                .extend_allocation(alloc_tokens, &mut lease)
-                .then_some(req.materialized_tokens)
+            // Continuation: extend the request's own pages; no re-match.
+            kv_manager.extend_allocation(alloc_tokens, &mut lease)
```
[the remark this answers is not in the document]
tianhaox: Agreed, fixed in 9d6911ce: `RadixRequestLease` now records `admission_reused_tokens` (the radix match made when the request was admitted from the waiting queue). Admission events report that value for every chunk, and the cache-hit metric (`PrefillFpmItem::cache_reused_tokens` → `sglang_cache_hit_tokens`/`sglang_cache_total_tokens`) credits it once on the first chunk and zero on continuations; `prefix_tokens` keeps the KV context for the forward-pass model. New test `test_chunked_cold_prompt_reports_no_cache_reuse` asserts a cold 12-token prompt split in two chunks reports reuse 0 on the continuation while `sum_prefill_kv_tokens` is 6.

crates/core/src/engine/scheduler/sglang/tests.rs
```diff
@@ -1526,19 +1605,90 @@ mod forward_pass_metrics {
...
+        assert_eq!(fpm3.num_decode_requests, 2, "r1 + r2 both in decode step");
         assert!(
-            fpm2.sum_decode_kv_tokens > 0,
+            fpm3.sum_decode_kv_tokens > 0,
             "decode requests should have KV context"
         );
     }
 
+    #[test]
+    fn test_prefill_pass_stalls_running_decodes_and_charges_no_decode_time() {
```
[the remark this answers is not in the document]
tianhaox: Added in 9d6911ce: pass 2 now asserts `num_decode_requests == 0` and `sum_decode_kv_tokens == 0`.

tianhaox:
Pushed 9d6911ce for the three CodeRabbit findings (prefill passes no longer run the decode retraction check; continuation chunks no longer count as radix-cache reuse; FPM decode assertions in the stall test). Behaviour on the aligned scenarios is unchanged (Mooncake TP8: TPOT 23.43 / 23.40 ms vs HiSim, E2E 4635 / 4678 ms), and the all-admission `prefix_cache_reused_ratio` now equals the first-admission ratio (0.21) instead of the inflated 0.66 it reported before. `cargo test --workspace` 1155 pass; Fast CI green on the new head.

crates/core/src/engine/scheduler/sglang/tests.rs
```diff
@@ -1835,6 +1991,339 @@ mod forward_pass_metrics {
...
+            "no retraction in a prefill pass"
+        );
+        assert!(
+            pass1.output_signals.is_empty(),
+            "first tokens wait for KV slots"
+        );
+        assert_eq!(core.running.len(), 2);
+        assert!(core.waiting.is_empty());
+
+        let pass2 = core.execute_pass(&mut collector, pass1.end_ms);
```
dreamtalen: Since both requests have max_output_tokens=1, I would expect them to finish in this prefill pass without an additional decode forward. Could we check whether deferring the first tokens here matches SGLang's behavior?

dreamtalen (approved):
Approving conditional on addressing the following `max_output_tokens=1` case.

crates/core/src/engine/scheduler/sglang/tests.rs
tianhaox: Good catch, and you are right: SGLang finishes a `max_new_tokens=1` request in `process_batch_result_prefill` → `check_finished()`, right after the prefill batch, and the sampled token never gets a KV slot; only requests that keep decoding get their slot in `prepare_for_decode` at the next decode step (where `check_decode_mem`/retraction live). Fixed in f8d3d4df: the prefill bookkeeping is now its own function (`prefill_first_tokens`). Requests whose first token is also their last complete in the prefill pass with no slot (`append_final_output_token`); continuing requests take their first-output slot after evicting cached pages, and only if that still does not fit is the token left for the next decode step. This test became `test_single_token_requests_finish_in_the_prefill_pass_without_kv_slots` (both complete in the prefill pass, no pressure event, no page spent on outputs); `test_prefill_pass_gives_continuing_requests_their_first_token_without_retraction` covers the `max_output_tokens = 2` case.

crates/core/src/engine/scheduler/sglang/prefill.rs
```diff
@@ -138,18 +168,19 @@ pub(super) fn get_new_batch_prefill(
 
         admissions.push(AdmissionEvent {
             uuid: req.uuid,
-            reused_input_tokens: prefix_len,
+            reused_input_tokens: admission_reused_tokens,
```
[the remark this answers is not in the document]
tianhaox: Done in f8d3d4df: `AdmissionEvent` is emitted only for a request's first chunk; continuation chunks are the same admission still running. `readmission_count` on the Mooncake trace (no retractions) is now 0 for every request instead of counting chunks.

tianhaox:
Pushed f8d3d4df (head now 6 commits):

- @dreamtalen's `max_output_tokens=1` point: requests whose first token is also their last now finish in the prefill pass without a KV slot, matching SGLang's `check_finished()` after `process_batch_result_prefill`; continuing requests take their first-output slot in the same bookkeeping step (evicting cached pages if needed) and only fall back to the next decode step when the pool is genuinely full. The prefill bookkeeping is a separate function (`prefill_first_tokens`) rather than a mode of the decode step.
- CodeRabbit: `AdmissionEvent` is emitted once per request (continuation chunks no longer count as readmissions); the `extend_allocation` doc now states that a `None` result may already have evicted cached pages.

`cargo test --workspace` 1156 pass, fmt/clippy clean. Aligned-scenario numbers are unchanged (Mooncake TP8 vs HiSim: TPOT 23.43 / 23.40 ms, E2E 4635 / 4678 ms, first-admission reuse 0.21 / 0.21); `readmission_count` on that trace is now 0 for every request, as it should be with no retractions.

Arsene12358 (approved):
**Approving at f8d3d4df.** The change I requested is in, and the follow-up commits also close every soft note from my first pass.

**Re-verified locally at f8d3d4df** (Full CI has not run for this PR yet, see below): `cargo test --workspace` 1211 passed / 0 failed / 1 ignored; `cargo fmt --all -- --check` clean; no clippy warnings in the six changed files; a test merge onto current `main` (5710bc6a) is clean with no overlapping files. The probe from my first review (20480 cached tokens + 4096 new, a 512-token request queued behind it) now runs as a single prefill pass computing 4608 tokens with both requests co-admitted and the cached prefix reported once (reused 20480 / 0), then decode from pass 2. That is the SGLang column of the original table.

**Fix 3 as landed** matches the SGLang source it now cites: whole-prompt read-only match once at admission (`prefix_match_hashes_len`), capped at `input_len - 1` and enforced in the allocator's lease state by `allocate_for_request_lease_capped`, chunk sized from the matched prefix, budgets charged for computed tokens only, and `extend_allocation` extends the request's own pages without re-matching. The re-match and page-adoption blocks are gone, and the attribution text in the docstring, the `prefill.rs` comment, the tests and the PR body is correct. `test_first_admission_matches_whole_prompt_once`, `test_small_request_is_co_admitted_behind_mostly_cached_long_request`, `test_fully_cached_prompt_still_computes_its_last_token`, `allocation_match_is_capped_like_sglang` and `chunked_continuation_extends_own_pages_without_rematching` pin the behaviour.

**Also closed by the later commits:** prefill passes emit exactly one first token per request with no speculative burst and no decode time (`StepKind::PrefillFirstToken`); the retraction ratio is taken from the pre-forward survivors; prefill passes never retract and defer first-token slots to the next decode step when they do not fit; single-token requests finish in the prefill pass without a KV slot (I checked that `finish` clamps caching to the lease's materialized tokens, so the uncomputed final token is never cached; disaggregated prefill-role requests carry an output limit of 1 and take this path, and the disaggregation tests pass); cache-hit metrics credit only the admission match; one `AdmissionEvent` per request.

**Outside my approval:** the third-party attribution question raised in the other review (whether a behavioural re-implementation needs a `THIRD_PARTY_NOTICES.md` entry) is a maintainer / OSRB call, and Full CI (`cargo test --workspace`) has only run locally so far since no `pull-request/120` branch exists; a vetter should trigger it before merge.

PR #127, opened 2026-09-08
fix(generator): preserve Sweeper KV capacity [6738325]

jasonqinzhou:
## Summary

- accept Sweeper recommendation candidates whose KV capacity uses backend defaults or a fixed block count
- preserve default-capacity semantics for vLLM, SGLang, and TensorRT-LLM
- render fixed capacity as `--num-gpu-blocks-override` for vLLM and as an exact token capacity for SGLang/TensorRT-LLM
- reject ambiguous candidates that specify both a memory fraction and fixed blocks

## Landing sequence

- canonical base: `main`
- clean source base: `e08f0c3f33bcd725669fbe105085a363eeadd4a9`
- after the main change merges, a fresh cherry-pick PR will target `release/0.12.0`

## Validation

- generator unit suite plus release backend-contract goldens: 350 passed, 4 skipped
- TensorRT-LLM fixed-capacity artifact probes: rc20 and rc23 both render `max_tokens: 16384` without a memory-fraction fallback
- clean no-commit merge with PR #128; combined generator/release-contract suite: 352 passed, 4 skipped
- changed-file Ruff check and format check: passed
- `git diff --check`: passed
- full exact-head CI: [run 34266444811](https://github.com/ai-dynamo/aisimulate/actions/runs/34266444811) (in progress)

PR #130, opened 2026-09-08
fix(generator): preserve Sweeper KV capacity [6738325]

jasonqinzhou:
## Summary
- preserve backend-default KV capacity when the Sweeper candidate omits an explicit memory fraction
- lower fixed GPU-block candidates to the backend-specific vLLM, SGLang, and TRT-LLM capacity fields
- cover current TRT-LLM rc14/rc23 templates and release backend contracts

## Validation
- `350 passed, 4 skipped` across generator unit tests and the release backend contract suite
- changed-file Ruff check and format check
- `git diff --check`

## Landing sequence
This is the canonical `main` change. It supersedes the release-first PR #127; after this merges, a fresh cherry-pick PR will target `release/0.12.0`.


jasonqinzhou:
Superseded by #127, whose existing head was safely rewritten to this same validated `main`-based SHA and retargeted to `main`.

PR #137, opened 2026-09-08
AIC-1933: expose typed operation-energy evidence

jasonqinzhou:
## Summary

- add public, runtime-neutral timing evidence types for operation and phase latency, modeled energy (W-ms), covered latency, and source provenance
- make the op-level AIC timing provider collect name-folded evidence across the Rust/Python boundary while keeping uncovered energy explicitly absent
- keep built-in fixed/polynomial timing and whole-model FPM latency-only; preserve FPM decode's exact total-past-KV query path
- document the contract and compile it from the external public-API fixture

Linear: https://linear.app/nvidia/issue/AIC-1933/power37perfmodel-expose-typed-operation-energy-evidence

## Validation

- `cargo test -p aisimulate-core engine::timing` (6 passed)
- `cargo test -p aisimulate-core --features embed-python python::tests::` (11 passed)
- external public-API fixture (5 passed; run standalone with a temporary local workspace marker)
- `cargo test -p aisimulate-core --features embed-python`: library 1,165 passed / 1 ignored and the artifacts, engine, facade, and firewall integration targets passed; the environment-dependent embedded round-trip then failed because the discoverable local AIC package emits EngineSpec v15 while this checkout requires v17
- strict repository-wide Clippy remains red on the existing baseline; the changed files add no new warning (the only warning in a changed file is the pre-existing large `ExecutionPayload` at `python.rs:29`)

## Stacking note

This is the typed provider boundary for the replay accumulation work in AIC-1935 / PR #62. That PR should consume `TimingModel::evidence_summary()` when it is restacked.


jasonqinzhou:
Superseded by #138, which preserves the same implementation on a clean DCO-signed branch without rewriting published history.

PR #141, opened 2026-09-08
AIC-1938: add fail-closed modeled-power qualification ledger

jasonqinzhou:
## Summary

The former qualification check validated report contents without reproducing execution. This change adds a versioned modeled-power release ledger and an executable data-gate producer, then makes Fast CI regenerate and verify evidence against the exact checkout.

- Scan every parquet in the checked-out source tree, with a complete-tree digest, per-file digests, row counts, units, pairing and non-finite checks, and explicit paired-zero unavailable sentinels.
- Generate and upload the current-head report and ledger in CI; never discover data through an unrelated importable package.
- Re-scan and compare the complete report during execution verification. Reject stale heads, dirty sources, forged report values, and relabeled old evidence.
- Require an execution verifier for every passing automated release gate. Planned runtime, parity, packaging, and silicon-accuracy gates remain unqualified.
- Preserve the historical report for audit history without using it to qualify the candidate.

The checked-in release state remains `not_qualified`. This PR supplies qualification machinery, not silicon-accuracy or full power-support evidence.

## Validation

Codex reviewed commit `1f1b36cadf3c570cd2fb0c9a9539501e34bef6a9`, refreshed against main `1694ffe7cf6785a8e31527f862bdd7735aadd6ca`.

- 264 qualification, CI workflow, selection-policy, and CODEOWNERS tests passed.
- Executed the producer on the clean commit and independently regenerated its result through `--verify-execution`; both passed.
- Fast CI and CodeRabbit passed at this head. Full CI is admitted and queued; applicable CODEOWNER approval and Kai's re-review remain required.
- Source/legal checks, generated CODEOWNERS, Ruff, and whitespace checks passed.

## Boundaries and tracking

Modeled power remains active forward-pass power per GPU. The 90% coverage gate and existing 1% AIC parity tolerance are preserved. Silicon MAPE and sample-count thresholds remain unapproved. AFD and EPD power integration stay explicitly blocked.

Closes: https://linear.app/nvidia/issue/AIC-1938/power77qarelease-qualify-parity-accuracy-and-migration-closeout
Related: #62 and AIC-1932 through AIC-1937.

## Updated stack base

This PR is based on #208 (`codex/power-contract-pr136-recovery`), the replacement for closed #136. Qualification machinery is a separate branch; final release qualification still depends on #138, #139, #62, #142, and #144. The existing migration-guide conflict remains to be resolved before merge.

Only the PR base and this stacking note changed. The validation results above describe the recorded head against its previous base; they do not establish readiness against #208. Fresh integration checks and applicable CODEOWNER approval remain required.


scripts/validate_power_qualification.py
```diff
@@ -0,0 +1,509 @@
...
+    _expect(isinstance(gates, list) and gates, "gates must be a non-empty array", errors)
+    if isinstance(gates, list):
+        for index, gate in enumerate(gates):
+            _validate_gate(gate, index, errors)
+        ids = [gate.get("id") for gate in gates if isinstance(gate, dict)]
+        if all(isinstance(gate_id, str) for gate_id in ids):
+            _expect(len(ids) == len(set(ids)), "gate ids must be unique", errors)
+        _validate_required_coverage(document, errors)
+
+    if require_release_ready or document.get("release_state") == "qualified":
```
jasonqinzhou, line 1292: [P1] Bind release evidence to one explicit candidate revision

The release-ready path only checks that each `source_revision` has 40 hex characters. A `qualified` ledger can therefore mix stale or unrelated commits across passing gates and still succeed, despite the documented exact-source guarantee. Add one candidate revision to the ledger, require an explicit expected revision at release closeout, and verify every release-blocking passing evidence record matches it. Please include a positive ready-state test so this remains usable as well as fail-closed.

docs/power/qualification-matrix.json
```diff
@@ -0,0 +1,298 @@
...
+          "when": {},
+          "expectation": "aic_parity",
+          "thresholds": {"maximum_relative_error_ratio": 0.01, "reference_must_be_immutable": true}
+        }
+      ],
+      "execution": {
+        "kind": "automated",
+        "status": "pending",
+        "command": "python -m pytest -q crates/core/parity_tests/perfmodel/test_engine_step_parity.py -k power",
+        "command_status": "available",
```
jasonqinzhou: [P1] Do not mark the zero-energy parity command as qualifying evidence

This gate requires non-zero covered power cases, but the selected `-k power` fixture is documented in the surrounding parity test as exercising well-typed zero-energy surfaces after the power-data prune. With `command_status: available`, a passing run can be attached and the release validator will accept the gate even though its own `aic_parity` assertion has not been demonstrated. Please keep this command planned/non-qualifying until a non-zero frozen-AIC fixture exists.

docs/power/qualification-matrix.json
```diff
@@ -0,0 +1,298 @@
...
+          "when": {},
+          "expectation": "packaged_output",
+          "thresholds": {"distribution_count": 1, "wheel_smoke_required": true, "packaged_data_required": true}
+        }
+      ],
+      "execution": {
+        "kind": "automated",
+        "status": "pending",
+        "command": "python scripts/build_release_artifacts.py --check-only",
+        "command_status": "available",
```
jasonqinzhou: [P1] The packaging command does not exercise the asserted gate

`build_release_artifacts.py --check-only` only validates manifests. It does not build/install the wheel, smoke the CLI, or verify packaged model data, while this gate explicitly requires wheel smoke and packaged data. Because the ledger accepts any pass evidence for an available command, this can falsely qualify packaging. Mark it planned until the full artifact-contract job is represented, or point it at an executable check that proves all three assertions.

docs/power/qualification-matrix.schema.json
```diff
@@ -0,0 +1,177 @@
...
+      "minItems": 1,
+      "items": {"$ref": "#/$defs/gate"}
+    }
+  },
+  "$defs": {
+    "stringArray": {
+      "type": "array",
+      "minItems": 1,
+      "uniqueItems": true,
+      "items": {"type": "string"}
```
jasonqinzhou: [P2] Encode the supported dimensions in the published schema

The interchange schema currently accepts any string for every matrix and `when` dimension (for example, `runners: ["bogus-runner"]` passes Draft 2020-12 validation), while the deterministic validator rejects it. Tooling that consumes the advertised schema can therefore accept a ledger the release checker considers invalid. Give each dimension its supported enum and add a schema-validation regression so the machine-readable contract cannot silently drift from the validator.

scripts/validate_power_qualification.py
```diff
@@ -0,0 +1,509 @@
...
+    errors: list[str] = []
+    _expect_keys(
+        document,
+        {"$schema", "schema_version", "release_target", "release_state", "policy", "gates"},
+        "document",
+        errors,
+    )
+    if not isinstance(document, dict):
+        raise QualificationError("\n".join(errors))
+    _expect(document.get("schema_version") == "1.0", "schema_version must be '1.0'", errors)
```
jasonqinzhou: [P2] Make the dependency-free validator honor the declared schema identity

The document requires `$schema`, but `validate_document` never checks its value, so replacing it with an unrelated URI still passes the canonical CI command even though the published schema describes a fixed qualification format. Since this validator intentionally runs without `jsonschema`, it should reject a ledger that no longer declares `./qualification-matrix.schema.json`, with a regression covering the mismatch.

scripts/validate_power_qualification.py
```diff
@@ -0,0 +1,509 @@
...
+    modeled = _gate_by_id(document, "modeled-power-matrix", errors)
+    for name, values in {
+        "model_families": {"dense", "moe"},
+        "deployments": {"aggregated", "disaggregated"},
+        "runners": {"standard", "dynamo"},
+        "timing_backends": {"default_op_level"},
+        "coverage_states": {"covered", "undercovered"},
+    }.items():
+        _expect_dimension(modeled, name, values, errors)
+    modeled_expectations = {item.get("expectation") for item in modeled.get("assertions", []) if isinstance(item, dict)}
```
jasonqinzhou: [P1] Pin every release-critical matrix and assertion, not just selected fields

The release-specific coverage check leaves several qualification claims mutable: for example, `aic-power-parity.runners` can be changed away from `standard`, `application-wheel.thresholds.wheel_smoke_required` can be set to false, and the modeled/unsupported/operation threshold objects can be weakened while `--require-release-ready` still accepts the ledger. A release ledger is itself editable input, so the fail-closed validator must compare each required gate matrix and assertion contract against the declared 0.13 policy (including positive coverage, no-fabrication, reconciliation, data, parity, accuracy, and packaging) before trusting attached pass evidence.

scripts/validate_power_qualification.py
jasonqinzhou: Fixed in def18eb344a3bf5015059c53affecce5c9073cea. Release qualification now requires an independently supplied expected revision, a matching top-level candidate revision, and matching source revisions on every release-blocking evidence record. Positive and stale/mixed-revision regressions cover the ready-state path.

docs/power/qualification-matrix.json
jasonqinzhou: Fixed in def18eb344a3bf5015059c53affecce5c9073cea. The zero-energy parity surface is no longer labeled qualifying: the gate points to the future non-zero AIC parity suite and remains planned, so it cannot transition to passed.

docs/power/qualification-matrix.json
jasonqinzhou: Fixed in def18eb344a3bf5015059c53affecce5c9073cea. The manifest-only command was removed from this gate. Packaging now remains planned against a dedicated artifact test whose documented contract includes build/install, CLI smoke, and packaged power-data verification.

docs/power/qualification-matrix.schema.json
jasonqinzhou: Fixed in def18eb344a3bf5015059c53affecce5c9073cea. The schema now enumerates every supported value for matrix and assertion dimensions, rejects unknown `when` keys, pins the declared schema path, and is exercised directly with Draft 2020-12 validation plus a bogus-runner regression.

scripts/validate_power_qualification.py
jasonqinzhou: Fixed in def18eb344a3bf5015059c53affecce5c9073cea. The dependency-free validator now requires the exact local schema identity, with a regression for a substituted URI; related scalar type checks also remain fail-closed.

scripts/validate_power_qualification.py
jasonqinzhou: Fixed in def18eb344a3bf5015059c53affecce5c9073cea. The validator now pins release-blocking status, all five dimensions, and complete assertion/threshold contracts for the functional, no-fabrication, operation, data, parity, silicon, packaging, AFD, and EPD gates. Mutation regressions cover every contract family and prevent downgrading a blocking gate.

jasonqinzhou:
[P1] Refresh the PR description to the exact reviewed head

The current description still identifies `06121d0e` as the Codex-reviewed commit and reports the pre-repair test inventory. It also omits the locally reproduced failure of the only currently available power-data invariant gate. I am updating the authored PR metadata to name exact head `def18eb3`, the expanded 145-test validation, all six fixed self-review findings, and the explicit remaining qualification/CI caveats.

docs/cli/migrate-from-aiconfigurator.md
```diff
@@ -82,6 +82,12 @@ does not make a capability available through the unified path.
 | Predicted `power_w` and power coverage | No AIC-equivalent unified output | Optional runner metadata is not a replacement for AIC power analysis |
 | `--save-dir`, `--deployment-target` | No direct flag mapping on the new `aisimulate` CLI | `aisimulate recommend --output-dir` writes recommendation results and prediction-ready configs, not deployment manifests. For programmatic generation, select a candidate and use `from_sweeper_candidate(...)` with `generate_from_request(...)`; continue using `aiconfigurator cli generate` for a direct CLI workflow. |
 
+The modeled-power migration is tracked by the
+[0.13 qualification ledger](../power/README.md). That ledger remains
+`not_qualified` until its release-blocking functional, parity, silicon, and
+packaging gates carry immutable passing evidence. Its presence does not change
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in 4771d092b897ce8ff7ce48b6c730a9f0b22219d3. The migration guide now states that all seven blocking gates span functional, diagnostics, data, parity, accuracy, and packaging categories.

docs/power/qualification-matrix.json
```diff
@@ -0,0 +1,299 @@
...
+          "thresholds": {"finite_nonnegative_values": true, "power_and_limit_columns_paired": true}
+        }
+      ],
+      "execution": {
+        "kind": "automated",
+        "status": "pending",
+        "command": "python -m pytest -q python/aisimulate/tests/unit/sdk/database/test_power_data_invariants.py",
+        "command_status": "available",
+        "evidence": [],
+        "blocked_by": []
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in 4771d092b897ce8ff7ce48b6c730a9f0b22219d3. I reran the available invariant command at exact head def18eb3, preserved the complete 15-file failure summary in a checked-in SHA-256-bound artifact, and moved the gate to failed with a matching non-passing evidence record. The qualification remains fail-closed.

docs/power/qualification-matrix.schema.json
```diff
@@ -0,0 +1,213 @@
...
+      "type": "object",
+      "additionalProperties": false,
+      "required": ["kind", "status", "command", "command_status", "evidence", "blocked_by"],
+      "properties": {
+        "kind": {"enum": ["automated", "manual"]},
+        "status": {"enum": ["pending", "blocked", "passed", "failed"]},
+        "command": {"type": ["string", "null"]},
+        "command_status": {"enum": ["available", "planned", "not_applicable"]},
+        "evidence": {"type": "array", "items": {"$ref": "#/$defs/evidence"}},
+        "blocked_by": {"type": "array", "items": {"type": "string", "minLength": 1}}
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in 4771d092b897ce8ff7ce48b6c730a9f0b22219d3. The interchange schema now rejects passed gates that retain a planned command, have no evidence, or contain non-passing evidence; schema-only regressions cover planned and evidence-free passed states.

scripts/validate_power_qualification.py
```diff
@@ -0,0 +1,936 @@
...
+        errors,
+    )
+    if isinstance(artifact, str) and artifact:
+        path_parts = Path(artifact).parts
+        _expect(
+            artifact.startswith("https://")
+            or (not Path(artifact).is_absolute() and ".." not in path_parts),
+            f"{context}.artifact must be HTTPS or a repository-relative path",
+            errors,
+        )
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in 4771d092b897ce8ff7ce48b6c730a9f0b22219d3. Artifact values containing a URL scheme are now accepted only for HTTPS, with a regression proving HTTP evidence is rejected.

scripts/validate_power_qualification.py
```diff
@@ -0,0 +1,936 @@
...
+        raise QualificationError("\n".join(f"- {error}" for error in errors))
+
+
+def load_and_validate(
+    path: Path,
+    *,
+    require_release_ready: bool = False,
+    expected_revision: str | None = None,
+) -> dict[str, Any]:
+    document = json.loads(path.read_text(encoding="utf-8"))
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in 4771d092b897ce8ff7ce48b6c730a9f0b22219d3. Ledger loading now uses a named parse_constant hook that raises QualificationError for NaN and Infinity; a temporary-ledger regression exercises the strict parser.

tests/test_power_qualification.py
```diff
@@ -0,0 +1,354 @@
...
+            gate["execution"]["command_status"] = "available"
+        gate["execution"]["evidence"] = [_passing_evidence(revision, gate["id"])]
+    return document
+
+
+def test_checked_in_power_qualification_ledger_is_structurally_valid() -> None:
+    document = QUALIFICATION.load_and_validate(LEDGER)
+
+    assert document["release_state"] == "not_qualified"
+    assert QUALIFICATION.main([]) == 0
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in 4771d092b897ce8ff7ce48b6c730a9f0b22219d3. The test suite now exercises main() end to end: the checked-in ledger fails release-ready mode, a qualified temporary ledger succeeds with a matching expected revision, and a mismatched revision fails with the expected diagnostic.

docs/power/qualification-matrix.json
```diff
@@ -0,0 +1,307 @@
...
+        {
+          "when": {},
+          "expectation": "data_invariants",
+          "thresholds": {"finite_nonnegative_values": true, "power_and_limit_columns_paired": true}
+        }
+      ],
+      "execution": {
+        "kind": "automated",
+        "status": "failed",
+        "command": "python -m pytest -q python/aisimulate/tests/unit/sdk/database/test_power_data_invariants.py",
```
jasonqinzhou: [P1] Bind the invariant command to this repository's Python project

The recorded `python -m pytest` command is environment-dependent. Reproducing it from this worktree imported `aiconfigurator_core` from another checkout (`/Users/jasonzho/codex/agent-sandbox/aic-pr-review-work/review-slots/slot-1/aic-core`), which produced the 21-file/40,549-row failure now stored as evidence. Running through this repository's uv project imports the package from this branch, scans the three packaged power-carrying files, and passes. The executable gate must use a repository-bound command, and its evidence/status must be regenerated from that exact source revision rather than preserving the unrelated-checkout result.

docs/power/evidence/power-data-invariants-def18eb3.txt
```diff
@@ -0,0 +1,30 @@
...
+- b200_sxm/linear_attention/trtllm/1.3.0rc20/mamba2_perf.parquet: 1 row
+- b200_sxm/mla/trtllm/1.3.0rc20/context_mla_perf.parquet: 1 row
+- b200_sxm/mla/trtllm/1.3.0rc20/generation_mla_perf.parquet: 2 rows
+- b200_sxm/mla/trtllm/1.3.0rc20/mla_context_module_perf.parquet: 18 rows
+- b200_sxm/mla_bmm/trtllm/1.3.0rc20/mla_bmm_perf.parquet: 23 rows
+- b200_sxm/moe/trtllm/1.3.0rc20/moe_perf.parquet: 39262 rows
+- b200_sxm/quantize/trtllm/1.3.0rc20/computescale_perf.parquet: 5 rows
+- b200_sxm/quantize/trtllm/1.3.0rc20/scale_matrix_perf.parquet: 3 rows
+- b200_sxm/sparse_attention/trtllm/1.3.0rc20/dsa_context_module_perf.parquet: 15 rows
+- b200_sxm/sparse_attention/trtllm/1.3.0rc20/dsa_generation_module_perf.parquet: 93 rows
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in a8f8f0dda966d322a4ba782dca7136fd0b2546f6. The invalid ambient-Python failure artifact was replaced with repository-bound passing evidence that records the 3 files and 122,362 rows scanned, units, key columns, paired-column result, and separate NaN/negative/non-positive counts in a machine-readable JSON block. A qualification regression parses and checks the evidence totals and both asserted invariants.

docs/power/qualification-matrix.json
jasonqinzhou: Fixed across bb31ab48cbe1e519683ce27bf7bc82f0e090ed67 and a8f8f0dda966d322a4ba782dca7136fd0b2546f6. The gate now invokes pytest through `uv run --project python/aisimulate --extra dev`, the invariant test enforces paired columns, and the evidence/status were regenerated against the repository-bound source revision. That exact command passes (1 test), scanning the three packaged power-carrying files.

jasonqinzhou:
@coderabbitai review

kaim-eng:
## Review finding: Power qualification evidence artifacts are not reproducible from the recorded command

**Severity:** High — the qualification ledger asserts evidence that cannot have
been produced by the process it documents.
**Found while verifying:** PR #141 review (the review flagged reproducibility
abstractly; the concrete proof below was missed).

## Problem 1 — the deleted FAIL transcript could not have come from this repo

The FAIL transcript removed by PR #141
(`docs/power/evidence/power-data-invariants-def18eb3.txt`) lists 15
`trtllm/1.3.0rc20` parquet files as having NaN or non-positive `power_limit`
values.

Two facts contradict that:

- `git diff def18eb3 bb31ab4 -- python/aisimulate/src/aiconfigurator_core/systems/data`
  is **empty** — the data did not change between the FAIL and PASS revisions.
- None of those 15 files contain a `power_limit` column at all.

So the recorded failure cannot have been generated from this repository's data.
These transcripts appear to be hand-authored against some other environment,
which undermines every evidence-backed assertion in
`docs/power/qualification-matrix.json`.

## Problem 2 — the invariant test binds to the installed package, not the repo

```python
# python/aisimulate/tests/unit/sdk/database/test_power_data_invariants.py:32
_DATA_ROOT = Path(aiconfigurator_core.__file__).parent / "systems" / "data"
```

Whatever `aiconfigurator_core` happens to be importable supplies the scanned
data. This directly contradicts the evidence file's "repository-bound scan"
wording and the ledger note "The repository-bound invariant test passes" — the
gate's meaning is environment-dependent. This is also the root cause of
Problem 1: a differently-installed package would scan different files.

## Problem 3 — the transcript cannot be produced by the command it records

`docs/power/evidence/power-data-invariants-bb31ab4.txt:9` records the producing
command as a `pytest -q` invocation. The underlying test only accumulates
`problems` strings and asserts; on success it prints nothing. None of the
per-file schemas, row counts, units, or the embedded JSON block can come from
that command's output.

Relatedly, `python/aisimulate/tests/test_power_qualification.py:82-110` compares
the hand-written JSON block against equally hand-written constants — it never
opens a parquet file, so it cannot detect divergence from reality.

## What is actually true

Independent verification did confirm the **file list**: grepping all 762
`*_perf.parquet` blobs for a literal `power_limit` column yields exactly the
three `b200_sxm/vllm/0.22.0` files named in the evidence. (Other `moe_perf`
hits for "power" are the `power_law_1.2` distribution value, not a column.) The
`122362` row count remains unverified.

## Suggested fix

1. Generate evidence artifacts from a script that actually emits the recorded
   content, and have CI regenerate and diff them rather than trusting a
   checked-in transcript.
2. Point `_DATA_ROOT` at the repository tree explicitly, or assert that the
   importable package resolves inside the repo.
3. Cross-check the artifact's `Source revision:` header against
   `evidence["source_revision"]` in the ledger — nothing does this today.
4. Drop or regenerate the ledger's `uv run --project ... --extra dev` command,
   which is asserted `command_status: "available"` but is not run by any
   workflow (no workflow uses `uv`).

## Related, lower priority

- `test_power_data_invariants.py:55-62` gates on `isna() | (power < 0)`. Since
  pandas' `use_inf_as_na` defaults to False, `+inf` passes both predicates while
  the ledger asserts `finite_nonnegative_values: true`. Latent coverage hole; no
  shipped file currently contains `+inf`.
- `tests/test_power_qualification.py:77-78` asserts a raw-byte SHA-256 over a
  text artifact with no `.gitattributes` coverage. Reproduced on Windows: a
  fresh worktree checkout materializes CRLF (129 CR-bearing lines), giving
  working-tree `5ad8dc5c…` against blob/ledger `e643168b…`. Developer-only (no
  Windows runners), but the assertion fails on a clean checkout.
- `tests/test_power_qualification.py:85` uses `text.split("```json\\n", 1)[1]`,
  raising `IndexError` instead of a readable assertion if the fence is absent.

jasonqinzhou:
Self-review after merging current `main`: Ruff 0.14.1 reports formatting drift in `tests/test_power_qualification.py` (now included in Fast CI) and the qualification validator. I am applying formatter-only changes; no qualification behavior or evidence is changing.

jasonqinzhou:
Self-review follow-up while fixing the new validation comments: the release-ready path has the same malformed-nested-object risk for `policy.silicon_accuracy` that CodeRabbit identified for gate `execution`. I will normalize that value to a checked mapping too and add a regression, so malformed ledgers consistently return accumulated `QualificationError` diagnostics instead of an `AttributeError` traceback.

tests/test_power_qualification.py
```diff
@@ -0,0 +1,431 @@
...
+    assert artifact.is_file()
+    assert hashlib.sha256(artifact.read_bytes()).hexdigest() == evidence["sha256"]
+    assert QUALIFICATION.main([]) == 0
+
+
+def test_power_data_invariant_evidence_is_complete_and_machine_readable() -> None:
+    document = _document()
+    evidence = _gate(document, "power-data-invariants")["execution"]["evidence"][0]
+    text = (ROOT / evidence["artifact"]).read_text(encoding="utf-8")
+    result = json.loads(text.split("```json\n", 1)[1].split("\n```", 1)[0])
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in `92851629950b5a341081e7ff397b0856a2b76102`. The invariant result now has a dedicated raw JSON sidecar, the ledger records that sidecar and its exact SHA-256, and the test loads it deterministically with `json.load` rather than parsing a Markdown fence.

scripts/validate_power_qualification.py
```diff
@@ -0,0 +1,877 @@
...
+    if isinstance(artifact, str) and artifact:
+        path_parts = Path(artifact).parts
+        _expect(
+            artifact.startswith("https://")
+            or ("://" not in artifact and not Path(artifact).is_absolute() and ".." not in path_parts),
+            f"{context}.artifact must be HTTPS or a repository-relative path",
+            errors,
+        )
+    _expect(
+        isinstance(evidence.get("sha256"), str) and bool(DIGEST_RE.fullmatch(evidence["sha256"])),
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in `92851629950b5a341081e7ff397b0856a2b76102`. Release-ready validation now resolves repository-relative artifacts from the repository root, streams HTTPS artifacts with a bounded timeout, rejects non-HTTPS redirects/unavailable content, computes SHA-256 from the actual bytes, and fails on digest mismatch. Tests cover local match/mismatch/missing artifacts and a mocked HTTPS fetch.

jasonqinzhou:
Resolved the complete fresh CodeRabbit review in `92851629950b5a341081e7ff397b0856a2b76102`:\n\n- release-ready validation now verifies actual local/HTTPS artifact bytes against each SHA-256\n- the checked-in power-data result is a deterministic raw JSON sidecar with a ledger-bound digest\n- malformed gate execution and silicon-policy objects accumulate validation errors instead of throwing attribute errors\n- additional gate thresholds reject NaN and infinities while preserving valid booleans/numbers\n- the AFD migration example explicitly says to save the YAML as `recommendation.yaml`\n\nValidation: 177 qualification/source/CODEOWNERS policy tests passed; the packaged power-data invariant passed; normal ledger validation succeeds and release-ready mode still fails closed by design. Both inline threads were replied to and resolved.

docs/power/evidence/power-data-invariants-bb31ab4.json
```diff
@@ -0,0 +1,132 @@
...
+      "result": "pass",
+      "power": {
+        "unit": "W",
+        "nan_count": 0,
+        "negative_count": 0
+      },
+      "power_limit": {
+        "unit": "W",
+        "nan_count": 0,
+        "non_positive_count": 0
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in 05edf925. The gate-specific report now records and the validator enforces positive_infinity_count, negative_infinity_count, the discovery root, all 762 discovered parquet files, the 759 files without power columns, and skipped-file anomalies. The evidence producer also rejects both infinities. The refreshed scan still passes: 3 power-carrying files and 122,362 rows.

scripts/validate_power_qualification.py
```diff
@@ -0,0 +1,964 @@
...
+    return False
+
+
+def _artifact_sha256(artifact: str) -> str:
+    digest = hashlib.sha256()
+    if artifact.startswith("https://"):
+        with urlopen(artifact, timeout=ARTIFACT_FETCH_TIMEOUT_SECONDS) as response:
+            final_url = response.geturl()
+            if not final_url.startswith("https://"):
+                raise ValueError(f"artifact redirected to a non-HTTPS URL: {final_url}")
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in 05edf925 by removing remote artifact fetching from this release gate. Evidence artifacts must now be repository-relative; the JSON Schema, validator, docs, and a no-network regression test enforce that boundary. A caller-supplied ledger can no longer cause any outbound request, including redirects or internal-host requests.

scripts/validate_power_qualification.py
```diff
@@ -0,0 +1,964 @@
...
+            while chunk := response.read(ARTIFACT_READ_CHUNK_BYTES):
+                digest.update(chunk)
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in 05edf925. Remote downloads were removed entirely, and repository-relative artifacts now have a 16 MiB pre-read size check plus a cumulative read bound to handle file growth during validation. The oversized-artifact negative test passes.

scripts/validate_power_qualification.py
```diff
@@ -0,0 +1,964 @@
...
+        )
+
+
+def _validate_evidence(
+    evidence: Any,
+    context: str,
+    errors: list[str],
+    *,
+    verify_artifact: bool,
+) -> None:
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in 05edf925. Every evidence artifact is now a versioned gate-specific JSON report, and validation binds its gate ID, source revision, exact matrix, every assertion and result, units, anomaly list, and non-empty details to the owning gate. Passing reports require all assertion results to pass and no anomalies. Regression tests independently corrupt each binding and confirm fail-closed behavior.

jasonqinzhou:
Resolved the four latest CodeRabbit findings in exact head 05edf925f1f4d7fe983330cc600d61f6957dd146. Evidence reports are now gate-bound and content-validated; the packaged data report includes full discovery plus NaN and both infinity counts; remote URLs are rejected before any network access; local reads are capped at 16 MiB. Local exact-head evidence: 187 qualification/source/CODEOWNERS tests passed, the packaged power-data invariant passed (3 files, 122,362 rows), Ruff and diff checks passed, and the worktree is clean. Automated exact-head checks are running; human approval and Full CI remain separate gates.

jasonqinzhou:
Final exact head 667cdb7d82842279f119204aceb7e5cba3cc438e is green. The follow-up commits only corrected repository-configured Ruff lint/format findings. Exact-head local evidence remains 187 qualification/source/CODEOWNERS tests passed plus the packaged power-data invariant passed; Fast CI, Python static, repository policy, Rust format, CODEOWNERS, DCO, and CodeRabbit all pass. There are no unresolved technical threads. Human CODEOWNER approval and Full CI remain separate pre-merge gates.

kaim-eng:
### Review at `667cdb7d`: reproducibility of the power-data-invariants evidence

I checked the outstanding reproducibility concern against the tree at `667cdb7d`. All four points hold. None is a false positive, though one of them is a general property of checked-in ledgers rather than a defect introduced here.

**1. The invariant test scans whichever `aiconfigurator_core` is importable — confirmed, and live in CI.**

`python/aisimulate/tests/unit/sdk/database/test_power_data_invariants.py:26` resolves the scan root as `Path(aiconfigurator_core.__file__).parent / "systems" / "data"`. Under the recorded command (`uv run --project python/aisimulate ...`) that lands on the checkout, because maturin installs the project editable with `python-source = "src"`. But `.github/workflows/ci.yml` installs the package **non-editable** (`python -m pip install './python/aisimulate[dev]'`) before running the unit suite, so in CI this test scans the installed wheel's data tree, not the working copy. The evidence file records `discovery_root: python/aisimulate/src/aiconfigurator_core/systems/data`, which is not the root CI actually walks. Worth confirming separately whether the wheel's `include` globs package all 762 parquet files a repo-root scan discovers — if they don't, the recorded discovery count and the CI scan are measuring different populations.

**2. The recorded command does not generate the JSON evidence — confirmed.**

The test only asserts; it prints nothing on success, and `pytest -q` produces exactly the "1 passed in 0.63s" recorded at the bottom of the `.txt`. The evidence, however, reports `total_discovered_parquet_count`, `files_without_power_columns`, per-file `key_columns`, and *separate* `positive_infinity_count` / `negative_infinity_count`. The test cannot produce the last of those even in principle — it folds `isna() | == inf | == -inf | < 0` into one aggregate count. So the JSON was produced by a scanner that is not in this PR. That collides with the policy this PR itself sets in `docs/power/README.md:90`: "Evidence should be public-safe and reproducible."

**3. The qualification test validates hand-authored values — confirmed.**

`tests/test_power_qualification.py:123-173` asserts `762`, `3`, `759`, `122362` and the zero-counts by reading the checked-in JSON. There is no `pyarrow` import in that file and no parquet is opened. It verifies the ledger's internal consistency, not the data the ledger describes. Secondary cost: those constants break on any power-data change, in a test whose failure message will not explain why.

**4. A self-consistent fabricated report would satisfy `--require-release-ready` — confirmed, with a caveat.**

`scripts/validate_power_qualification.py` is a document linter. `_validate_data_invariant_details` (line 141) checks types, `files_scanned == len(files)`, `total_discovered == files_scanned + without_power`, and empty anomalies. `_validate_evidence` (line 579) hashes the artifact the ledger points at — self-referential, since ledger and artifact land in the same commit. Nothing re-executes a gate or reads a parquet.

The caveat: this is structural to any human-maintained ledger, and the PR never claims otherwise — the checked-in state is deliberately `not_qualified`, and the digest/revision binding does defend against the distinct failure of mixing artifacts across revisions. As "the validator cannot distinguish a run from a well-formed transcript of one," it's fair; as a bug, it overreaches.

### Suggested resolution

One change collapses all four:

1. Commit the scanner that produced the JSON as `scripts/`-level code, with the report schema it already emits.
2. Have it assert its resolved discovery root is inside the repository (closes 1), and record that root as it resolved it.
3. Make the gate's recorded command that script (closes 2).
4. Have `tests/test_power_qualification.py` run the scanner and diff against the checked-in artifact rather than asserting literals (closes 3, and makes 4 detectable by re-run).
5. Run it in CI so the artifact drifting from the data fails the build.

Separately worth deciding: whether the invariant test should bind to the repository tree or continue to follow the installed package. Both are defensible — the installed package is what ships — but the evidence should then say which one it measured.

Everything else in the ledger, schema, and validator looks sound to me; this is the one gap between the artifact and the honesty standard the PR sets for itself.

kaim-eng (changes requested):
The qualification-evidence reproducibility issue remains at the current head. The validator checks report self-consistency and hashes, but does not establish that the evidence came from executing the recorded gate: the pytest command does not generate the sidecar, the qualification test reloads hard-coded report values rather than rescanning the checkout's parquet data, and the invariant scan root can resolve from whichever `aiconfigurator_core` is importable. Please bind each automated gate report to actual execution against the checked-out tree—preferably CI-generated evidence or an immutable CI attestation—and verify or regenerate it in CI.

jasonqinzhou:
Addressed the September 12 review at `1f1b36cadf3c570cd2fb0c9a9539501e34bef6a9`. The data gate now generates its report by scanning the checked-out data tree. Fast CI executes the producer, re-scans and compares the complete report, and uploads the candidate ledger/report. Release mode rejects unimplemented automated verifiers, stale or dirty sources, and fabricated evidence. 264 focused checks and an actual clean-head generation/verification run pass.

The review decision still needs re-review. Applicable CODEOWNER approval and successful exact-head Full CI remain separate merge gates.

jasonqinzhou:
Continued in #213 using an upstream branch at the same commit. GitHub rejects reopening this PR after it closed during a base update. The replacement targets #208 and preserves this discussion as review history; previous approvals and checks do not qualify the replacement.

PR #143, opened 2026-09-08
feat(perfmodel): enable empirical database mode in native forward-pass models

simone-chen:
## Why and what changed

Native forward-pass construction now preserves database mode, shared-layer, transfer-policy, and strict-provenance settings across `EngineConfig`, the Rust builder, Python compilation, the serialized spec, and Rust reload.

Formula-only `EMPIRICAL` and `SOL` engines may load without a perf-data directory. Explicit policy errors remain fail-closed. If Python discovery returns no view, the requested policy is still serialized so native reload produces the authoritative input error instead of silently reverting to defaults. `SOL_FULL` remains a per-call diagnostic and is rejected as an engine-wide default at builder, deserialization, and common engine-build boundaries.

Version-slot validation is intentionally unchanged. Raw vLLM 0.25.1 still requires the existing explicit allow-unlisted-version escape.

## Review map

- Risk level: medium
- Start with: `python/aisimulate/src/aiconfigurator_core/sdk/engine.py` for policy loading and serialization; `crates/core/src/perfmodel/py.rs` for Rust-to-Python propagation; `crates/core/src/perfmodel/engine/runtime.rs` for native reload and engine-default validation.
- Public or serialized contract changed: yes; `compile_engine` gains four defaulted keyword-only policy arguments, `AicEngineBuilder` gains matching controls, and `DatabaseMode` is publicly exported. Existing `EngineConfig` wire fields are now preserved by native construction.
- Compatibility or rollback concern: existing callers retain their defaults. Dynamo's current callback remains default-only. When Dynamo exposes these knobs, its builder calls and compiled-engine cache key must be extended together; verified against Dynamo `main` at `53ced98cd1e3ed30297b50e53298f91227b71fa0`. Rollback requires reverting the feature and its follow-up fixes together.

## Evidence

- Exact reviewed head: `e664f619b1892350f33b7f4ee4e7d119280145a8`.
- Local core on the exact head: 1,160 passed, 1 ignored; all integration targets passed (7, 30, 1, 2, and 15 tests).
- Focused Python on the exact head: 68 passed across compile-policy propagation, literal-version handling, native-step fallback, and public engine behavior.
- The exact malformed-system-YAML regression that failed the prior Full CI run now passes locally while the requested database policy remains serialized without a Python view.
- Fast CI: passed on the exact head ([run 34318764419](https://github.com/ai-dynamo/aisimulate/actions/runs/34318764419)).
- Full CI: passed on the exact head across amd64 and arm64 ([run 34319205272](https://github.com/ai-dynamo/aisimulate/actions/runs/34319205272)); `Full CI Success` passed.
- CodeRabbit reviewed commit: exact head, with no unresolved review threads. Its downstream Dynamo cache warning was verified as not currently reachable and dispositioned in the PR discussion.
- Codex reviewed commit: exact head; no unresolved P0/P1 findings.
- Negative or boundary cases: explicit policy exception; no Python view with explicit wire policy; legacy all-default tolerant load; malformed system YAML diagnostic; estimate-only empirical reload; `SOL_FULL` rejection in the public builder, spec reload, and common engine build; spec/database mode mismatch; stable Python signature contract.
- Expected-value derivation: N/A; no numerical formula or golden value changed.

No performance-data artifact or golden changed. The original feature-head validation reported 286/286 predictions with zero failures for the MiniMax-M2.7/H200/vLLM 0.25.1/TEP4 slice after enabling empirical plus shared calibration. On that original head (`3100b7c1be3f10036d72ddc2e35af619449c616b`), silicon MAPE was 27.87% and empirical MAPE was 28.19%. This demonstrates reachability only; it does not claim empirical accuracy is better, and silicon remains the preferred default when available.

## Tracking

- Closes: N/A
- Related PRs or issues: N/A


python/aisimulate/src/aiconfigurator_core/sdk/engine.py
```diff
@@ -741,12 +754,34 @@ def _evaluate_single_op(
...
+            backend_version,
+            systems_paths=systems_path,
+            allow_missing_data=formula_only,
+            database_mode=database_mode,
+            shared_layer=shared_layer,
+            transfer_policy=transfer_policy,
+            strict_provenance=strict_provenance,
+        )
     except Exception:
+        if database_mode is not None:
```
jasonqinzhou: **[P1] Keep every explicit database policy fail-closed**

This rethrows only when `database_mode` is set. `AicEngineBuilder::strict_provenance(true)` can leave that field unset; if strict validation raises, this catches the error, returns `None`, and `_engine_config_dict` then serializes `strict_provenance=false`, allowing the native engine to build non-strict. Explicit `shared_layer` and `transfer_policy` settings can be lost similarly. Please re-raise whenever any database-policy argument is explicit and add a regression for strict provenance with the default mode.

python/aisimulate/src/aiconfigurator_core/sdk/engine.py
```diff
@@ -741,12 +754,34 @@ def _evaluate_single_op(
...
+    database_mode: str | None,
+    shared_layer: bool | None,
+    transfer_policy: str | list[str] | None,
+    strict_provenance: bool | None,
+) -> Any:
     try:
         from aiconfigurator_core.sdk import perf_database
 
-        return perf_database.get_database(system, backend, backend_version, systems_paths=systems_path)
+        formula_only = database_mode is not None and database_mode.upper() in {"EMPIRICAL", "SOL"}
```
jasonqinzhou, line 799: **[P1] Carry formula-only EMPIRICAL tolerance through the Rust reload**

This correctly requests `allow_missing_data=True` for `EMPIRICAL`, but the resulting spec does not mark estimate-only loads as tolerant, and `Engine::from_spec_bytes` currently skips the missing-directory gate only for `DatabaseMode::Sol` or directory-less `next`. Python compilation therefore succeeds and the Rust half fails for an empirical view with no matching data directory—the scenario this path is intended to support. Please extend the Rust tolerance (or carry an explicit wire flag) and add an end-to-end regression.

crates/core/src/perfmodel/common/enums.rs
```diff
@@ -67,6 +66,19 @@ impl Default for DatabaseMode {
...
 
+impl DatabaseMode {
+    /// Python `common.DatabaseMode` member name used by `compile_engine`.
+    pub(crate) fn as_str(self) -> &'static str {
+        match self {
+            Self::Silicon => "SILICON",
+            Self::Hybrid => "HYBRID",
+            Self::Empirical => "EMPIRICAL",
+            Self::Sol => "SOL",
+            Self::SolFull => "SOL_FULL",
```
jasonqinzhou, line 79: **[P2] Do not expose SOL_FULL as a selectable database default**

The new public `DatabaseMode` makes `SolFull` discoverable through `AicEngineBuilder::database_mode`, but Python deliberately rejects `SOL_FULL` as a database default because it is only a per-call diagnostic. The direct builder therefore fails unexpectedly, and `best_available` may classify that configuration error as an unsupported model and fall back to regression. Please exclude this variant from the builder-facing API or reject it as `InvalidEngineConfig` before entering Python, with a public-API regression.

python/aisimulate/src/aiconfigurator_core/sdk/engine.py
jasonqinzhou: Fixed in 7b901376: the database-loading helper now treats every explicitly supplied policy field as fail-closed, including strict provenance, shared layer, and transfer policy. The parametrized regression test covers each field while preserving the legacy tolerant all-default path.

python/aisimulate/src/aiconfigurator_core/sdk/engine.py
jasonqinzhou: Fixed in 7b901376: native deserialization now permits a missing perf-data directory for formula-only EMPIRICAL as well as SOL, with an end-to-end from-spec-bytes regression test using the estimate-only H100 fixture.

crates/core/src/perfmodel/common/enums.rs
jasonqinzhou: Fixed in 7b901376: SOL_FULL is documented as per-call-only and is now rejected before Python compilation by the public builder, and again during direct EngineSpec deserialization. Both entry points have regression coverage.

python/aisimulate/src/aiconfigurator_core/sdk/engine.py
```diff
@@ -741,12 +754,37 @@ def _evaluate_single_op(
...
+) -> Any:
+    explicit_policy = any(
+        value is not None for value in (database_mode, shared_layer, transfer_policy, strict_provenance)
+    )
     try:
         from aiconfigurator_core.sdk import perf_database
 
-        return perf_database.get_database(system, backend, backend_version, systems_paths=systems_path)
+        formula_only = database_mode is not None and database_mode.upper() in {"EMPIRICAL", "SOL"}
+        return perf_database.get_database_view(
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in 78132006: the helper now raises a clear configuration error when an explicitly requested policy produces no database view, while preserving the legacy None result for the all-default path. Added a regression for the non-exception None return.

crates/core/src/perfmodel/engine/runtime.rs
```diff
@@ -1914,6 +1929,53 @@ mod tests {
...
+
+        assert!(matches!(
+            result,
+            Err(AicError::InvalidEngineConfig(message))
+                if message.contains("SOL_FULL") && message.contains("per-call diagnostic")
+        ));
+    }
+
+    #[test]
+    fn build_rejects_sol_full_as_database_default() {
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in cb2f1e53: the direct Engine build test now keeps the spec at its allowed SILICON default and supplies a SOL_FULL database view, so it reaches and verifies the loaded-database guard.

jasonqinzhou:
Addressed the outside-diff CodeRabbit finding in accd287a: Engine::build now rejects any mismatch between the spec database mode and the loaded database mode, with a negative regression covering SILICON spec versus EMPIRICAL database.

python/aisimulate/src/aiconfigurator_core/sdk/engine.py
```diff
@@ -741,12 +754,40 @@ def _evaluate_single_op(
...
+            backend_version,
+            systems_paths=systems_path,
+            allow_missing_data=formula_only,
+            database_mode=database_mode,
+            shared_layer=shared_layer,
+            transfer_policy=transfer_policy,
+            strict_provenance=strict_provenance,
+        )
+        if database is None and explicit_policy:
+            raise ValueError("performance database unavailable for explicit database policy")
```
jasonqinzhou: [P1] Preserve native input diagnostics without dropping the requested database policy

Full CI fails this path on both amd64 and arm64. When get_database_view returns None for a malformed system YAML, this generic error is raised before the native loader can report the underlying YAML error. Please keep the explicit policy on the serialized EngineConfig even without a Python view, so native loading can retain the specific diagnostic while still failing closed instead of silently reverting policy.

jasonqinzhou: Fixed in e664f619. Explicit database mode, shared-layer, transfer, and provenance settings are now serialized independently of the optional Python view. A missing view therefore reaches native reload with its requested policy intact, preserving the underlying YAML diagnostic and preventing silent policy fallback. Added unit coverage for the no-view wire config; the failed integration regression and the related 68-test SDK suite pass locally.

jasonqinzhou:
Disposition of the CodeRabbit cross-repository cache warning at e664f619: verified against ai-dynamo/dynamo main 53ced98cd1e3ed30297b50e53298f91227b71fa0. The current build_rust_engine and create_aic_callback signatures do not accept database mode, shared-layer, transfer-policy, or strict-provenance inputs, and the callback never invokes the new builder methods. Therefore every currently reachable Dynamo cache entry uses the same default database policy; two policy-distinct callback configurations cannot yet reach this cache. When Dynamo adopts these policy knobs, its builder calls and cache identity must be extended together with a cache-reuse regression. That downstream adoption work is not a correctness blocker for this AISimulate API-enablement PR.

jasonqinzhou (approved):
Approved at e664f619b1892350f33b7f4ee4e7d119280145a8. Reviewed database-policy propagation through the public builder, Python compiler, serialized EngineConfig, and native reload; validated fail-closed policy and SOL_FULL boundaries plus malformed-system diagnostics. Fast CI, CodeRabbit, exact-head Full CI across amd64/arm64, and the local core suite are green. The Dynamo cache-key warning is not reachable in its current default-only callback and is documented as required follow-up when those knobs are adopted.

PR #161, opened 2026-09-10
feat(replay): add aggregated steppable replay (external DES driver)

ajcasagrande:
## Why and what changed

Adds the Rust replay seam for external discrete-event drivers. `SteppableAgg` owns aggregated workers while the caller controls clock advancement, dynamic submission, cancellation, and report collection. `SteppableEngine` provides the single-worker wrapper.

Terminal events hold their timestamp before the next pass is committed, allowing same-time replacement work. Nonterminal tokens can commit the next pass before returning; drivers requiring `run()` batch composition must submit known same-timestamp arrivals before stepping. Cancellation is immediately terminal, and final reports require an idle runtime.

Review fixes make failed placement leave no retained arrival, retire ownership when a pending final output is canceled, preserve capture settings between report epochs, and calculate each epoch through its simulated report boundary, including trailing idle time. Explicit UUIDs may be reused after a successful report drain.

## Review map

- Risk level: high; scheduling, cancellation ownership, and public report semantics.
- Start with `crates/core/src/replay/loadgen/steppable.rs` (public driver contract and boundary tests), `agg.rs` (admission and delta-cycle implementation), `engine/scheduler/rank.rs` (pending-output retirement), and `replay/report.rs` (report boundaries and rates).
- Public or serialized contract changed: new Rust `SteppableReplay`, `SteppableAgg`, `SteppableEngine`, `EngineEvent`, and `StepOutcome` exports. No Python/CLI binding or serialized event format is introduced by this PR.
- Compatibility or rollback concern: the ordinary consuming collector finalization keeps its existing terminal-based duration. Reusable reports cover each report interval, while per-request timestamps remain absolute. Idle gaps require explicit `advance_now_ms`; same-time nonterminal ordering is documented above. This seam currently supports aggregated topology.

```mermaid
flowchart LR
    D[External discrete-event driver] -->|submit / cancel| S[SteppableAgg]
    D -->|step_until / next_event_ms| S
    S --> A[Aggregated native engine]
    A -->|tokens + terminal| S
    S -->|EngineEvent / ReplayReport| D
```

---

### DES event-loop view

exec-1375680a-90ef-4d02-a845-15eae0cd58a5

## Evidence

Current reviewed head: `3330ff54431dd15f0f6f0965e9a5c79243568670`, including current `main` at `da882665e690c94d67df096bfd190cc3f48e00c8`.

- Local tests: `cargo test --workspace --jobs 4` passed: 1,197 library tests and 55 integration tests, zero failures. One existing R17-pair test is ignored because it requires `AIC_FPM_R17_PAIR_DIR`.
- Formatting: `cargo fmt --all -- --check` passed; `git diff --check` passed.
- Fast CI: [passed on 3330ff54](https://github.com/ai-dynamo/aisimulate/actions/runs/34531267856). DCO and CODEOWNERS also passed on that head.
- Full CI: not yet run on 3330ff54. [Copy-PR vetting is required](https://github.com/ai-dynamo/aisimulate/pull/161#issuecomment-5625556301) before NVIDIA runner workflows can start. The [previous Full CI](https://github.com/ai-dynamo/aisimulate/actions/runs/34523253152) passed on 4fa2309a and is historical evidence only.
- CodeRabbit reviewed commit: completed review of 4fa2309a; incremental review of 3330ff54 is in progress.
- Codex reviewed commit: 3330ff54; reviewed the complete PR diff, relevant runtime accounting, report finalization, existing feedback, and corrective tests. All ten existing threads were addressed and resolved.
- Negative or boundary cases: failed preferred-rank placement and malformed placement identity, cancellation after a final pass starts, duplicate live/retained UUIDs, UUID reuse after reporting, non-finite report time, reporting while busy, bounded deadlines, tokenless and cascade terminal holds, multi-epoch capture, trailing idle time, empty epochs, and worker shrink/re-add/expansion all pass.
- Expected-value derivation: epoch duration is report boundary minus preceding boundary; request rate and goodput are completed/eligible requests divided by elapsed seconds; provisioned worker-seconds are worker count times elapsed seconds. Request terminal timestamps remain absolute. Round-robin expectations follow the sorted active worker set and independent per-worker rank rotation.
- Before/after evidence: four new regression cases failed before the fixes (malformed-placement retention, trailing-idle duration, empty-epoch duration, and second-epoch capture) and pass afterward. Seven new tests cover these cases plus UUID reuse and worker/rank rotation.
- Prior one-off parity evidence: [5,000-request Mooncake campaign and reproduced 4fa2309a results](https://github.com/ai-dynamo/aisimulate/pull/161#issuecomment-5624680213). This was requested as one-off evidence rather than a permanent fixture/CI requirement. It has not been rerun on 3330ff54. No production performance claim is made.

The added regression cases use synthetic requests and the existing default engine configuration; simulation timestamps are milliseconds, rates are per second, and provisioned usage is worker-seconds. No measured performance dataset or timing model is added. The linked historical parity campaign records its pinned Mooncake source, checksums, configuration, repeated-process comparisons, and coverage limits.

## Tracking

- Closes: none.
- Related PRs or issues: review discussions and parity evidence on this PR.


ajcasagrande:
@PeaBrane

PeaBrane (commented):
Could we run a one-off Mooncake parity check for this PR, comparing ordinary aggregated `run()` with `SteppableAgg`? The existing tests cover individual API behaviors; a differential comparison would help establish that external clock control preserves admission, batching, and request timing across a realistic workload.

This is intended as one-off validation evidence, not a permanent checked-in fixture, regression suite, or CI requirement. A temporary harness and a summary of the configuration and results on this PR would be sufficient.

A fixed, checksummed slice—such as the 5,000-request window used by the Dynamo replay parity workflow—could exercise two modes:

- **Trace arrivals:** preserve the original timestamps, with the external driver submitting requests through bounded stepping.
- **Fixed concurrency:** preserve request order and refill slots immediately after terminal events, exercising the same-timestamp replacement behavior.

Both paths would use identical request IDs, token/prefix identities, engine settings, and round-robin placement. We could compare canonical per-request records and semantic report fields, excluding measured execution-time fields, and repeat each arm to check determinism.

It would also help to verify that the trace slice actually exercises overlapping requests, cache reuse, and replacements while other requests remain active. The existing focused rejection and cancellation tests could continue covering cases the trace may not naturally trigger.

crates/core/src/replay/agg.rs
```diff
@@ -1524,3 +1578,235 @@ mod host_offload_tests {
...
+            }
+            super::state::AggRequestPhase::Running => {
+                let scheduler_id = scheduler_id.ok_or_else(|| {
+                    anyhow::anyhow!("offline replay running request {uuid} has no scheduler")
+                })?;
+                let effects = self.engine.apply_command(
+                    scheduler_id,
+                    Command::CancelRequest {
+                        request_id: uuid,
+                        discard_pending_output: true,
```
PeaBrane, line 1672: [P1] Retire replay ownership when canceling a pending final output

For a one-output request, submit followed by step_until(0.0) can start a nonzero-duration final pass without delivering its completion. Canceling then returns Noop from the native scheduler because it has already computed the final output and removed the request. This flag suppresses the pending terminal, but EngineComponent retires its tracked UUID only from command retired_requests or terminal completion outputs; neither is present in this case. Public in_flight() becomes zero, while internal ownership remains one permanently. Once the pass settles, next_event_ms() is None, is_idle() stays false, and take_report() remains blocked. Please retire replay ownership exactly once when suppressing a pending terminal. This concerns per-request cancellation while the simulation continues.

crates/core/src/replay/report.rs
```diff
@@ -1339,6 +1343,30 @@ impl TraceCollector {
...
+        let stats = self.requests.get(&uuid)?;
+        Some((
+            stats.first_admit_ms?,
+            stats.first_admission_reused_input_tokens,
+        ))
+    }
+
+    /// Drain measurements while retaining the configuration that applies to
+    /// each reporting epoch of a reusable runtime.
+    pub(crate) fn take_report(&mut self) -> ReplayReport {
```
PeaBrane: [P2] Use the reporting period's elapsed duration after draining measurements

This resets request totals and accumulated worker-seconds, but the runtime keeps its absolute clock. finish() still computes duration_ms as the maximum absolute terminal timestamp and uses it for throughput and goodput. After a report at 1,000 ms, a second workload finishing at 2,000 ms therefore divides only its new requests/tokens by two seconds, while worker-seconds and GPU-hours cover only the second period. Please retain the reporting period's start and use elapsed period duration for aggregate rates, while preserving absolute per-request timestamps.

crates/core/src/replay/agg.rs
```diff
@@ -1524,3 +1578,235 @@ mod host_offload_tests {
...
+
+    pub(crate) fn take_step_terminals(&mut self) -> Vec<(Uuid, ReplayTerminalStatus)> {
+        std::mem::take(&mut self.step_terminals)
+    }
+
+    /// Admit `request` at the current simulated time. The returned id
+    /// correlates the request with later measurements.
+    pub(crate) fn submit_dynamic(&mut self, request: DirectRequest) -> anyhow::Result<Uuid> {
+        let arrival_time_ms = self.now_ms;
+        let uuid = self.assign_request(
```
PeaBrane, line 1632: [P2] Avoid retaining an arrival when dynamic submission fails

With the default single-rank engine, submit a request with UUID X and preferred_dp_rank: Some(1). assign_request() records the collector arrival before round-robin placement rejects the out-of-range rank, and this error returns without rollback. X never becomes live, so cancel(X) cannot remove it; correcting the rank and retrying X then fails with "already retained". A report also includes an incomplete request that was never accepted. The old consuming run() discarded its runtime on this error, but the new reusable submit API exposes the partial update. Please validate before committing arrival/placement bookkeeping, or roll back a failed submission.

crates/core/src/replay/loadgen/steppable.rs
```diff
@@ -0,0 +1,698 @@
...
+
+    /// Advance one logical timestamp of work.
+    fn step(&mut self) -> anyhow::Result<StepOutcome> {
+        self.step_until(f64::INFINITY)
+    }
+
+    /// Advance at most through `until_ms`, emitting events at or before that
+    /// deadline. When the next internal event is later than the deadline the
+    /// runtime stops at the deadline, so an external discrete-event driver can
+    /// interleave its own arrivals and firing gates without changing batch
```
PeaBrane, line 134: [P3] Document the same-timestamp ordering and qualify batching parity

Please document that terminal events yield before the next pass is scheduled, while nonterminal token events may schedule the next pass before returning. If A emits a nonfinal token at time t and the driver then submits B at t, A's next pass may already be committed, so B can join a later batch than it would under run(), which releases ready arrivals before driving workers. This is a documentation-only request; retaining the current ordering is acceptable. The difference requires an arrival at the same timestamp as the internal event, including deliberately synchronized first-token-triggered work.

PeaBrane (commented):
Please address the cancellation accounting, reporting-period duration, and failed-submission bookkeeping issues below. The same-time batching item is a nonblocking documentation request; strict parity with run() for that ordering is not required.

These findings are from static review; no tests were run.

PeaBrane (approved):
✅ Approved contingent on addressing the three correctness findings in [the review comments](https://github.com/ai-dynamo/aisimulate/pull/161#pullrequestreview-5171446736): cancellation accounting, reporting-period duration, and failed-submission bookkeeping, plus documenting the same-time ordering.

The same-time ordering item is documentation-only; retaining the current scheduling behavior is acceptable. The fixes have not yet been verified.

ajcasagrande:
# One-off Mooncake parity: `run()` vs `SteppableAgg` (PR #161)

**Result: PASS.** Both modes are byte-identical between the two arms, and both arms are
internally deterministic across fresh processes.

This is one-off validation evidence for PR #161. Nothing here is checked in, added to CI,
or proposed as a regression fixture.

## What was compared

Both arms execute the *same* revision, trace slice, engine descriptor, topology, and
placement. They differ only in who owns the loop:

| Arm | Driver |
| --- | --- |
| `run` | `Replayer::run()` drives itself from its own `AdmissionQueue`. |
| `steppable` | An external discrete-event driver owns the clock: it pulls ready turns from an identically constructed `WorkloadDriver`, submits them through `SteppableAgg::submit`, and advances with bounded `step_until` calls, feeding tokens and terminals back into the workload. |

The external loop mirrors `AggRuntimeImpl::run()`'s ordering — settle the current instant,
admit whatever the workload releases at that instant, then advance to
`min(next authored arrival, next engine event)`.

Comparison object: `CanonicalReplayRecord` (`coverage` + `metadata` + `per_request` +
`planner` + `summary`), whose closed `CANONICAL_RESULT_EXCLUSIONS` already drops the
measured-execution-time fields — `/summary/wall_time_ms`,
`/summary/processed_tokens_per_s`, `/summary/processed_output_tokens_per_s`. Everything
else, including all 5,000 per-request records and every virtual-time field, is compared.

## Revisions, toolchain, host

- aisimulate PR #161 head: `2f720ccfd3515abc1e1db907c9717417ec21b930`
  (`ajc/replay-steppable-aggregated`), merge-base with `origin/main`
  `b6b0a38618aa59181a6bdc80ae9d6f43082c17c1`.
- Detached worktree, clean: `/home/anthony/nvidia/projects/aisimulate-pr161`.
- `aisimulate-core` with `default-features = false` (`default = []`).
- `cargo 1.97.1 (c980f4866 2026-06-30)` / `rustc 1.97.1 (8bab26f4f 2026-07-14)`, release profile.
- Host: Intel Core i9-14900K, 32 logical CPUs, 125 GiB RAM, Linux. No CPU pinning
  (semantic virtual-time results only; no performance claim is made).
- Harness source SHA-256: `548961bb2b195bb993fead693aced114d9dd8cce96107ec222336cadb1c2e33f`.
- Runner binary SHA-256: `35936802895443660a451bf5c0b7c43d8429800f7b37e031ec23b9e99b462561`
  (4,087,928 bytes). One binary served every row.

## Workload

The fixed, checksummed 5,000-request Mooncake window from the replay-parity workflow:

- upstream `https://github.com/kvcache-ai/Mooncake` @ `dedfbde5daa3d5bc020adfd1d6e01c1d544f2098`,
  path `FAST25-release/arxiv-trace/mooncake_trace.jsonl`;
- full file: 23,608 rows, SHA-256 `b434f1816a707f4bac697235588184ebc374c9907cb981bb65fb0643471fe711`;
- slice: rows 0–4,999 in source arrival order, SHA-256
  `3892ae19ae480b643155f0c6b9d798591cbe2e73bec6a0fa5ae3d3bc0332fb8a`;
- trace block size 512, arrival speedup 4.0, `normalize_session_starts()`.

Both checksums were reproduced from the pinned commit before the campaign and re-verified
after each run.

## Frozen engine configuration

Identical in every row: vLLM backend, aggregated topology, 4 logical workers, zero startup
delay, 6,144 GPU blocks, engine block size 64, `max_num_seqs` 16, `max_num_batched_tokens`
8,192, polynomial timing, DP=1, TP=1, `round_robin` placement, `none` scaling.

Request identity is assigned once, on the shared `WorkloadDriver`, via
`with_deterministic_request_ids(1)` — so both arms carry the same ordinal UUIDs, and the
steppable arm does not depend on `Replayer`'s own lowering. Token and prefix identities are
therefore byte-identical: `ReadyTurn`'s `DirectRequest` and the deferred payload the
self-driving arm queues both materialize through `synthesize_validated_trace_tokens`.

## Modes

| Mode | `run` arm | `steppable` arm |
| --- | --- | --- |
| Trace arrivals | `spec.max_in_flight = None` → `ReplayMode::Trace`; `WorkloadDriver::new_trace`. | Same driver; the external loop advances to each authored timestamp and submits there. |
| Fixed concurrency | `spec.max_in_flight = Some(64)`; `WorkloadDriver::new_concurrency(.., 64)`. | Same driver; the cap is owned by the driver on both arms, and replacements are submitted at the terminal's instant. |

## Results

| Mode | Canonical SHA-256 (both arms, both repetitions) |
| --- | --- |
| Trace arrivals | `0a5626e72831a48812b08c20beac57f30511f9120e23d99c19866c84ec83a1a8` |
| Fixed concurrency (C=64) | `b0781c690879322d32217129e7fd6dd78704028e6393220353e4eb9e2cb5107f` |

Eight fresh processes total (2 modes × 2 arms × 2 repetitions). `cmp` on the raw
`canonical.jsonl` bytes — not only the digests — passed for every determinism pair and both
cross-arm pairs.

Shared semantic values (identical across arms by construction of the byte comparison):

| Signal | Trace arrivals | Fixed concurrency |
| --- | ---: | ---: |
| Completed / total requests | 5,000 / 5,000 | 5,000 / 5,000 |
| Rejected, canceled, failed, short-output | 0 | 0 |
| Input tokens | 46,542,297 | 46,542,297 |
| Requested = emitted output tokens | 922,544 | 922,544 |
| Decode workers observed | 0–3 | 0–3 |
| Placements | 5,000 `agg_immediate` | 5,000 `agg_immediate` |
| Virtual duration (ms) | 588,733.1598880008 | 649,416.7956339957 |
| Mean TTFT / ITL / E2E (ms) | 179,349.838 / 38.539 / 186,422.173 | 2,006.452 / 33.346 / 8,125.814 |

The trace-arrivals row reproduces the qualified `vllm-aggregated` golden point of the
replay-parity skill exactly (5,000 completed, 46,542,297 input tokens, 922,544 output
tokens, 4,910 requests with reuse, decode workers 0–3, virtual duration
588,733.1598880008), which independently confirms the harness is wired to the same
semantics.

## The slice does exercise the behaviors under test

| Behavior | Trace arrivals | Fixed concurrency |
| --- | ---: | ---: |
| Max concurrently *scheduler-admitted* requests | 62–64 | 62 |
| Requests admitted while others were live | 4,991 | 4,991 |
| Requests first-admitted at exactly a terminal timestamp | 4,672 | 2,793 |
| Requests with prefix reuse | 4,910 | 4,920 |
| Reused input tokens | 14,957,184 | 14,997,376 |
| `prefix_cache_reused_ratio` | 0.3214 | 0.3222 |
| Preemption readmissions | 2 | 1 |

External-driver counters (steppable arm only):

| Counter | Trace arrivals | Fixed concurrency |
| --- | ---: | ---: |
| `step_until` calls | 66,818 | 93,268 |
| Steps that held their instant (delta cycle) | 5,032 | 4,830 |
| Submissions | 5,000 | 5,000 |
| Submissions with others in flight | 4,999 | 4,999 |
| **Same-instant replacements** | 0 | **4,936** |
| Token / terminal events | 922,544 / 5,000 | 922,544 / 5,000 |
| Max submitted-not-terminal | 3,243 | 64 |

Reading these: the trace arm has 0 same-instant replacements *by construction* — admission
there is gated by authored timestamps, not by terminals — which is exactly why the
fixed-concurrency arm is needed. The fixed-concurrency arm exercises the same-timestamp
replacement path 4,936 times and still lands byte-identical, which is the strongest single
piece of evidence here: the held-instant delta cycle reproduces `run()`'s in-drain
admission rather than deferring a replacement by a step.

The 3,243 vs 62–64 gap in the trace arm is not a discrepancy: `SteppableAgg::in_flight()`
counts submitted-not-yet-terminal requests, while the report sweep-line counts
scheduler-admitted ones, and 4 workers × `max_num_seqs` 16 = 64 is the batching ceiling.

## Seam gaps found (not defects; scope notes for the PR)

1. **No session-metadata seam.** `run()` threads `session_id` / `turn_index` (and
   `authored_request_id` / `play_id`) from the ready arrival into
   `TraceCollector::on_session_metadata` / `on_agentic_metadata`. `SteppableAgg::submit`
   takes a bare `DirectRequest` and has no equivalent. Rather than exclude those fields
   from the comparison, both arms ran with `WorkloadDriver::without_session_metadata()`, so
   the byte comparison stays total. `AggregatedRoundRobinPlacement` ignores `session_id`, so
   this affects reported metadata only — not admission, batching, or timing.
2. **No capture-options seam.** `SteppableAgg` exposes `set_capture_per_request` and
   `set_sla_thresholds` but not `ReplayCaptureOptions`, so `capture_canonical_evidence`
   (pressure and KV-ingest evidence) and `capture_lifecycle_evidence` cannot be turned on.
   Both arms therefore ran with them off, and `pressure_records` is 0 on both. The
   preemption edge is still evidenced and compared: `readmission_count` lives in the
   per-request records, and shows 2 (trace) / 1 (concurrency) readmissions, consistent with
   the 2 preemptions the skill's golden point records for this capacity.
3. **Idle-gap advance is caller-mandatory.** When the engine has no pending event and the
   next work is an external arrival, `step_until(t)` cannot advance time — its internal
   `next_timestamp()` is `None`, so the bounded-step branch is skipped and `now_ms` stays
   put. The caller must call `advance_now_ms(t)`. This is documented on the trait, but it is
   an easy way to write a spinning driver; a doc example of the canonical pump loop would
   help.

## Not covered

- Disaggregated topology: PR #161 adds only the aggregated seam.
- vLLM preemption / SGLang retraction at a tuned capacity edge, attention-DP ranks,
  handoff, agentic and multi-turn-delta workloads, cancellation, and rejection. The trace
  does not naturally force most of these; the PR's existing focused tests continue to cover
  cancellation, rejection, duplicate-id, and held-instant behavior directly.
- Performance. No timing gate was run and none is claimed; `measured_process_ms`
  (~0.3–0.6 s per row) is a diagnostic only.

## Reproduce

```
worktree   /home/anthony/nvidia/projects/aisimulate-pr161 @ 2f720ccf
harness    /tmp/aisimulate-pr161-steppable-parity/runner
configs    /tmp/aisimulate-pr161-steppable-parity/configs/{trace_arrivals,fixed_concurrency}-{run,steppable}.json
outputs    /tmp/aisimulate-pr161-steppable-parity/out/<row>-<rep>/{canonical.jsonl,qualification.jsonl}
```

Config SHA-256:

| Config | SHA-256 |
| --- | --- |
| `trace_arrivals-run.json` | `c30418cac2b29a39a7b19cfaac4952a447069b3860dacd45e9f89b4406cc99f2` |
| `trace_arrivals-steppable.json` | `89821937b713bbb88f4a985ac173d0683cb1d33fcb1dcc9d865468c1154decd9` |
| `fixed_concurrency-run.json` | `c4a1895b83bc3dcf739539b089bb53ea73768e9ff4c890644a9be7e39ecbc155` |
| `fixed_concurrency-steppable.json` | `4a26cb24a4dd1c0d7c1ac222882b8a931aca510686501f831bc070f92a994641` |

The two configs in a mode differ only in the `arm` field.

crates/core/src/replay/agg.rs
ajcasagrande: Fixed in 4fa2309a. Suppressing a pending final output now emits the request retirement delta, so replay releases scheduler ownership when the scheduled pass settles. Added a regression covering cancellation after the final pass starts.

crates/core/src/replay/report.rs
ajcasagrande: Fixed in 4fa2309a. The collector now tracks an absolute reporting-epoch start; aggregate durations and static worker-seconds use elapsed epoch time while request timestamps stay absolute. The next epoch starts at the runtime timestamp after a successful drain. Added a two-epoch duration regression.

crates/core/src/replay/agg.rs
ajcasagrande: Fixed in 4fa2309a. Placement now succeeds before collector and traffic arrival is committed, and aggregated round-robin validates before advancing its cursors. A rejected preferred rank therefore leaves no retained request and the same UUID can be corrected and resubmitted. Added a regression.

crates/core/src/replay/loadgen/steppable.rs
ajcasagrande: Documented in 4fa2309a. Terminal events expose the held admission boundary, whereas nonterminal tokens may already have committed the next pass. Same-timestamp arrivals that require run-path batch composition must be submitted before advancing the stepper.

ajcasagrande:
# Re-run against `4fa2309a` — unchanged: PASS

Re-ran the same one-off differential campaign against the updated head. **Both canonical
digests are bit-for-bit identical to the `2f720ccf` run**, and every semantic counter is
unchanged.

| Item | Previous | This run |
| --- | --- | --- |
| PR head | `2f720ccfd3515abc1e1db907c9717417ec21b930` | `4fa2309ab909ee05fdba2f62438d6446b3e04153` |
| Merge-base with `main` | `b6b0a38618aa59181a6bdc80ae9d6f43082c17c1` | unchanged |
| Harness source SHA-256 | `548961bb…c2e33f` | identical |
| Config SHA-256 (all four) | — | identical |
| Runner binary SHA-256 | `35936802…62561` (4,087,928 B) | `dbadcd68…41005a` (4,082,448 B) |

Same pinned 5,000-row Mooncake slice (`3892ae19…32fb8a`), re-verified before and after
sampling. Same frozen engine configuration. Fresh worktree checkout, fresh build directory,
fresh binary; no artifact was reused across the revision change.

## Digests

| Mode | Canonical SHA-256 (both arms × both repetitions, 8 fresh processes) |
| --- | --- |
| Trace arrivals | `0a5626e72831a48812b08c20beac57f30511f9120e23d99c19866c84ec83a1a8` |
| Fixed concurrency (C=64) | `b0781c690879322d32217129e7fd6dd78704028e6393220353e4eb9e2cb5107f` |

`cmp` on the raw `canonical.jsonl` bytes passed for all four determinism pairs and both
cross-arm pairs. Diffing the full qualification summaries against the previous campaign
returns `IDENTICAL` for all four rows once the binary digest and the process-time
diagnostic are set aside — including the external-driver counters, which sit outside the
canonical record: 66,818 / 93,268 `step_until` calls, 5,032 / 4,830 held-instant steps, and
**4,936 same-instant replacements** in the fixed-concurrency arm.

## Why the new commit does not move these rows

`fix(replay): make steppable runtime reusable after errors` is confined to paths this trace
does not produce:

- `round_robin.rs` now advances `next_worker` and the per-worker rank counter only after
  `rank_id` succeeds. All 5,000 placements here succeed, so the placement sequence is
  unchanged; the fix matters when a placement fails and previously still consumed a slot.
- `assign_request` now calls `place()` before `collector.on_arrival`. Reordering only — the
  recorded values are the same, which the byte comparison confirms.
- `TraceCollector::report_start_ms` makes `duration_ms` relative to the previous report
  epoch. Both arms take exactly one report, so the epoch starts at 0.0 on both and
  `duration_ms` is unchanged (588,733.1598880008 / 649,416.7956339957).
- `rank.rs` publishes a retirement delta when a final output is suppressed — a cancellation
  path. Zero requests are canceled here (5,000/5,000 completed).

So this campaign does not exercise the new fixes; the three focused tests added alongside
them do, and that is the right place for them.

## One note on the new module doc

The added caveat — *"Only a terminal frees this externally visible admission boundary… a
caller that needs that batch composition must submit its known same-timestamp arrivals
before it advances the steppable runtime"* — is exactly the property the trace-arrivals arm
depends on. The driver loop submits everything `WorkloadDriver::pop_ready(now_ms, ..)`
releases at the current instant *before* calling `step_until`, and bounds every step by
`min(next authored arrival, next engine event)` so time cannot cross an arrival. That is
what makes the trace arm land byte-identical despite admission there being timestamp-gated
rather than terminal-gated. Worth keeping that ordering requirement prominent — a driver
that steps first and submits second would silently produce different batches, and only a
differential run like this one would catch it.

## Previously reported seam gaps: all three still stand

The new commit adds documentation and tests to `steppable.rs`, not API surface, so:

1. No session-metadata seam on `SteppableAgg::submit` (both arms ran with
   `without_session_metadata()` so the byte comparison stays total).
2. No `ReplayCaptureOptions` seam, so pressure and KV-ingest evidence cannot be enabled;
   the preemption edge is still compared via per-request `readmission_count` (2 trace /
   1 concurrency).
3. `step_until(t)` cannot advance time when the engine is idle and the next work is an
   external arrival — the caller must call `advance_now_ms(t)`. A doc example of the
   canonical pump loop would help here, and would be a natural home for the new
   same-timestamp-submission caveat too.

Coverage limits are unchanged from the previous comment: aggregated topology only, no tuned
capacity edge, no attention-DP, no agentic or multi-turn-delta workloads, and no
performance measurement.

## Reproduce

```
worktree   /home/anthony/nvidia/projects/aisimulate-pr161 @ 4fa2309a
harness    /tmp/aisimulate-pr161-steppable-parity-r2/runner
configs    /tmp/aisimulate-pr161-steppable-parity-r2/configs/{trace_arrivals,fixed_concurrency}-{run,steppable}.json
outputs    /tmp/aisimulate-pr161-steppable-parity-r2/out/<row>-<rep>/{canonical.jsonl,qualification.jsonl}
```

dreamtalen (approved):

crates/core/src/replay/report.rs
```diff
@@ -1435,7 +1473,7 @@ impl TraceCollector {
             total_output_tokens += output_length;
             total_reused_tokens += stats.reused_input_tokens;
             total_first_admission_reused_tokens += stats.first_admission_reused_input_tokens;
-            duration_ms = duration_ms.max(terminal_time_ms);
+            duration_ms = duration_ms.max((terminal_time_ms - report_start_ms).max(0.0));
```
jasonqinzhou, line 1479: [P2] Measure reusable reports through the report boundary

The new epoch start fixes the leading-window drift, but this still ends `duration_ms` at the last completed request rather than at the reporting boundary. For example, if a request terminates at 100 ms and the external driver advances an idle runtime to 1,000 ms before `take_report`, `advance_now_ms` accumulates worker-seconds through 1,000 ms while this report divides throughput and goodput by only 100 ms. That mixes a 100 ms rate window with 1,000 ms of provisioned cost and overstates the reported rates. Please pass the current simulated report end into finalization (or otherwise use `report_end_ms - report_start_ms`) and add a regression with trailing idle time; a no-request epoch would also exercise the same boundary.

jasonqinzhou (approved):
**APPROVE · 8.1/10 · high confidence**

## What this PR does

This PR adds a public, externally clocked steppable interface over aggregated replay, including dynamic submission and cancellation, bounded event stepping, token and terminal events, and reusable report collection. It also adjusts request ownership, round-robin error atomicity, and report epoch bookkeeping to support that reusable runtime.

The delta-cycle design is carefully documented, and the focused tests cover difficult terminal-without-token, cancellation-after-final-pass, duplicate-ID, failed-submission, and multi-epoch cases that make this kind of simulation seam easier to trust.

## Why this score

8.1/10 and COMMENT: the implementation is generally sound and the focused exact-head tests pass. One reachable P2 reporting defect remains: reusable reports exclude trailing idle time from throughput duration even though worker-seconds continue through the report boundary, producing internally inconsistent rate and cost metrics. The requested Fable model was not observed, so consensus is unavailable and this result must not be described as dual-reviewed.

crates/core/src/replay/agg.rs
jasonqinzhou: Verified the fix from 4fa2309a on 3330ff54. The cancellation-after-final-pass regression passes: public and native ownership drain, next_event_ms returns None, and take_report succeeds. Full local Rust workspace tests also pass.

crates/core/src/replay/report.rs
jasonqinzhou: Verified and completed in 3330ff54. Each report now covers the elapsed interval between report boundaries, including trailing idle time and empty epochs; per-request timestamps remain absolute. The original two-epoch regression and the new rate/worker-seconds/goodput regressions pass.

crates/core/src/replay/agg.rs
jasonqinzhou: Verified the preferred-rank rollback fix from 4fa2309a on 3330ff54. Failed submission leaves the UUID reusable; the existing retry regression passes. Added passing coverage that failed rank lookup preserves both worker and per-worker rank rotation, plus pre-bookkeeping validation of immediate placement identity.

crates/core/src/replay/loadgen/steppable.rs
jasonqinzhou: Verified the module documentation on 3330ff54: terminal events expose the held admission boundary, while nonterminal tokens may already commit the next pass; callers requiring run-path batch composition must submit known same-timestamp arrivals before stepping. The held-instant and cascade regressions pass.

crates/core/src/replay/agg.rs
```diff
@@ -303,16 +324,15 @@ where
             request.metadata_mut().arrival_timestamp_ms = Some(arrival_time_ms);
         }
 
+        let effects = self
+            .placement
+            .place(&request, metadata, session_id, self.now_ms)?;
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in 3330ff54. Immediate placement identity is validated before collector arrival/context or offered-traffic bookkeeping. The new malformed-placement regression verifies no retained arrival, no offered traffic, no live ownership, and an empty report. It failed before the fix and now passes.

crates/core/src/replay/agg.rs
```diff
@@ -1524,3 +1577,237 @@ mod host_offload_tests {
...
+            anyhow::anyhow!("offline replay lost request state while canceling {uuid}")
+        })?;
+        CoreAdmissionSource::on_terminal(
+            &mut self.admission,
+            uuid,
+            self.now_ms,
+            ReplayTerminalStatus::Canceled,
+        )?;
+        self.progress.inc_completed();
+        self.step_freed_slot = true;
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in 3330ff54 by removing the ineffective step_freed_slot write from cancel_dynamic. Cancellation remains immediately terminal; the existing single-worker, multi-worker, and pending-final-pass cancellation tests pass.

crates/core/src/replay/agg.rs
```diff
@@ -1524,3 +1577,237 @@ mod host_offload_tests {
...
+        &mut self,
+        wall_ms: f64,
+    ) -> anyhow::Result<crate::replay::ReplayReport> {
+        anyhow::ensure!(wall_ms.is_finite(), "replay report wall_ms must be finite");
+        anyhow::ensure!(
+            self.is_workload_done(),
+            "replay report requires an idle runtime"
+        );
+        self.collector
+            .set_runtime_evidence(std::mem::take(&mut self.evidence).finish());
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in 3330ff54. Report draining replaces the evidence collector with a fresh collector carrying the existing ReplayCaptureOptions. The new two-epoch regression verifies canonical pressure/KV capture remains enabled after each drain; it failed in the second epoch before the fix and now passes.

crates/core/src/replay/core/round_robin.rs
```diff
@@ -293,15 +292,21 @@ impl AggregatedRoundRobin {
...
+                    .unwrap_or_default()
+                    % self.dp_size
             }
         };
-        rank_id(worker_id, rank).ok_or_else(|| {
+        let scheduler_id = rank_id(worker_id, rank).ok_or_else(|| {
             anyhow!("logical worker {worker_id} does not expose preferred attention-DP rank {rank}")
-        })
+        })?;
+        self.next_worker = index + 1;
```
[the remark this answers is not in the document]
jasonqinzhou: Added in 3330ff54: aggregated placement coverage for shrinking, removing/re-adding, and expanding the active worker set, including retained versus reset per-worker DP-rank rotation. Also verifies failed rank lookup advances neither cursor. All pass; production topology-ordering behavior is preserved.

crates/core/src/replay/loadgen/steppable.rs
```diff
@@ -0,0 +1,760 @@
...
+        }
+        if let Some(uuid) = request.uuid
+            && self.runtime.collector().contains_request(uuid)
+        {
+            anyhow::bail!("steppable replay request {uuid} is already retained");
+        }
+        let uuid = self.runtime.submit_dynamic(request)?;
+        self.live.insert(uuid);
+        Ok(uuid)
+    }
```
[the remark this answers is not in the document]
jasonqinzhou: Documented and tested in 3330ff54. UUIDs are unique among live requests and measurements retained in the current epoch. A successful idle take_report permits reuse; old request queries are cleared and subsequent events/reports reflect only the new request. The new reuse regression passes.

crates/core/src/replay/report.rs
jasonqinzhou: Fixed in 3330ff54. take_report passes the current simulated boundary into collector finalization and starts the next epoch at that boundary. Added two-epoch trailing-idle coverage for duration, rates, goodput, worker-seconds, and absolute terminal timestamps, plus empty-epoch coverage. Both regressions failed before the fix and now pass.

PR #164, opened 2026-09-10
fix(perfmodel): route DSA attention through the DSA rooflines in the FPM SOL dispatcher [AIC-1943]

Arsene12358:
## Why and what changed

Routes the DeepSeek-sparse-attention (DSA) module operators through the existing DSA rooflines in the FPM whole-forward SOL dispatcher, so off-lattice FPM queries for GLM-5.2 / DeepSeek-V3.2-class models resolve through the SOL-ratio neighbour transfer instead of failing. Consumer: `aisimulate predict` / `recommend` with `timing: {type: default, forward_model: fpm}` on a DSA model. Linear: AIC-1943.

Today `crates/core/src/perfmodel/operators/fpm_sol.rs::op_sol_latency_ms` has arms for GEMM, embedding, elementwise, dense context/generation attention, MoE, dispatch, collectives, overlap and fallback, and returns `SolNotImplemented` for everything else. GLM-5.2's attention ops are `Op::DsaContext` / `Op::DsaGeneration`, so the whole-forward SOL cannot be formed and every off-site query dies in the `ScatteredSites` resolver with `no usable neighbour site … SOL roofline has no Rust implementation for op context_attention`. A chunked prefill with 8,192-token steps queries KV totals of 8,192·k, of which only k ∈ {1, 2, 4, 8, 16} are collected sites in the GLM cells, so a GLM-5.2 FPM prediction with inputs longer than one chunk failed at the third chunk. The same gate blocked the high-KV-pressure prefill batch clamp and any decode batch without a (C, C+1) capture-rung pair.

- `fpm_sol.rs`: two match arms and two helpers, `dsa_context_module_sol` and `dsa_generation_module_sol`, that call `perf_database::dsa::dsa_context_sol_ms` / `dsa_generation_sol_ms` with the op's dims (`dsa_dims(architecture)`), `index_topk`, quant modes and `num_heads`, exactly as `operators/dsa.rs::query_context_table` / `query_generation_table` do under `DatabaseMode::Sol`. Context blends the full and skip-indexer variants by the configured `full_frac` (SOL mode weights by `full_frac` directly, as `DsaModuleOp::query_context` does) and applies `scale_factor`; generation uses the bf16 attention group and no skip variant, as in the op-level decode SOL. DSA context parallelism composes latency-only table deltas in the op-level path and has no roofline, so `cp_size > 1` stays a typed `SolNotImplemented`. No new roofline formula; FPM hands per-request coordinates as f64 and they are rounded to the integers the roofline takes (documented in the module's approximation list).
- Input validation: `dsa_context_module_sol` rejects a non-finite or out-of-[0, 1] `full_frac` with the operators' existing typed configuration error before the blend (0.0 and 1.0 stay valid); the op-level path and the constructors are untouched (follow-up below).
- Tests: `fpm_sol.rs` asserts the FPM SOL leaf equals the DSA roofline functions for a GLM-5.2-shaped op (context, context with `full_frac`/`scale_factor`, generation), pins two frozen hand-derived records at fractional FPM coordinates (batch not dividing the totals, so the `f64::round` half-away-from-zero coordinate mapping is under test: a ties-even mapping fails them), covers the `full_frac` boundaries 0.0/1.0 and the five invalid classes, and checks that the CP guard and the pre-existing unsupported-op error keep their typed shape. `fpm_forward.rs` adds two integration tests through `FpmForwardOp::query_totals` on synthetic pairs, an off-site prefill KV between two collected sites and a decode batch without a rung pair, each asserting the exact transfer value computed from the resolver's documented form (inverse-distance mean of the neighbours' SOL-utilizations, SOL(query) divided by it) at 1e-9 relative tolerance.
- Docs and comments that described DSA as an unported SOL family are updated: the FPM design doc's SOL-coverage bullet (`docs/fpm/aic-fpm-rust-port-design-20260802.md`) moves DSA to the covered list with the `cp_size > 1` residual gap recorded; MSA and the MLA-module family stay listed as unported.

Exact hits and in-curve interpolation never call SOL, so on-lattice results are unchanged.

## Review map

- Risk level: low. Additive match arms on a path that previously returned a typed error; on-lattice answers are untouched and both frozen parity surfaces are unchanged.
- Start with: `crates/core/src/perfmodel/operators/fpm_sol.rs` (`dsa_context_module_sol`, `dsa_generation_module_sol`, the two `Op::Dsa*` arms), then the two new integration tests at the end of `crates/core/src/perfmodel/operators/fpm_forward.rs`, which show the two consumer paths the arms unblock.
- Public or serialized contract changed: no. No wire or schema change; `crates/tests/public-api` unchanged.
- Compatibility or rollback concern: none. Reverting restores the typed `SolNotImplemented` for DSA on the transfer paths; no data or golden migration is involved.

## Evidence

- Tests: `cargo fmt --all -- --check` clean; `cargo test -p aisimulate-core`: 1209 passed, 1 pre-existing ignored, 0 failed (new on this branch: 4 + 3 DSA SOL unit tests incl. the two frozen records and the `full_frac` domain cases, 2 integration tests); `cargo test --manifest-path crates/tests/public-api/Cargo.toml`: 4 passed; clippy adds no warnings on the two files.
- Fast CI: pass on `82ffa025` (Repository Policy, Python Static Checks, Rust Format; run 34482278471).
- Full CI: pending on this PR.
- CodeRabbit reviewed commit: `ab171b9a` (three actionable comments: exact transfer oracles, `full_frac` domain check, frozen fractional-coordinate cases; all addressed in `82ffa025`, threads resolved) and `82ffa025` (one outside-diff comment: `dsa_dims` maps every architecture other than GLM-5.2's to the DeepSeek-V3.2 dims instead of rejecting unknown strings; pre-existing behaviour of the shared helper that the op-level path relies on, so it is left for the constructor-level validation follow-up below rather than patched in the FPM arm).
- Codex reviewed commit: N/A (low risk).
- Negative or boundary cases: `cp_size > 1` context returns the typed `SolNotImplemented` (unit test); an op family the dispatcher still does not cover keeps its typed error and does not block exact hits or in-curve lerps (`unsupported_sol_family_is_lazy`, `Op::MlaBmm`); the integration tests use a KV strictly between two collected sites and a decode batch with no (C, C+1) pair.
- Expected-value derivation: the unit tests compute the expected value from `perf_database::dsa::dsa_context_sol_ms` / `dsa_generation_sol_ms` on the same integer coordinates, apply the blend `w·sol(full) + (1−w)·sol(skip)` and `scale_factor` explicitly, and assert equality with the dispatcher's answer.
- Before/after: on main, the disaggregated dep16+dep16 FPM replay of the GLM-5.2 AgentX tier means (inputs 79k–99k tokens, chunked prefill at 8,192) fails at `total_kv_read_tokens=24576` with the error above; on this branch it completes at all four tiers (c16 ITL 16.44 ms, c32 16.65, c64 16.87, c128 17.16 ms). The on-lattice aggregated smoke (synthetic 1,024/32 at concurrency 32) is byte-for-byte unchanged: TTFT 297.12 ms, ITL 17.20 ms, 58.14 tokens/s/user. The replay numbers are functional checks, not a fidelity claim: the synthetic replay has no prefix reuse, so the prefill worker throttles arrivals (TTFT 11–52 s vs 0.8–1.9 s in the truth) and the decode engine runs far below the nominal concurrency; the decode-side fidelity yardstick for this pair remains the step-level score (3.6 % MAPE).
- Parity suites (`.claude/rules/rust-core/parity.md` Rule 1.3), goldens untouched:

  ```
  AIC_ALLOW_UNLISTED_VERSIONS=1 python -m pytest -c python/aisimulate/pytest.ini \
    crates/core/parity_tests/perfmodel/test_engine_step_parity.py \
    crates/core/parity_tests/perfmodel/test_compile_engine_parity.py -q -p no:timeout
  363 passed, 1 warning in 92.31s
  ```

  Without the environment variable the same run is 6 failed / 357 passed; all six are the `qwen38-max-gb300-vllm-0271` cases failing in Python's version-slot gate (`sdk/perf_database.py`) before any Rust query, unrelated to this branch.

No new formula: the arms reuse the DSA SOL ports in `perf_database/dsa.rs` verbatim. The real-data checks above used the GLM-5.2-NVFP4 GB200 DEP16 FPM pair collected with the Dynamo vLLM self-benchmark (AISimPlayground `glm52-gb200-dep16-disagg-fpm-fidelity`), installed locally as `gb200/vllm/0.28.0` with `AIC_ALLOW_UNLISTED_VERSIONS=1`; it is not bundled here.

The transfer is an estimate: it scales a neighbouring site's measured latency by the roofline ratio between the queried and the neighbouring shape. Its accuracy for DSA prefill was measured on a chunk-boundary prefill collection of the same GB200 DEP16 pair (7,204 points, KV sites at every 8,192·k for batches 1–8, real-seeded, 0 seed misses): the collection was filtered to the power-of-two KV lattice the published cells use (2,407 points), and every held-out chunk-boundary site (4,797) was predicted from it through this branch's extension, via `RustForwardPassPerfModel.estimate_forward_pass_time_ms` with one FPM record per point, and compared with the site's own measured wall (on `main` every one of these queries fails with the error above).

| held-out points | MAPE | median APE | p90 APE | bias |
|---|--:|--:|--:|--:|
| all resolved (4,218) | 5.8 % | 3.2 % | 14.0 % | −0.7 % |
| production chunk sizes, tokens ≥ 4096 (292) | 0.6 % | 0.3 % | 1.4 % | +0.4 % |
| tokens ≥ 4096, KV < 200k, interpolation (239) | 0.4 % | 0.3 % | 1.0 % | +0.2 % |
| tokens ≥ 4096, KV ≥ 200k, hold beyond the deepest lattice site (53) | 1.5 % | 0.8 % | 3.4 % | +1.2 % |
| tokens < 1024 (3,552) | 6.5 % | 3.9 % | 15.1 % | −1.0 % |

At production chunk sizes the transfer error is at the level of run-to-run repeatability between two collections of the same cell (1.0 % MAPE, −0.5 % bias at shared coordinates). Tiny prefill steps (< 1024 tokens) carry sporadic 3–5× measurement outliers in the collections themselves, which dominate the tail there. 579 held-out queries at total KV ≥ 237,568 were refused by the existing collected-domain gate (the lattice-only cell's deepest site is 131,072 and the domain extends to 1.75× of it); the published cells reach 253,952–262,080 and do not hit it. Per-point CSVs and the driver are attached to AIC-1943. Two behaviours worth knowing when reading transfer results for DSA models:

- The high-KV-pressure prefill batch clamp, newly reachable for DSA, rescales the measured ceiling row by the whole-model SOL ratio and caps that ratio at 1.0. DSA's KV-cache bytes grow linearly with batch once the top-k window saturates, so the ratio can exceed 1 and the cap binds; the result then degenerates to the raw clamped row. Bounded and safe, but the leave-one-out study should expect it at the deepest batch rows.
- `dsa_generation_module_sol` does not blend a skip-indexer variant, matching the op-level decode SOL, which ignores the skip split. If the decode-side SOL ever gains a skip split, this helper must follow.

## Tracking

- Closes: AIC-1943 (Linear).
- Related PRs or issues: #126 (`forward_model` on the unified CLI, the path on which this failure surfaced); #13 / AIC-1770 (no overlap, `fpm_sol.rs` untouched there). Follow-ups: the MLA-module, MSA and DSv4 operators have the same dispatcher gap; `DsaModuleOp::full_frac` should be validated at construction (`py_ops.rs` and the wire form) so the op-level blend inherits the check the FPM arm now applies, and the same constructor check should reject architectures `dsa_dims` does not know instead of defaulting them to the DeepSeek-V3.2 dims (CodeRabbit's outside-diff comment on `82ffa025`); the DSA generation roofline's compute terms are not pinned by a frozen record because the decode roofline is memory-bound at every reachable FPM shape (a test for them belongs in `perf_database/dsa.rs`).


Arsene12358:
@coderabbitai The three inline comments on `ab171b9a` are addressed in `82ffa025`:

1. **Exact transfer oracles.** Both integration tests (`prefill_off_site_kv_resolves_through_the_dsa_sol_transfer`, `decode_pairless_batch_resolves_through_the_dsa_sol_transfer`) now assert the exact transfer value computed in the test from the resolver's documented form (inverse-distance mean of the neighbours' SOL-utilizations, `SOL(query)` divided by that mean, `perf_interp.rs`) at 1e-9 relative tolerance. A raw-neighbour or arithmetic-mean answer no longer passes (the arithmetic-mean alternative differs by 1.7e-4 / 1.2e-2 relative on these fixtures).
2. **`full_frac` domain.** `dsa_context_module_sol` rejects a non-finite or out-of-`[0, 1]` `full_frac` with the operators' existing typed configuration error before the blend; 0.0 and 1.0 stay valid. Tests cover 0.0, 1.0, −0.1, 1.1, NaN, +inf and −inf and assert the variant.
3. **Independent oracle and fractional coordinates.** Two frozen hand-derived records (context, math-bound; generation, memory-bound) at FPM totals the batch does not divide pin the `f64::round` half-away-from-zero coordinate mapping; a ties-even mapping fails both while the pre-existing oracle-equality tests would not notice. The `full_frac` boundary cases 0.0 and 1.0 are covered on the context arm.

On the outside-diff comment about `dsa_dims` defaulting unknown architecture strings to the DeepSeek-V3.2 dimensions: agreed that unknown strings should be rejected, but this is the shared helper's pre-existing contract that the op-level DSA path (`operators/dsa.rs`) relies on, and the FPM arms call it exactly as that path does. The right place for the check is the `DsaModuleOp` constructors (`py_ops.rs` and the wire form), together with the constructor-level `full_frac` validation, so that both the op-level and the FPM paths inherit it; it is recorded as a follow-up in the PR description rather than patched in the FPM arm alone.

jasonqinzhou (approved):

PR #167, opened 2026-09-10
feat: load whole-forward FPM data from external parquet

simone-chen:
## Summary
- Load whole-forward FPM data from an external `fpm_parquet_path` across prediction/recommendation YAML, Python, PyO3, and Rust APIs; remove the checked-in B200/H200 pairs.
- Preserve each role's path through search samples, deployment metadata, runtime arguments, candidate YAML, and regular prefill/decode companions in AFD+PD. Companion compilation also preserves accepted precision overrides and custom `systems_path`, using the same alias validation as regular replay.
- Anchor relative external paths at engine construction for both cache identity and lazy loading.
- Reject empty or non-UTF-8 paths and paths on non-FPM engines before loading data; retain the current speculative-FPM shape validation.
- Keep external tables memoized and allow covered cells plus default KV-capacity estimation without a backend/version data directory.
- Preserve sidecar schema, digest, row-count, and identity validation; document concrete parquet/sidecar names and external-input examples.
- Replace bundled-data replay/CLI-detail coverage with a temporary external pair and explicit numerical assertions.

## External parquet example
- Supply a local whole-forward FPM parquet and its adjacent sidecar:
  - `/data/reviewed-fpm.parquet`
  - `/data/reviewed-fpm.metadata.json`
- The pair must match the model, hardware, backend/version, parallelism, and precision settings. For the unlisted version below, run with `AIC_ALLOW_UNLISTED_VERSIONS=1`.

```python
from aiconfigurator_core.sdk.engine import EngineHandle

engine = EngineHandle.compile(
    model_path="MiniMaxAI/MiniMax-M2.7",
    system="h200_sxm",
    backend="vllm",
    backend_version="0.25.1",
    tp_size=4,
    moe_tp_size=4,
    moe_ep_size=1,
    forward_model="fpm",
    fpm_parquet_path="/data/reviewed-fpm.parquet",
)

latency_ms = engine.predict_prefill_latency(bs=1, isl=1024)
```

- For prediction or recommendation YAML, add this timing configuration to the matching worker:

```yaml
engine:
  workers:
    aggregated:
      timing:
        type: default
        forward_model: fpm
        fpm_parquet_path: /data/reviewed-fpm.parquet
```

- Disaggregated configurations use `workers.prefill.timing` and `workers.decode.timing`; each role may use a different pair.
- Rust reads and validates the pair on the first FPM query. Files stay outside the repository and must be accessible to the executing process. Invalid or missing artifacts raise an error without silently switching to operator-level prediction.

## Validation
- Merged `main` at `ba64235a` and resolved compiler/deployment metadata conflicts, preserving external FPM and speculative-decoding fields.
- `cargo test -p aisimulate-core --features python`: **1,519 passed, 1 ignored**. Updated hand-built Rust fixtures to resolve declared FP8-block donors after the data correction on `main`; assertions and production timing behavior are unchanged.
- Runner, AFD, configuration, and sweeper suites: **841 passed, 5 downstream integration-only skips**.
- FPM and cross-package public API suites: **81 passed**. Ngram speculation and unified traffic suites: **69 passed**. Python tests used `-p no:timeout` and the native extension rebuilt after the merge.
- Companion replay tests cover both phases, two output lengths, plain and `aic_` precision names, all five quantization fields, and a system present only in the supplied custom systems directory with the environment override removed.
- Regression coverage also verifies live engines across directory changes, lazy loading after a directory change, and missing/corrupt external artifacts.
- Synthetic pairs produce 22 ms prefill and 6 ms decode; a second dataset produces 99 ms prefill. These fixtures verify data selection and execution, not silicon accuracy.
- Changed-file Ruff checks, Rust formatting, documentation links, whitespace, and packaged legal files passed.

- CI baseline provenance now references permanent main commit `ba64235a` instead of a pre-squash PR commit. All eight numerical sentinels passed independently at that main commit and this PR, with exactly matching results; values, query inputs, and tolerances are unchanged.
- All 291 repository-policy tests passed locally using the CI tool versions.

- Alias-routing follow-up (`38effc46`): normalize plain and `aic_` timing-selection aliases before choosing external FPM. The four plain-name replay cases fail before the fix and pass afterward; 183 runner/AFD tests and 81 FPM/public API tests passed, including duplicate/invalid alias checks.


python/aisimulate/src/aisimulate/config/engine.py
```diff
@@ -73,6 +73,7 @@ class KvCachePredictionConfig(StrictModel):
 class TimingConfig(StrictModel):
     type: Literal["default", "fixed", "polynomial"] = "default"
     forward_model: Literal["op_level", "fpm"] = "op_level"
+    fpm_parquet_path: str | None = None
```
jasonqinzhou, line 76: `TimingConfig` is shared by prediction and recommendation, but the recommendation pipeline only copies `forward_model`: `_engine_search_space` has no per-role path field, `SearchSpace`/`unroll_sample` cannot retain it, and `build_backend_deployment` plus `_candidate_prediction` emit no path. I reproduced an accepted recommendation with `/artifacts/reviewed-fpm.parquet`; the sample had no path, both deployment timing metadata and `aic_fpm_parquet_path` were `None`, and the candidate YAML contained only `forward_model: fpm`. Because this PR removes the bundled pairs, those candidates cannot reach their configured data. Please carry the path per role through the search space, sample, deployment metadata/runtime args, and emitted candidate YAML, with a round-trip regression that starts from a recommendation containing the path.

python/aisimulate/src/aiconfigurator_core/sdk/engine.py
```diff
@@ -292,6 +293,7 @@ def _engine_config_dict(
         # Always a literal version directory name, never a slot alias — the
         # Rust side reloads the perf database from this string verbatim.
         "backend_version": _literal_backend_version(system, backend, backend_version, systems_path, database),
+        "fpm_parquet_path": fpm_parquet_path,
```
jasonqinzhou, line 296: This serializes `fpm_parquet_path` independently of the selected model shape, and the direct Python/Rust surfaces do not apply `TimingConfig` validation. I reproduced `EngineHandle.compile(..., fpm_parquet_path='/definitely/missing/reviewed-fpm.parquet')` with no `forward_model`: it built 14 op-level context ops, contained no `FpmForward`, then successfully returned an op-level timing result while silently ignoring the configured missing artifact. Please reject empty paths and any path whose compiled spec is not exactly the one-prefill/one-decode FPM shape before database loading; keep a Rust-side defense for hand-built specs and cover `compile_engine`, `AicEngineBuilder`, `ForwardPassPerfModel::from_native`, and raw AIC timing payloads.

crates/core/src/perfmodel/engine/runtime.rs
```diff
@@ -452,7 +452,12 @@ impl Engine {
             matches!(
                 spec.engine.database_mode,
                 DatabaseMode::Empirical | DatabaseMode::Sol
-            ) || spec.engine.tolerate_dirless_version,
+            ) || spec.engine.tolerate_dirless_version
+                // An external whole-forward FPM artifact is self-contained
+                // for covered cells. Keep op tables lazy so no in-repository
+                // backend/version directory is required just to load it.
+                || spec.engine.fpm_parquet_path.is_some(),
```
jasonqinzhou, line 459: Using `fpm_parquet_path.is_some()` as `tolerate_missing_data` takes the early branch in `load_resolved_shared_in`, which deliberately bypasses `SHARED_TABLES`. Every engine in a sweep therefore gets a separate `FpmForwardTable`/`OnceLock` and re-hashes and re-parses the same external parquet on first use, even though this PR already adds the path to `shared_tables_key`. Please separate the external-FPM directory allowance from estimate-only missing-data tolerance, keep the memo active for a valid external FPM shape, and add a same-path sharing regression.

tests/test_unified_traffic_runtime.py
```diff
@@ -272,29 +284,15 @@ def test_prediction_spec_lowers_forward_model_per_role_in_disaggregated_mode() -
...
-    # check for the data path, not an accuracy statement about either model.
+def test_engine_stack_fpm_timing_fails_closed_when_external_parquet_is_missing(monkeypatch) -> None:
+    # The explicit external path is absent. FPM must refuse rather than fall
+    # back to op_level; the same shape still replays under op_level timing.
     monkeypatch.setenv("AIC_ALLOW_UNLISTED_VERSIONS", "1")
     engine = _fpm_engine()
-    engine["workers"]["aggregated"]["parallelism"] = {"tensor": 2, "moe_tensor": 2, "moe_expert": 1}
 
-    with pytest.raises(RuntimeError, match="FPM"):
+    with pytest.raises(RuntimeError, match="reviewed-fpm.parquet"):
```
jasonqinzhou, line 293: The replacement runtime test proves only that a missing external file fails. The new successful tests stop at `select_cell`, while this PR deletes the prior CLI/runtime success case, so no checked-in test proves that a valid external parquet and sidecar survive YAML/Python/PyO3/Rust lowering and drive prefill and decode. Please add a temporary valid pair through `fpm_parquet_path`, omit the backend/version data directory, run the real final consumer, and assert the expected prefill/decode results; retain this missing-file negative case.

crates/core/src/perfmodel/py.rs
```diff
@@ -1434,6 +1454,10 @@ fn engine_build_request(config: &EngineConfig, systems_path: Option<&str>) -> En
         kv_block_size: config.kv_block_size,
         systems_path: systems_path.map(str::to_owned),
         forward_model: config.forward_model.clone(),
+        fpm_parquet_path: config
```
jasonqinzhou, line 1457: `ForwardPassPerfModel::from_native` accepts an OS-native `PathBuf`, but `to_string_lossy()` replaces invalid bytes before the request crosses Python. The resulting engine points at a different path and reports the configured artifact as absent. Please use explicit UTF-8 validation and return a targeted configuration error instead of silently changing the path, with a Unix regression for a non-UTF-8 component.

crates/core/src/perfmodel/perf_database/fpm_forward.rs
```diff
@@ -8,11 +8,12 @@
 //! collector pair
 //!
 //! ```text
-//! <data_root>/fpm_forward_perf.parquet
-//! <data_root>/fpm_forward_perf.metadata.json
+//! <configured path>.parquet
+//! <configured path>.metadata.json
```
jasonqinzhou, line 12: `fpm_parquet_path` is the complete parquet filename and `with_extension("metadata.json")` replaces its `.parquet` suffix. The current `<configured path>.parquet` / `<configured path>.metadata.json` notation can be read as requiring `reviewed-fpm.parquet.metadata.json`, while the loader expects `reviewed-fpm.metadata.json`. Please show one concrete pair, such as `/data/reviewed-fpm.parquet` and `/data/reviewed-fpm.metadata.json`, so the module contract matches the implementation and public guide.

jasonqinzhou (dismissed):
**REQUEST_CHANGES · 5.5/10 · high confidence**

## What this PR does

This PR removes the checked-in B200 and H200 whole-forward FPM datasets and adds an external `fpm_parquet_path` across prediction YAML, Python, PyO3, Rust configuration, and the Rust FPM parquet loader. The loader retains sidecar digest, schema, row-count, and identity validation while allowing covered FPM cells to run without an in-repository backend/version directory.

The external loader preserves the existing fail-closed sidecar and row-identity checks, includes the configured path in the shared-table identity, documents the intended public surfaces, and adds focused arbitrary-stem and directory-less loader tests.

## Why this score

The technical score is 5.5 and the recommendation is REQUEST_CHANGES. The basic external prediction path works and the focused and parity suites are green, but two P1 contract failures remain: recommendation lowering drops the accepted external path before candidate execution, and the direct Python/Rust core APIs silently accept an FPM path on an op-level engine and return op-level predictions. The external path also disables the existing shared-table memo, there is no checked-in successful final-consumer regression, and two narrower path/documentation issues remain.

crates/core/src/perfmodel/perf_database/mod.rs
```diff
@@ -59,11 +59,13 @@ fn shared_tables_key(
...
     resolver: &SourceResolver,
+    fpm_parquet_path: Option<&Path>,
 ) -> String {
     format!(
-        "{}\x1f{system}\x1f{backend}\x1f{version}\x1f{}",
+        "{}\x1f{system}\x1f{backend}\x1f{version}\x1f{}\x1f{}",
         systems_root.display(),
-        resolver.identity_key()
+        resolver.identity_key(),
+        fpm_parquet_path.map_or_else(String::new, |path| path.display().to_string())
```
Arsene12358, line 68: **[P1] Anchor the external path before caching and lazy loading**

The new cache key uses the raw relative string, and `FpmForwardTable` also retains that unresolved path. All new public boundaries accept relative paths. With the same absolute systems root, two valid pairs named `fpm_forward_perf.parquet` in different directories collide if the process changes directory between engine constructions and the first engine stays alive.

I reproduced this through `EngineHandle.compile` using synthetic schema-v6 H200/vLLM pairs for `Qwen/Qwen3-0.6B`. The first pair has a 22 ms prefill and the second a 99 ms prefill, both at `bs=1, isl=512`. After querying the first engine, constructing/querying the second engine from its directory with the same relative filename returns **22 ms**. Passing the second file's absolute path returns **99 ms**. A separate fresh table constructed in the first directory but first queried after changing to the second directory returns **99 ms instead of 22 ms**, so fixing only the memo key is insufficient.

| Current | Required |
| --- | --- |
| The raw relative path identifies the shared table and is resolved implicitly at first read. | Resolve the accepted path once against the working directory at construction and use the same absolute path for both table identity and loading; alternatively, explicitly reject relative inputs consistently. |

Please cover both two live engines across a directory change and a directory change before the first query. The systems root remains absolute and unchanged in these reproductions; this is introduced by the external-path field, separate from any preexisting relative systems-root behavior.

python/aisimulate/src/aisimulate/compiler.py
```diff
@@ -360,6 +363,8 @@ def _worker_engine_args(
     if worker.timing.type == "default" and worker.timing.forward_model != "op_level":
         # Only the non-default forward model is spelled out, so op_level specs stay byte-identical.
         payload["aic_forward_model"] = worker.timing.forward_model
+        if worker.timing.fpm_parquet_path is not None:
+            payload["aic_fpm_parquet_path"] = worker.timing.fpm_parquet_path
```
Arsene12358, line 450: **[P1] Preserve the external path through the regular AFD+PD companion estimator**

This field is preserved into deployment arguments, but the ordinary P/D companion uses a separate execution path: `AICAFDCompanionPerformanceModel.measure` in `runner.py` builds kwargs containing `forward_model` and calls `cli_estimate` without `fpm_parquet_path`. `cli_estimate` also lacks the new parameter. The same gap applies to the arguments produced by `sweeper/deploy.py`.

The affected configuration is, for example, `engine.mode: afd`, `engine.afd.phase: decode`, `combined_with_pd: true`, and an ordinary `workers.prefill` worker with `timing.forward_model: fpm` plus an external path. The AFD attention/FFN workers are not using these profiles; the regular prefill companion is. Existing `test_public_afd_companion_forward_model_reaches_estimator` already covers FPM selection on both companion phases, but mocks the estimator and does not exercise the new file input.

I supplied a valid H200/vLLM 0.24.0/Qwen3-0.6B pair whose 8-token prefill is 22 ms. Direct `EngineHandle.compile(..., fpm_parquet_path=...)` returns **22 ms**. Public configuration validation and lowering preserve the path, but actual `EngineReplayRunnerFactory().create(0).run(spec)` raises `InvalidRunnerError` while looking for bundled `systems/data/h200_sxm/vllm/0.24.0/fpm_forward_perf.parquet`. Captured companion kwargs omit the external path. The reproduction uses the repository's AFD performance fixture to isolate the companion; companion estimation and the Rust FPM lookup are real.

| Current | Required |
| --- | --- |
| The regular companion accepts an external pair, then performs the legacy bundled lookup. | Carry the pair to an external-file-aware compiled-engine consumer and test real prefill and decode companion execution with numerical assertions. Simply adding an unsupported keyword to `cli_estimate` would not complete the fix. |

Pure AFD and companions using `op_level` are unaffected.

python/aisimulate/src/aiconfigurator_core/sdk/engine.py
```diff
@@ -422,6 +424,7 @@ def compile_engine(
     shared_layer: bool | None = None,
     transfer_policy: str | list[str] | None = None,
     strict_provenance: bool | None = None,
+    fpm_parquet_path: str | None = None,
```
Arsene12358: **[P2] Update the public API signature contract with the new parameter**

This intentionally adds `fpm_parquet_path` to `compile_engine`, but `python/aisimulate/tests/cross_package/test_core_public_api.py:75-88` still expects its signature to end at `strict_provenance`. The unchanged test fails on the rebuilt head; the declaration and frozen expectation agree on the merge base. `REVIEW.md` requires public Python surfaces and fixtures to change together.

```bash
python -m pytest -p no:timeout -c python/aisimulate/pytest.ini \
  python/aisimulate/tests/cross_package/test_core_public_api.py::test_stable_function_signatures -q
```

Observed: **1 failed**. The assertion differs solely by `, fpm_parquet_path: 'str | None' = None`.

| Current expectation | Required expectation |
| --- | --- |
| `strict_provenance: 'bool \| None' = None) -> 'bytes'` | `strict_provenance: 'bool \| None' = None, fpm_parquet_path: 'str \| None' = None) -> 'bytes'` |

Please update the frozen expectation and rerun the cross-package public API contracts.

Arsene12358 (changes requested):
Request changes on `4a6717415a4a59da874f284820559d15cb031f53`.

The primary external-parquet workflow works: the new synthetic pair drives real replay and CLI details without backend timing directories, invalid configurations are rejected, and the Rust and engine parity suites pass. Three verified findings remain, detailed in the inline comments.

**Spec: two P1 findings**

- Accepted relative paths can select a different external dataset through cache reuse or lazy loading after a working-directory change. A dataset containing a 99 ms prefill returned 22 ms from another pair; its absolute-path control returned 99 ms.
- The regular prefill/decode companion in an AFD+PD deployment accepts an external FPM path but drops it at its final estimator. This affects a companion explicitly configured for external FPM; the AFD attention/FFN workers themselves, pure AFD, and companions using `op_level` are unaffected.

**Standards: one P2 finding**

- The intentional new `compile_engine` parameter is missing from the frozen public API signature expectation, causing `test_stable_function_signatures` to fail.

**Independent validation on the reviewed head**

| Suite | Result |
| --- | --- |
| `cargo test -p aisimulate-core --features python` | 1,458 passed; 1 ignored |
| Config, runner, unified traffic, and sweeper tests | 806 passed; 5 downstream integration cases skipped |
| FPM, memory, KV budget, and cross-package tests | 258 passed; 1 failed, the signature contract |
| `test_engine_step_parity.py` and `test_compile_engine_parity.py` | 383 passed; no golden changes |

Both runtime defects were independently reproduced and reverified with synthetic external parquet/sidecar pairs. Those fixtures establish the input-selection and execution failures, not predictive accuracy against silicon. Tests ran locally on macOS ARM64 with CPython 3.12 and a native extension rebuilt from this commit; pytest used `-p no:timeout` as documented for macOS.

crates/core/src/perfmodel/engine/runtime.rs
```diff
@@ -2546,11 +2570,36 @@ mod tests {
...
+    fn external_fpm_rejects_invalid_specs_before_database_loading() {
+        let tmp = tempfile::tempdir().unwrap();
+        for path in ["", "/missing/reviewed-fpm.parquet"] {
+            let mut config = fixture_engine_config(None);
+            config.fpm_parquet_path = Some(path.into());
+            let spec = EngineSpec::new(config, context_ops(), generation_ops());
+            let err = Engine::from_spec_bytes(&spec.to_bincode().unwrap(), tmp.path()).unwrap_err();
+            assert!(matches!(err, AicError::InvalidEngineConfig(_)), "{err}");
+            assert!(err.to_string().contains("fpm_parquet_path"), "{err}");
+        }
```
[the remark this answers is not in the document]
simone-chen: Fixed in b3089b0e85a485fdcd251b54519a8b0b2deadea0.

Added a Rust engine regression with a valid FPM shape and either a missing parquet or malformed sidecar. Construction succeeds; the first prefill query returns the artifact error. Existing shape-validation coverage remains intact. The full Rust suite passes (1,459 passed, 1 ignored).

docs/core-api.md
```diff
@@ -138,6 +138,42 @@ Use `from_native(...)` instead when native AIC support is required and an
 unsupported configuration or native data failure should surface rather than
 fall back. This strict-native constructor does not take `worker_type`.
 
+For whole-forward prediction, set `forward_model` to `"fpm"` and provide an
+absolute `fpm_parquet_path`. AISimulate reads that file directly; it does not
+require the FPM data to be installed or copied into the repository. The
```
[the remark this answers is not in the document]
simone-chen: Fixed in b3089b0e85a485fdcd251b54519a8b0b2deadea0.

Updated the core API guide to accept absolute or relative paths. Relative paths are anchored to the working directory at engine construction and remain fixed after later directory changes.

README.md
```diff
@@ -13,6 +13,11 @@ configurations offline, without bringing up a GPU serving cluster.
 [FPE Support Matrix](https://ai-dynamo.org/aisimulate/fpe-support-matrix/) ·
 [Legacy AIC Support Matrix](https://ai-dynamo.org/aisimulate/support-matrix/)
 
+Whole-forward FPM data is supplied at runtime rather than shipped in this
+repository. Set `forward_model="fpm"` and pass `fpm_parquet_path` through the
+prediction/recommendation YAML, Python API, or Rust API; see the
```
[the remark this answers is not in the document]
simone-chen: Fixed in b3089b0e85a485fdcd251b54519a8b0b2deadea0.

The README now requires both files and gives the concrete same-stem example `reviewed-fpm.parquet` / `reviewed-fpm.metadata.json`.

crates/core/src/perfmodel/perf_database/mod.rs
simone-chen: Fixed in b3089b0e85a485fdcd251b54519a8b0b2deadea0.

Anchored the external path before shared-table lookup and passed that absolute path to the lazy loader. The explicit table constructor also anchors relative input. Both requested regressions pass with an unchanged absolute systems root: two live engines return their respective 22 ms and 99 ms datasets, and changing directory before the first query preserves the first dataset. Both tests reproduced the incorrect values with the previous native extension.

python/aisimulate/src/aisimulate/compiler.py
simone-chen: Fixed in b3089b0e85a485fdcd251b54519a8b0b2deadea0.

Regular companions with an external pair now use `EngineHandle.compile` and Rust static prediction, preserving the path through the actual consumer. Decode totals are divided by the generated-step count to retain TPOT semantics. Public-config replay tests cover both companion phases and two output lengths, asserting 22 ms prefill / 6 ms decode without a backend data directory. Only the AFD phase measurement is a fixture; companion compilation, loading, prediction, and replay are real. These are synthetic plumbing tests, not silicon accuracy evidence.

python/aisimulate/src/aiconfigurator_core/sdk/engine.py
simone-chen: Fixed in b3089b0e85a485fdcd251b54519a8b0b2deadea0.

Updated the frozen `compile_engine` signature to include `fpm_parquet_path`. All 32 cross-package public API tests pass; together with the FPM suite, 73 tests pass against the rebuilt native extension.

python/aisimulate/src/aisimulate/runner.py
```diff
@@ -221,14 +224,39 @@ def measure(self, spec: ReplaySpec) -> AFDCompanionTiming:
...
+                    key: value
+                    for key, value in kwargs.items()
+                    if key not in {"mode", "backend_name", "isl", "osl", "batch_size"}
+                }
+                engine = EngineHandle.compile(
+                    model_name,
+                    hardware,
+                    deployment.backend,
+                    fpm_parquet_path=fpm_parquet_path,
+                    **compile_kwargs,
```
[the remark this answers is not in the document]
simone-chen: Fixed in e29ced6e, with main merged in 3f748a9b.

- External-FPM companions now use the same AIC alias/validation helper as regular replay. All five accepted dtype overrides map to the compiled API's quantization parameters, and `systems_path` is forwarded. No `attention_backend` replay field was added.
- Real companion replay tests cover prefill and decode, output lengths 2 and 4, plain and `aic_` dtype aliases, a non-default FP8 cell identity, and a system available only under the supplied custom systems root. The environment override is removed, so built-in/default lookup cannot mask a dropped root.
- Updated API documentation to distinguish ReplaySpec overrides from public worker timing YAML.
- Validation: 81 FPM/public API tests, 841 runner/config/AFD/sweeper tests, and 69 speculation/traffic tests passed (five downstream integration-only skips). The rebuilt native extension includes current main. All 1,519 Rust tests passed with one ignored after fixing stale primary-only FP8-block test fixtures in 74a1e011.

python/aisimulate/src/aisimulate/runner.py
```diff
@@ -221,14 +224,50 @@ def measure(self, spec: ReplaySpec) -> AFDCompanionTiming:
         hardware = args.get("aic_system")
         if not isinstance(model_name, str) or not model_name or not isinstance(hardware, str) or not hardware:
             raise ValueError(f"{role} AFD companion requires aic_model_path and aic_system")
+        fpm_parquet_path = args.get("aic_fpm_parquet_path")
```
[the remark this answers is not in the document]
simone-chen: Fixed in 38effc46.

- Normalize AIC aliases once before choosing either the forward model or external parquet path, and read both from the canonical mapping. Removed the redundant prefixed-only validation.
- Extended real external-FPM companion replay coverage to plain `forward_model` and `fpm_parquet_path` together with non-default precision and a custom systems root. All four plain-alias cases failed before the fix and now pass, covering both phases and output lengths 2 and 4.
- Added plain-name coverage for legacy estimator selection and invalid values, plus duplicate-selection-alias rejection before estimation. Updated the API docs.
- Validation: 183 runner/AFD tests and 81 FPM/public API tests passed; Ruff, documentation links, and whitespace checks passed.

Arsene12358 (approved):
All identified review findings are addressed at 09e7ff73df249416caa85ab721f6838521525b49. The original path, companion execution, and API contract fixes are verified. Conflicting companion identity settings now fail before prediction, while matching settings remain accepted.

Validation: 296 focused Python tests passed, including real external-FPM companion replay for both phases. Changed-file and CI-scoped Ruff, formatting, and documentation checks passed. Required CI remains a separate merge gate.

PR #169, opened 2026-09-10
fix(resources): guard host budgets and recover partial sweeps

jasonqinzhou:
Local simulations can allocate more host RAM than is available, and independent worker checks can overcommit the same headroom. This change discovers host/container RAM and CPU capacity automatically, admits the combined memory estimate of each candidate wave, and supervises the owned execution processes throughout initialization, replay and reporting.

No manual limits are required. `execution.resources` is optional; memory settings use decimal GB. Defaults reserve 1 GB for the host and use `max(0, min(0.9 * available_ram, available_ram - 1 GB))` as the execution budget. CPU capacity caps parallel workers and generally leaves one core free. Normal commands perform admission automatically; there is no dry-run option.

- Stream supported JSON/JSONL trace metadata instead of applying fixed 16 MiB/256 KiB cutoffs. Check the conservative file-storage estimate against live headroom before parsing large scalar strings. Unknown allocation models can execute serially under supervision when baseline headroom exists.
- Admit concrete recommendation candidates instead of rejecting the entire domain because its largest load does not fit. Oversized candidates are explicitly `resource_limited`; fitting candidates and their selected prediction files remain available. Result schema 1.1 records resource-limited counts separately from evaluated counts and explicitly upgrades existing 1.0 input.
- Monitor total owned RSS and live headroom, bound initialization/shutdown, and checkpoint completed candidates. On recoverable memory pressure, reap the current workers and retry only unfinished work at lower parallelism, at most twice. Persistent candidate refusal preserves partial results and exits with status 3. A supervisor stop preserves bounded event evidence.
- Preserve existing output artifacts when configuration loading or overrides fail, retain complete runtime diagnostic envelopes for early failures, and use CLI status 1 for signal-derived execution failures.
- Update CLI launchers, installed-package/release contracts, documentation and generated ownership. Reuse the needed supervision foundation from companion draft #170 while preserving the current CLI and power-output contracts.

RSS sampling is best effort, not an operating-system memory sandbox or proof that the original Mac incident is resolved. Spawned SDK execution requires pickleable factories/providers and a guarded script entrypoint.

## Validation

Reviewed head: `fda69bd4bce4dace3a43566ef6b25e51c8c13e65`. The final two commits change documentation only; application code matches `9d69d781`.

Commands below run from the repository root with the isolated Python 3.12 development environment activated. It was installed from the frozen project lock, and the native extension/editable package were rebuilt. `-p no:timeout` follows the repository's macOS guidance; repository warning/error settings remain enabled.

| Evidence | Revision | Command | Result |
|---|---|---|---|
| Local bounded base/head native comparison | `1615509a` → `fda69bd4` | [Reproduction script and machine-readable results](https://github.com/ai-dynamo/aisimulate/pull/169#issuecomment-5722982880), run with the locked development Python | 3 cases: all simulated predictions identical; 1/8/32 request counts conserved; sampled owned RSS below the per-case estimate |
| Local resource, process, and CLI regression | `9d69d781` | `PYTHONPATH=python/aisimulate/src python -m pytest -c pytest.ini -p no:timeout tests/test_host_resources.py tests/test_resource_scheduler.py tests/test_resource_supervision.py tests/test_unified_cli.py -q` | 178 passed |
| Local broad application and complete engine CLI regression | `a6aec88a` | `PYTHONPATH=python/aisimulate/src python -m pytest -c pytest.ini -p no:timeout tests/test_host_resources.py tests/test_resource_scheduler.py tests/test_resource_supervision.py tests/test_unified_cli.py tests/test_cli_config.py tests/test_afd_cli.py tests/test_epd_cli.py tests/sweeper tests/e2e/test_unified_cli_engine.py -q` | 982 passed; 5 optional cross-repository Dynamo contracts skipped |
| Local CI policy contracts | `ecd655b2` | `python -m pytest -c pytest.ini -p no:timeout tests/test_ci_workflow_contracts.py -q` | 189 passed, including real pip-licenses policy checks |
| Local runtime dependency audit, separate clean runtime-only environment | `f7a4210c` | `python scripts/check_python_licenses.py` | Passed; 79 installed runtime/audit packages also pass `uv pip check` |
| Local changed-Python checks | `9d69d781` | `ruff check --config python/aisimulate/pyproject.toml python/aisimulate/src/aisimulate/supervision.py tests/test_resource_supervision.py` and the same paths with `ruff format --check` | Passed |
| Local documentation and whitespace | `fda69bd4` | `python scripts/check_documentation_links.py` and `git diff --check` | Passed |
| Hosted Fast CI | `fda69bd4` | [Run 35289532607](https://github.com/ai-dynamo/aisimulate/actions/runs/35289532607) | Passed |
| Hosted Full CI | `fda69bd4` | [Run 35289797545](https://github.com/ai-dynamo/aisimulate/actions/runs/35289797545) | Passed, including Full CI Success; 54 jobs passed and one lifecycle-only job skipped |

The local resource/CLI suites include fake runners and injected probes for deterministic boundaries, plus real owned-subprocess cleanup and native-engine SDK/CLI cases. The complete engine CLI suite also replays selected recommendation files. These establish software behavior and accounting, not hardware predictive accuracy or calibrated RSS bounds. No production workload or original OS-panic qualification is claimed. Five optional Dynamo contracts were skipped; native golden and prediction-regression evidence belongs to the hosted Full CI run and its stated SHA.

The [allocation provenance](https://github.com/ai-dynamo/aisimulate/blob/fda69bd4bce4dace3a43566ef6b25e51c8c13e65/docs/local-resources.md#allocation-model-provenance) identifies concrete u32 storage separately from engineering allowances. No Rust latency/energy formula or golden answer is changed. Final Codex self-review found no remaining actionable issue. All actionable local review findings are fixed and resolved; the release-scoped response to the linked-repository findings is [recorded here](https://github.com/ai-dynamo/aisimulate/pull/169#issuecomment-5722890084); [CodeRabbit reviewed the final documentation follow-up](https://github.com/ai-dynamo/aisimulate/pull/169#issuecomment-5722932776) with no actionable findings. The automatic report is complete; its remaining suggestions are [assessed with evidence](https://github.com/ai-dynamo/aisimulate/pull/169#issuecomment-5723014711). No actionable issue remains, and all review threads are resolved. CODEOWNER approval is recorded and remains valid under repository rules. Final current-head Fast CI and Full CI are both successful. CODEOWNER approval and resolved conversations were reverified before merge.

Tracks [AIC-1952](https://linear.app/nvidia/issue/AIC-1952). Integrated qualification of the original workload remains separate in #174.


python/aisimulate/src/aisimulate/resources.py
```diff
@@ -0,0 +1,340 @@
...
+            if kind == "cgroup" and not options.intersection(controllers.split(",")):
+                continue
+            mount_root = Path(fields[3].replace("\\040", " "))
+            mount_point = root / fields[4].replace("\\040", " ").lstrip("/")
+            try:
+                relative = Path(member).relative_to(mount_root)
+            except ValueError:
+                # Cgroup namespaces expose membership relative to the mounted root.
+                if member != "/":
+                    continue
```
jasonqinzhou: [P1] Do not discard container limits when the cgroup membership cannot be resolved. A namespace-relative non-root membership falls through this branch, and unreadable membership/mountinfo files also become an empty directory list. Both cases return the unrestricted host snapshot even though the process may be constrained. Resolve namespace-relative descendants conservatively and fail with a resource diagnostic when membership cannot be inspected, with fixtures for both cases.

jasonqinzhou: Fixed in b5e5446e48eb0c7f0a020e010da8f1809e0db78d. Namespace-relative descendants retain ancestor limits; missing/unreadable membership and unresolved mounts now fail closed. Added regression fixtures; the focused suite passes 100 tests.

python/aisimulate/src/aisimulate/resources.py
```diff
@@ -0,0 +1,346 @@
...
+        "isl": source.get("new_input_tokens_per_turn" if session else "input_tokens", 1024),
+        "osl": source.get("output_tokens_per_turn" if session else "output_tokens", 128),
+        "turns_per_session": source.get("session", {}).get("turns", 1),
+        "concurrency": int(concurrency) if concurrency else None,
+        "request_count": count,
+    }
+
+
+def estimate_workload(workload: Mapping[str, Any], *, stack: str, concurrency: int | None = None) -> ResourceEstimate:
+    if workload.get("trace_paths") or workload.get("trace_path"):
```
jasonqinzhou, line 345: [P1] Preserve the existing supported trace and fixed-capacity KV-relative CLI paths. Exact-head Full CI fails ten existing engine CLI cases because all traces and unresolved KV-relative domains are refused. Add bounded trace metadata inspection and a conservative bound for fixed-capacity KV-relative counts, retain refusal for genuinely unqualified inputs, and run the full existing CLI matrix without weakening its assertions.

python/aisimulate/src/aisimulate/resources.py
```diff
@@ -0,0 +1,346 @@
...
+            mount_root = Path(fields[3].replace("\\040", " "))
+            mount_point = root / fields[4].replace("\\040", " ").lstrip("/")
+            try:
+                relative = Path(member).relative_to(mount_root)
+            except ValueError:
+                # Cgroup namespaces expose membership relative to the mounted root.
+                relative = Path(member.lstrip("/"))
+                if ".." in relative.parts:
+                    raise ResourceLimitError("container resource membership is outside its visible namespace") from None
+            directory = mount_point / relative
```
jasonqinzhou, line 105: [P1] Reject parent-directory components after both cgroup membership resolution paths. Path.relative_to and Path.is_relative_to are lexical, so a membership such as /../outside also succeeds against mount root / and can escape the mount during reads. Check the resolved relative parts before traversal, with a fixture proving no host-wide fallback.

python/aisimulate/src/aisimulate/resources.py
jasonqinzhou: Fixed in 267754e121db2f27a685fe80b6ca27507627cf42. Added bounded JSON metadata inspection (including scalar token counts, hash expansion, and cumulative delta/tool turns) and fixed-capacity KV-relative domain bounds. All 125 focused and existing CLI matrix tests pass on macOS; original CLI assertions are unchanged.

python/aisimulate/src/aisimulate/resources.py
jasonqinzhou: Fixed in 267754e121db2f27a685fe80b6ca27507627cf42. Both membership-resolution paths reject parent-directory components before reading cgroup files. Added a root-mount regression fixture.

docs/local-resources.md
```diff
@@ -0,0 +1,93 @@
...
+`aisimulate predict`, `aisimulate recommend`, and the public
+`run_recommendation` Python API check host resources before preparing a run.
+These controls describe the machine running AISimulate. Simulated GPU counts,
+KV-cache memory fractions, traffic concurrency, and request counts retain their
+original meaning.
+
+```yaml
+execution:
+  resources:
+    memory_limit_gib: auto
```
jasonqinzhou: Clarify before this example that no manual resource configuration is required and the entire execution.resources section may be omitted. The schema supplies automatic defaults, but presenting a full YAML block without labeling it optional makes those settings look required. State that AISimulate detects local RAM and CPU capacity automatically, and label this example as optional settings with defaults shown.

jasonqinzhou: Fixed in eefbfe607e0523913e95df59daadf0f5db4cedba. The opening now explicitly says no manual resource configuration is required, explains automatic local RAM/CPU detection and planning, and states that execution.resources can be omitted. The YAML example is labeled optional with defaults shown. Verified against configuration/planner defaults; git diff --check passed.

python/aisimulate/src/aisimulate/main.py
```diff
@@ -72,6 +79,9 @@ def build_parser() -> argparse.ArgumentParser:
         child.add_argument("--output-dir", default="./aisimulate-output")
         child.add_argument("--overwrite", action="store_true")
         child.add_argument("--format", choices=("table", "json"), default="table")
+        child.add_argument(
+            "--dry-run", action="store_true", help="validate core configuration and plan host resources without replay"
```
jasonqinzhou: Remove the optional --dry-run CLI mode as requested. Resource detection, admission checks, and refusal before runner creation already run automatically for predict and recommend. Remove the flag and early-return branches, update the documentation and tests, and retain resource_limited / exit 3 with the refusal diagnostic plan for oversized workloads.

jasonqinzhou: Fixed in 2c11fb63e3079614f40761dec108e85c77a44d70. Removed --dry-run from both CLI commands and their early-return paths. Automatic preflight and resource_limited / exit 3 with diagnostic plan output remain intact. Updated documentation and regression tests. All 64 host-resource/unified CLI tests passed with repository warnings-as-errors settings; Ruff check, formatting, and whitespace checks passed.

python/aisimulate/src/aisimulate/config/common.py
```diff
@@ -108,6 +108,20 @@ class EvaluationConfig(StrictModel):
     sla: SlaConfig | None = None
 
 
+class ResourceConfig(StrictModel):
+    """Execution-host limits, independent of the simulated GPU configuration."""
+
+    memory_limit_gib: PositiveFiniteFloat | Literal["auto"] = "auto"
+    cpu_limit: PositiveStrictInt | Literal["auto"] = "auto"
+    reserve_memory_gib: float = Field(default=2.0, strict=True, ge=0, allow_inf_nan=False)
```
jasonqinzhou: Update the requested default host reserve to exactly 1 GB (1,000,000,000 bytes), with no additional percentage-of-total-RAM reserve by default. Use decimal GB consistently for the optional memory settings and diagnostics, as discussed, rather than relabeling binary values. Keep the separate 90%-of-available-RAM cap and existing CPU worker cap. Document the distinction and verify the reserve, unit conversion, overrides, and refusal boundary.

jasonqinzhou: Fixed in e2157755d970b9ab30d26fff58639dc158a613f7. The default host reserve is exactly 1,000,000,000 bytes and the percentage-of-total-RAM reserve defaults to zero. Optional memory settings and diagnostics now use decimal GB. The independent 90%-available-RAM cap and CPU worker cap remain documented. All 202 resource, CLI and configuration tests passed, including reserve boundaries, large hosts, explicit overrides and exact byte conversions; Ruff and whitespace checks passed.

python/aisimulate/src/aisimulate/resources.py
```diff
@@ -0,0 +1,473 @@
...
+def resource_policy(policy: ResourceConfig):
+    token = _POLICY.set(policy)
+    try:
+        yield
+    finally:
+        _POLICY.reset(token)
+
+
+def guard_replay(spec: Any, *, stack: str, factory: Any = None) -> dict[str, Any]:
+    plan = build_plan(spec.workload, stack=stack, concurrency=spec.concurrency, factory=factory)
```
jasonqinzhou, line 463: [P1] Recheck and reserve the combined candidate wave before dispatch. Per-worker checks can each admit a 6 GB candidate against the same 7 GB live budget after startup, even though the two workers need 12 GB together. Coordinate admission in the parent, account for in-flight reservations, and reduce concurrency when live headroom shrinks.

python/aisimulate/src/aisimulate/resources.py
```diff
@@ -0,0 +1,473 @@
...
+            for path in files:
+                file_count += 1
+                if file_count > 1024:
+                    return unqualified("trace metadata exceeds the bounded 1024-path inspection limit")
+                if path.is_symlink():
+                    return unqualified("trace symlinks have no stable resource identity")
+                if not path.is_file() or (format_name == "weka" and path.suffix.lower() not in {".json", ".jsonl"}):
+                    continue
+                total_bytes += path.stat().st_size
+                if file_count > 1024 or total_bytes > 16 * MIB:
```
jasonqinzhou: [P2] Avoid rejecting otherwise supported traces solely because metadata exceeds the fixed 16 MiB total or 256 KiB record limits. Inspect large traces with bounded-memory streaming and retain validation without allocating the full trace. Unsupported allocation models should have a supervised execution path rather than a blanket refusal when safe monitoring is available.

python/aisimulate/src/aisimulate/recommend.py
```diff
@@ -38,11 +39,21 @@ def run_recommendation(
...
     """Run a public recommendation through the existing Sweeper core."""
 
+    plan = build_plan(
+        workload_bounds(config),
+        stack=stack,
+        policy=config.execution.resources,
+        requested_parallelism=config.optimizer.parallelism,
+        factory=runner_factory,
+    )
+    require_plan(plan)
```
jasonqinzhou: [P2] Admit recommendation candidates individually instead of refusing the whole domain when its largest workload does not fit. Continue evaluating fitting candidates, record resource-limited candidates explicitly, and preserve truthful search coverage rather than silently reporting a complete sweep.

python/aisimulate/src/aisimulate/resources.py
```diff
@@ -0,0 +1,473 @@
...
+@dataclass
+class GuardedRunner:
+    runner: Any
+    stack: str
+    policy: ResourceConfig
+    factory: Any
+
+    def run(self, spec, *, output_requirements=None):
+        with resource_policy(self.policy):
+            guard_replay(spec, stack=self.stack, factory=self.factory)
```
jasonqinzhou, line 477: [P1] Add runtime supervision and bounded recovery for changing memory pressure. A pre-run estimate cannot protect against later growth or competing applications. Monitor owned processes, clean up their descendants on refusal, retain completed evidence, and retry interrupted work at lower concurrency with a finite retry limit.

jasonqinzhou:
@coderabbitai full review

Please review current head 0a45b4f8496076eeecb159f908af714ee9d1baef, including combined live-wave admission, streamed trace estimation, partial candidate results, and bounded supervised recovery. The PR remains draft as requested by the author.

python/aisimulate/src/aisimulate/resources.py
jasonqinzhou: Fixed by 599e514c, present at 0a45b4f8496076eeecb159f908af714ee9d1baef. The coordinator reserves the combined estimates for an entire wave against a fresh host snapshot, including existing child RSS. It shrinks the wave before creating workers, and reaps old workers before admitting replacements. The regression test exercises two 6 GB candidates against less than 8 GB of live capacity.

python/aisimulate/src/aisimulate/resources.py
jasonqinzhou: Fixed by 599e514c, present at 0a45b4f8496076eeecb159f908af714ee9d1baef. JSON/JSONL metadata is parsed through ijson events with a 64 KiB input buffer; total trace and record size caps are removed. A regression fixture exceeds both previous limits without materializing full records. Unqualified estimates run serially only under supervision and with baseline headroom.

python/aisimulate/src/aisimulate/recommend.py
jasonqinzhou: Fixed by 599e514c, present at 0a45b4f8496076eeecb159f908af714ee9d1baef. Removed whole-domain worst-case rejection. Concrete candidates are admitted individually or in fitting waves. Results v1.1 carry resource_limited status/counts separately from evaluated, retain fitting candidates and their selected outputs, and return CLI exit 3 for partial resource coverage. Reading v1.0 explicitly upgrades its envelope.

python/aisimulate/src/aisimulate/resources.py
jasonqinzhou: Fixed by 599e514c, present at 0a45b4f8496076eeecb159f908af714ee9d1baef. Added owned-process RSS/headroom supervision, initialization/shutdown bounds, descendant cleanup and per-candidate checkpoints. Pressure stops and reaps the wave, then retries unfinished candidates at lower parallelism at most twice. Tests cover allocation refusal, live pressure, bounded retries, completed-result retention, resistant descendants, cancellation and the executor reaping race. RSS polling remains best effort and is documented as such.

python/aisimulate/src/aisimulate/resource_scheduler.py
```diff
@@ -0,0 +1,130 @@
...
+        interrupted = None
+        try:
+            active = {pool.submit(evaluate, specs[index]): index for index in wave}
+            started = time.monotonic()
+            initialized = False
+            initialization_timeout = getattr(getattr(factory, "policy", None), "initialization_timeout_seconds", 60.0)
+            deadline = started + timeout if timeout else None
+            while active:
+                if not initialized:
+                    initialized = len(list(Path(ready.name).iterdir())) == len(wave)
```
jasonqinzhou: [P2] Compare readiness with workers actually spawned. ProcessPoolExecutor can reuse an idle worker while submit() is still filling a wave, so the number of processes may be smaller than the candidate count. Requiring one readiness file per candidate can incorrectly time out a healthy long replay. Wait for the actual worker PID set and cover pool reuse in a regression test.

python/aisimulate/src/aisimulate/resources.py
```diff
@@ -0,0 +1,550 @@
...
+            files = source.rglob("*") if source.is_dir() else (source,)
+            for path in files:
+                if path.is_symlink():
+                    return unqualified("trace symlinks have no stable resource identity")
+                if not path.is_file() or (format_name == "weka" and path.suffix.lower() not in {".json", ".jsonl"}):
+                    continue
+                # Parse events rather than loading a document or an entire JSONL
+                # record. Metadata memory depends on nesting, not trace length.
+                maps: list[bool] = []
+                with path.open("rb") as stream:
```
jasonqinzhou, line 304: [P1] Apply the trace storage estimate before parsing metadata. ijson streams records and arrays but still materializes a scalar string, so a trace with a very large string can allocate heavily before build_plan compares its estimate to the budget. Reject a file whose existing conservative storage estimate already exceeds live headroom using stat(), then stream only admitted-size inputs. Add a sparse-file/parser sentinel regression.

jasonqinzhou:
@coderabbitai review

Please review final head fcdb7a153a7789118a37de26e2344a0a4608925f. Since your completed review of 0a45b4f8, this adds two corrections: readiness uses actual spawned worker PIDs, and the trace storage estimate is checked against live headroom before ijson can materialize a large scalar. Both have regression tests; the 59 resource tests pass in the clean locked environment. Please identify any remaining actionable findings on this exact head.

python/aisimulate/src/aisimulate/resource_scheduler.py
jasonqinzhou: Fixed in fcdb7a153a7789118a37de26e2344a0a4608925f. Readiness now compares marker PIDs with the actual executor worker PID set, including reuse of a single ready worker across a larger candidate wave. The regression keeps replay active past the initialization deadline and verifies it succeeds.

python/aisimulate/src/aisimulate/resources.py
jasonqinzhou: Fixed in fcdb7a153a7789118a37de26e2344a0a4608925f. build_plan passes its resolved live headroom to trace inspection; stat-based storage estimates that already exceed it return before invoking ijson. The sparse 128 MiB file sentinel verifies the parser is never called, while the streamed 18 MB corpus with large records remains admitted. The scalar-memory limitation is documented.

jasonqinzhou:
@coderabbitai full review

The incremental request has not produced a review response. Please review exact current head fcdb7a153a7789118a37de26e2344a0a4608925f and report any remaining actionable findings. Your prior full review covered 0a45b4f8; the only subsequent code changes guard trace parsing with the storage estimate and check readiness against actual spawned worker PIDs. The final head passed 950 local tests and Fast CI. Keep the PR draft.

python/aisimulate/pyproject.toml
```diff
@@ -54,6 +54,8 @@ dependencies = [
     "plotly>=6.0.1",
     "prettytable>=3.16.0",
     "pydantic>=2.5",
+    "psutil>=7.0,<8",
+    "ijson>=3.4,<4",
```
jasonqinzhou, line 58: [P1] Include the streaming parser dependency in the Python license policy. The installed ijson 3.5.1 wheel declares BSD-3-Clause AND ISC, but scripts/check_python_licenses.py does not accept that expression, so exact-head Full CI fails Python Dependency Licenses and blocks wheel validation. Verify the packaged ijson/YAJL notices, add the exact expression without exempting the package from auditing, and cover it with a real pip-licenses regression plus a clean runtime dependency audit.

jasonqinzhou: Fixed in ecd655b2bb0e1aa65660c8c5e02b3b6469ad7e7f. Verified the installed ijson/YAJL notices and added only the exact BSD-3-Clause AND ISC expression. All 189 workflow contract tests pass, including real pip-licenses acceptance and GPL-expression rejection. In a clean 79-package runtime environment, the previous policy fails on ijson 3.5.1 and the updated policy passes; dependency compatibility also passes.

jasonqinzhou:
@coderabbitai Please review the current exact head `ecd655b2bb0e1aa65660c8c5e02b3b6469ad7e7f` before Full CI is restarted. Your previous complete review found no actionable issues on `fcdb7a153a7789118a37de26e2344a0a4608925f`. The only subsequent change adds the installed ijson/YAJL license expression `BSD-3-Clause AND ISC` to the Python policy and extends the real pip-licenses regression for PEP 639 acceptance and GPL-expression rejection.

Codex reviewed the final two-file diff and found no remaining actionable issue. All 189 workflow contract tests pass. A clean 79-package runtime audit reproduced the old failure specifically on ijson 3.5.1, passes with the new policy, and passes dependency compatibility. The new review thread is fixed and resolved. Please confirm this exact current head has no remaining actionable findings, or report any concrete issues.

python/aisimulate/src/aisimulate/resource_scheduler.py
```diff
@@ -0,0 +1,132 @@
...
+    pending = list(range(len(specs)))
+    attempts = dict.fromkeys(pending, 0)
+    capacity = max(1, workers)
+    while pending:
+        wave = pending[:capacity]
+        while True:
+            try:
+                plan = factory.admit_wave([specs[index] for index in wave])
+                checkpoint("wave_admitted", plan)
+                break
```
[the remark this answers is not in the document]
jasonqinzhou: Confirmed on the current head: the admission catch also intercepts checkpoint-budget failures. I will move checkpointing outside that catch and add a sentinel regression proving that an evidence failure stops before any worker starts.

python/aisimulate/src/aisimulate/supervision.py
```diff
@@ -0,0 +1,460 @@
...
+    parser = build_parser()
+    args = parser.parse_args(arguments)
+    raw = None
+    try:
+        raw = _load_mapping(args.config)
+        _apply_overrides(raw, args.overrides, command=args.command)
+        policy = ResourceConfig.model_validate(raw.get("execution", {}).get("resources", {}))
+    except (OSError, ValueError, AttributeError):
+        # The child retains stack/schema error ordering under conservative limits.
+        policy = ResourceConfig()
```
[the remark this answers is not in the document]
jasonqinzhou: Confirmed: missing/malformed config and failed overrides can reach parent output preparation. I will skip output and diagnostic writes when parent preparation fails, retain child-side error ordering, and test existing-artifact preservation through the public CLI.

python/aisimulate/src/aisimulate/supervision.py
```diff
@@ -0,0 +1,460 @@
...
+        parser.error(str(exc))
+    event_output = output / "execution-events.jsonl"
+    try:
+        report = run_process(
+            [sys.executable, "-m", "aisimulate.resource_worker", "cli", *arguments],
+            policy=policy,
+            events_output=str(event_output),
+        )
+    except ResourceLimitError as exc:
+        report = exc.plan
```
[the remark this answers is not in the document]
jasonqinzhou: Confirmed: an early discovery/budget failure emits an incomplete runtime envelope. I will supply explicit unavailable values for unobserved fields and preserve any detailed fields already present in the exception plan.

python/aisimulate/src/aisimulate/supervision.py
```diff
@@ -0,0 +1,460 @@
...
+        _save_runtime_report(args.output_dir, report, overwrite=args.overwrite)
+    except (OSError, ValueError) as exc:
+        sys.stderr.write(f"could not save resource runtime report: {exc}\n")
+    if report["status"] == "resource_limited":
+        return 3
+    if report["status"] == "cancelled":
+        return 130
+    if report["status"] == "timed_out":
+        return 124
+    return int(report.get("exit_code") or 0)
```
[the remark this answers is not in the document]
jasonqinzhou: Confirmed: negative subprocess signal codes currently escape as the CLI return value. I will retain the raw code in diagnostics and return execution-failure status 1 from the CLI, with a regression for signal and positive error codes.

scripts/build_release_artifacts.py
```diff
@@ -84,7 +84,7 @@ def check_manifests() -> tuple[str, str]:
     assert not any(str(dep).lower().startswith(("dynamo", "ai-dynamo")) for dep in dependencies)
     assert app["scripts"] == {
         "aiconfigurator": "aiconfigurator.main:main",
-        "aisimulate": "aisimulate.main:main",
+        "aisimulate": "aisimulate.supervision:main",
```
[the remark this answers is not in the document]
jasonqinzhou: Confirmed the migration guide still denies the installed aisimulate command. I will synchronize that passage with both current entry points and link to the installation guide for publication status; no version or dependency-pin change is needed.

tests/test_resource_supervision.py
```diff
@@ -0,0 +1,419 @@
...
+        Path("tests/e2e/configs/unified_cli/recommend/engine/01-default-preset-throughput.yaml").read_text()
+    )
+    raw["traffic"] = {
+        "source": {"type": "synthetic", "input_tokens": 10240, "output_tokens": 1024},
+        "load": {"type": "concurrency", "concurrency": 64512},
+        "stop": {"requests_per_load_unit": 100},
+    }
+    raw["optimizer"].update(max_trials=1, parallelism=1)
+    config = CoreRecommendationConfig.model_validate(raw)
+    result = run_recommendation(config, stack="engine", runner_factory=EngineReplayRunnerFactory(), show_progress=False)
```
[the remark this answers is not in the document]
jasonqinzhou: Confirmed the oversized SDK test relies on automatic host capacity. I will set an explicit 2 GB budget before calling the public API so the huge candidate remains inadmissible on high-memory runners.

python/aisimulate/src/aisimulate/supervision.py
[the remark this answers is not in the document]
tedzhouhk: [P2] Independently reproduced on ecd655b2: with an existing output directory and `--overwrite`, a missing config file, malformed YAML, and a malformed `--set` assignment each exit with status 2 after deleting the previous `prediction.json` and `execution-events.jsonl`. The same three public CLI invocations on base 1615509a preserve both artifacts.

I think it is better to gate all output cleanup and diagnostic writes on successful configuration loading and override application. Please include the malformed-override case in the regression coverage too: `raw` can remain non-null and pass core validation even though `_apply_overrides` failed. This adds the runtime/base comparison to the existing finding.

tedzhouhk (approved):
Reviewed ecd655b2bb0e1aa65660c8c5e02b3b6469ad7e7f, including host-budget admission, combined candidate waves, bounded retries, process cleanup, and partial-result output.

Validation: rebuilt the native extension and installed the frozen lock in an isolated environment; 833 focused resource/CLI/config/Sweeper tests and all 13 engine prediction CLI cases passed. Five cross-repository Dynamo contract tests were skipped. A real supervised sweep retained 3 feasible candidates and their prediction files alongside 4 resource-limited candidates and 1 cache hit, and exited with status 3.

Approving the reviewed implementation, with the acknowledged output-preservation P2 tracked in the existing inline thread. Added independent base/head reproduction evidence for missing config, malformed YAML, and malformed overrides: https://github.com/ai-dynamo/aisimulate/pull/169#discussion_r4042177169.

python/aisimulate/src/aisimulate/resource_scheduler.py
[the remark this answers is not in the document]
jasonqinzhou: Fixed in a6aec88a2d24286529b51310eab991d7161a76ff. Checkpoint failure now propagates outside the admission retry handler. A sentinel test confirms no worker starts and no candidate is reclassified. The complete resource/CLI/Sweeper and engine CLI suite passed: 982 tests, with five expected optional Dynamo skips. Lint, formatting, documentation links, and whitespace checks pass.

python/aisimulate/src/aisimulate/supervision.py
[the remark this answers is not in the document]
jasonqinzhou: Fixed in a6aec88a2d24286529b51310eab991d7161a76ff. Parent output/event/diagnostic writes are skipped when config loading, overrides, or resource validation fail. Eight public CLI regressions preserve existing prediction/runtime artifacts and stack-before-config error ordering. The complete resource/CLI/Sweeper and engine CLI suite passed: 982 tests, with five expected optional Dynamo skips. Lint, formatting, documentation links, and whitespace checks pass.

python/aisimulate/src/aisimulate/supervision.py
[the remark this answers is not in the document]
jasonqinzhou: Fixed in a6aec88a2d24286529b51310eab991d7161a76ff. Early resource failures now include the runtime envelope, requested policy, and null values for unobserved fields. Four regressions cover both probes and preserve detailed exception-plan fields. The complete resource/CLI/Sweeper and engine CLI suite passed: 982 tests, with five expected optional Dynamo skips. Lint, formatting, documentation links, and whitespace checks pass.

python/aisimulate/src/aisimulate/supervision.py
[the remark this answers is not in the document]
jasonqinzhou: Fixed in a6aec88a2d24286529b51310eab991d7161a76ff. Negative child signal codes now return CLI status 1; raw codes remain in the diagnostic. Regressions cover SIGKILL/SIGTERM codes and preserve positive config-error status 2. The complete resource/CLI/Sweeper and engine CLI suite passed: 982 tests, with five expected optional Dynamo skips. Lint, formatting, documentation links, and whitespace checks pass.

scripts/build_release_artifacts.py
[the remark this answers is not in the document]
jasonqinzhou: Fixed in a6aec88a2d24286529b51310eab991d7161a76ff. The migration guide now documents both installed commands and automatic supervision, with a link to current installation/publication guidance. Versions and Dynamo pins are unchanged. The complete resource/CLI/Sweeper and engine CLI suite passed: 982 tests, with five expected optional Dynamo skips. Lint, formatting, documentation links, and whitespace checks pass.

tests/test_resource_supervision.py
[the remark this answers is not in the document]
jasonqinzhou: Fixed in a6aec88a2d24286529b51310eab991d7161a76ff. The oversized SDK test now supplies an explicit 2 GB budget, so high-memory hosts cannot admit the enormous candidate. The complete resource/CLI/Sweeper and engine CLI suite passed: 982 tests, with five expected optional Dynamo skips. Lint, formatting, documentation links, and whitespace checks pass.

jasonqinzhou:
@coderabbitai Please review exact current head `a6aec88a2d24286529b51310eab991d7161a76ff`, including the complete first-parent correction from `ecd655b2bb0e1aa65660c8c5e02b3b6469ad7e7f`, before final Full CI dispatch.

This fixes all six findings from your automatic review: checkpoint failures propagate outside admission retries; invalid configuration/overrides preserve existing outputs while retaining child-side stack error ordering; early resource reports have explicit null fields plus the requested policy and retain existing detailed plans; signal exits map to CLI failure 1 while diagnostics retain the raw code; migration documentation names both installed commands; the oversized SDK test uses an explicit 2 GB limit. Each original thread has a fix reply and is resolved.

Codex re-read the correction and surrounding error paths and found no remaining actionable issues. The final local resource, CLI, config, AFD/EPD, Sweeper, and complete engine CLI suite passed 982 tests, with five expected optional Dynamo skips. Lint, formatting, documentation links, and whitespace pass. Full CI passed on the prior head; a new run will validate this exact commit after review. Please confirm no actionable findings on this head, or identify concrete remaining defects.

python/aisimulate/docs/aisimulate_migration.md
```diff
@@ -17,10 +17,12 @@ python -m pip uninstall -y aiconfigurator aiconfigurator-core
 python -m pip install "aisimulate==0.12.0"
 ```
 
-The `aisimulate` wheel contains the complete application and installs the
-established `aiconfigurator` command. It does not install a second top-level
-application command named `aisimulate`. The command name is independent of the
-distribution name and does not represent a separately published AIC artifact.
+The `aisimulate` wheel contains the complete application and installs both the
```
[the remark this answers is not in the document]
jasonqinzhou: Confirmed the existing stable-version install example is misleading before publication. I will replace that example with installation from the repository source, explicitly link the source prerequisites and published-version guide, and remove the second stable-wheel installation claim. Publishing a release or changing Dynamo pins is outside this PR; the source install provides the launcher being documented without claiming a package has shipped.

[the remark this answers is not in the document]
jasonqinzhou: Fixed in 524debc98b5717f8f1a508e04f34b53c99a1aeae. The guide now explicitly uses a current source checkout and the same uv sync / environment activation sequence as docs/installation.md. Both stable-wheel install claims were removed. No release publication or cross-repository pin change is implied. Documentation links and whitespace checks pass; application code is unchanged from the 982-test validated head.

jasonqinzhou:
@coderabbitai Please review exact current head `524debc98b5717f8f1a508e04f34b53c99a1aeae`. The only change from your completed review of `a6aec88a2d24286529b51310eab991d7161a76ff` fixes its one remaining finding: the migration guide uses the source-install flow from docs/installation.md instead of pinning an unpublished stable wheel. Both stable-wheel installation claims are removed. Source prerequisites are linked; no release or Dynamo pin change is claimed. The original thread is answered and resolved.

Codex reviewed the final one-file diff and found no remaining actionable issue. Documentation links and whitespace pass. Application code is unchanged from the head that passed all 982 focused and complete engine CLI tests. Please confirm no actionable findings on this exact head, or report any concrete remaining issue, so we can dispatch final Full CI.

python/aisimulate/src/aisimulate/supervision.py
```diff
@@ -0,0 +1,477 @@
...
+        report = run_process(
+            [sys.executable, "-m", "aisimulate.resource_worker", "cli", *arguments],
+            policy=policy,
+            events_output=event_output,
+        )
+    except ResourceLimitError as exc:
+        report = {
+            "schema_version": 1,
+            "status": "resource_limited",
+            "reason": str(exc),
```
jasonqinzhou, line 401: [P2] Confirmed the latest CodeRabbit outside-diff finding: pre-spawn discovery/budget refusals return exit 3 and write resource-runtime.json but omit resource-plan.json. Persist the refusal plan before returning; retain the known host snapshot and resolved budget where available, and use null for unavailable host/budget/estimate fields. Keep invalid-config paths free of output writes and preserve detailed exception plans. Add public CLI regressions for these pre-spawn failures.

jasonqinzhou: Fixed in 9d69d781b912382275a30d40c7da7d00a63c97c5. Pre-spawn refusal now saves resource-plan.json and resource-runtime.json. The preflight exception retains a known host snapshot and resolved budget; unavailable fields remain null, and detailed exception data is preserved. Tests cover discovery failure, resolver failure, an insufficient budget before spawn, and preservation of invalid-input outputs. All 178 resource/scheduler/supervision/unified CLI tests pass, plus lint, formatting, documentation links, and whitespace.

jasonqinzhou:
@coderabbitai Please review exact head `9d69d781b912382275a30d40c7da7d00a63c97c5`. It fixes the outside-diff P2 from your completed review of `524debc98b5717f8f1a508e04f34b53c99a1aeae`: pre-spawn refusals now persist resource-plan.json as well as the runtime report, with a known host/budget retained and unavailable fields null. Invalid-input paths still skip all output writes. The finding is tracked and resolved at https://github.com/ai-dynamo/aisimulate/pull/169#discussion_r4042302587.

All 178 focused resource, scheduler, supervision, and unified CLI tests pass. Lint, formatting, documentation links, and whitespace pass. Codex reviewed the three-file correction and found no remaining actionable issues. Please confirm no actionable findings on this exact head or identify a concrete issue; final Full CI and the authorized squash merge follow the completed review.

jasonqinzhou:
Codex confirmed the final checklist finding on `9d69d781b912382275a30d40c7da7d00a63c97c5` (the affected CLI pages are outside the current diff):

[P2] Synchronize all CLI contract pages with resource-limited outcomes. The guide currently says every no-feasible recommendation exits 1 and omits resource_limited rows, exit 3, and the new diagnostic files from output/overwrite descriptions. Update this guide, the migration walkthrough, and the prediction interpretation guide together; retain the separate prediction-details schema version. Also link the local-resource policy and clearly identify its heuristic allowances so documentation does not imply calibrated RSS or changed latency models.

jasonqinzhou:
Fixed the confirmed CLI documentation finding in `f7a4210c30948b79ba24925551c452fc58b129a2` (original discussion: https://github.com/ai-dynamo/aisimulate/pull/169#issuecomment-5722785752). The user guide, migration walkthrough, and prediction interpretation guide now agree on resource-limited results, exit 3, diagnostic files, overwrite behavior, and optional execution settings. Prediction-details schema examples retain version 1.0. Documentation links and whitespace pass; application code is unchanged from `9d69d781`.

The allocation-provenance section now identifies concrete u32 token storage and explicitly labels every extra multiplier/baseline as an engineering allowance, not a measured object size or calibrated RSS model. The generic performance-model evidence check does not establish a missing prediction-accuracy gate here: this PR changes host admission/accounting, not the Rust latency/energy oracle. Boundary/sentinel/process tests establish the stated control-flow invariants; the existing prediction/golden jobs verify modeled output stability. Original production-workload qualification remains explicitly separate in #174, and no calibrated RSS or OS-panic resolution is claimed. Ted's independent review also reports a real partial sweep retaining fitting candidates alongside refused ones.

@coderabbitai Please review exact head `f7a4210c30948b79ba24925551c452fc58b129a2`. Your completed review of `9d69d781` generated no actionable comments; this four-file documentation follow-up addresses the final checklist clarification. Codex reviewed the documentation against the implementation and found no remaining actionable issue. Final Full CI and the authorized squash merge await current-head review.

jasonqinzhou:
Validated the remaining local documentation findings on `f7a4210c`: the troubleshooting row in `docs/cli/user-guide.md` and the Dynamo deployment walkthrough still imply that every empty recommendation exits 1. The actual consumer returns 3 whenever `counts.resource_limited` is nonzero, whether or not a configuration was selected. I will synchronize these passages and make the three already-updated guides state both conditions explicitly.

The two linked-repository findings need to remain tied to the release they describe. This PR neither publishes/replaces the existing `aisimulate==0.12.0` artifacts nor changes Dynamo's pinned dependency. Its new runtime dependencies are declared in this package, and the local migration guide deliberately uses a source checkout. The frozen AIConfigurator guidance about the already-published 0.12.0 wheel must not claim a launcher added only by this unreleased PR. When Dynamo adopts a release containing this change, its `--no-deps` image installation must explicitly supply both `psutil` and `ijson`; I will record that integration requirement in the local resource guide. These are not current-breakage findings against the pinned downstream release.

docs/cli/migrate-from-aiconfigurator.md
```diff
@@ -173,8 +173,11 @@ aisimulate predict \
 ```
 
 Check `budget-selected/prediction.json` for latency and throughput under the saved workload.
-If the search finds no feasible candidate, it writes `recommendation.json`, exits with status 1,
-and produces no selected YAML; inspect that report before running the prediction command.
+If the completed search has no feasible or resource-limited candidate, it writes
+`recommendation.json`, exits with status 1, and produces no selected YAML. Resource-limited
+candidates instead produce exit 3, including when fitting candidates and selected YAML remain
+available. Inspect the ledger and [resource diagnostics](../local-resources.md) before predicting
+a selected configuration or treating the search as complete.
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in fda69bd4bce4dace3a43566ef6b25e51c8c13e65. All three passages now explicitly require no selected configuration AND zero resource-limited candidates for status 1. The troubleshooting table and Dynamo deployment walkthrough use the same rule; status 3 covers any resource-limited candidate. Local documentation links and whitespace checks passed.

jasonqinzhou:
@coderabbitai review

Please review final head `fda69bd4bce4dace3a43566ef6b25e51c8c13e65`. The last change is documentation only. The remaining local exit-code inconsistencies and ambiguous wording are fixed in all affected guides. The runtime image integration requirement is now explicit in `docs/local-resources.md`; the release-scoped response to the two outside-diff linked-repository findings is at https://github.com/ai-dynamo/aisimulate/pull/169#issuecomment-5722890084. No downstream version pin or published artifact changes here.

Codex self-review of this head found no further actionable issue. Documentation-link and whitespace checks passed; application behavior remains at `9d69d781` with the previously recorded native and subprocess regression evidence. Resource allocation coefficients are explicitly identified as engineering allowances; no predictive RSS calibration or original-workload qualification is claimed. The available independent native sweep and regression evidence is in the human review and PR validation table. Full CI will be dispatched against this exact SHA after this review completes.

jasonqinzhou:
@coderabbitai resume

jasonqinzhou:
Additional bounded native evidence for `fda69bd4bce4dace3a43566ef6b25e51c8c13e65`:

Compared the public engine CLI at base `1615509a` and this head on three fixed synthetic workloads (1, 8, and 32 completed requests). Every simulated prediction field was identical. Only `wall_time_ms`, `processed_tokens_per_s`, and `processed_output_tokens_per_s` were excluded from equality because they measure actual replay wall time; their values remain in the JSON below. There is no claimed speedup.

An external observer sampled the owned process tree every 10 ms, independently of the new supervisor. Observed RSS was 128–133 MB at base and 169–170 MB at head. The head's process count increased from one to two, as expected for the added supervisor. All three observations were below the corresponding 537 MB worker estimate. This is a bounded synthetic comparison, not calibration, a hard RSS bound, or qualification of the original failing workload. Sampling can miss short peaks. The runs had a separate 1.5 GB / 60 second stop guard.

The artifact includes per-case input, host snapshots, budgets, estimates, admission results, counts, and measured RSS. Assertions check finite values, nonnegative ordered bounds, estimate monotonicity, request-count conservation, and unchanged simulated predictions. Recommendation selection/partial-result evidence remains separately recorded in [the independent human review](https://github.com/ai-dynamo/aisimulate/pull/169#pullrequestreview-5242322465) and the regression tests; these three cases exercise prediction.

Reproduce from this exact checkout with its locked Python 3.12 development environment: save the script below as `/tmp/pr169-resource-comparison.py`, then run `python /tmp/pr169-resource-comparison.py`. Base Python orchestration is extracted with `git archive`; both revisions use the same built native extension, whose hash is recorded. The script asserts that the PR changes no Rust source/manifests/lockfile. Output is written to a new temporary directory, without editing the checkout.

<details><summary>Machine-readable comparison</summary>

```json
{
  "base": "1615509afe61c2072a49769b9e5fee082f753836",
  "head": "fda69bd4bce4dace3a43566ef6b25e51c8c13e65",
  "platform": "macOS-26.7-arm64-arm-64bit",
  "python": "3.12.11",
  "native_extension_shared": true,
  "native_extension_sha256": "7ad3edff5f229017cf30de70d1b403e6143871d7bc284c577e8d8d51f0347be9",
  "sampling_interval_seconds": 0.01,
  "external_stop_bytes": 1500000000,
  "cases": [
    {
      "concurrency": 1,
      "requests_per_load_unit": 1,
      "input_tokens": 16,
      "output_tokens": 8,
      "base": {
        "exit_status": 0,
        "externally_sampled_owned_peak_rss_bytes": 132857856,
        "maximum_observed_processes": 1,
        "completed_requests": 1
      },
      "head": {
        "exit_status": 0,
        "externally_sampled_owned_peak_rss_bytes": 169443328,
        "maximum_observed_processes": 2,
        "completed_requests": 1
      },
      "plan": {
        "schema_version": 1,
        "status": "admitted",
        "stack": "engine",
        "host": {
          "total_memory_bytes": 38654705664,
          "available_memory_bytes": 11556306944,
          "cpu_count": 12.0,
          "process_memory_bytes": 121602048
        },
        "budget": {
          "memory_limit_bytes": 10407193804,
          "cpu_limit": 11,
          "reserved_host_memory_bytes": 1000000000,
          "coordinator_memory_bytes": 430407680
        },
        "estimate": {
          "allocation_model": "engine-session-metadata-v1",
          "request_count": 1,
          "input_token_bytes": 64,
          "lower_bound_bytes": 32,
          "estimated_peak_bytes": 536875840,
          "reason": "",
          "api_version": 1
        },
        "requested_parallelism": 1,
        "effective_parallelism": 1,
        "reason": ""
      },
      "measured_wall_fields": {
        "base": {
          "processed_output_tokens_per_s": 3635.055016557676,
          "processed_tokens_per_s": 10905.165049673027,
          "wall_time_ms": 2.200792
        },
        "head": {
          "processed_output_tokens_per_s": 11573.236889692585,
          "processed_tokens_per_s": 34719.71066907776,
          "wall_time_ms": 0.69125
        }
      },
      "prediction_summaries_identical": true
    },
    {
      "concurrency": 2,
      "requests_per_load_unit": 4,
      "input_tokens": 128,
      "output_tokens": 8,
      "base": {
        "exit_status": 0,
        "externally_sampled_owned_peak_rss_bytes": 128401408,
        "maximum_observed_processes": 1,
        "completed_requests": 8
      },
      "head": {
        "exit_status": 0,
        "externally_sampled_owned_peak_rss_bytes": 169934848,
        "maximum_observed_processes": 2,
        "completed_requests": 8
      },
      "plan": {
        "schema_version": 1,
        "status": "admitted",
        "stack": "engine",
        "host": {
          "total_memory_bytes": 38654705664,
          "available_memory_bytes": 11483742208,
          "cpu_count": 12.0,
          "process_memory_bytes": 121438208
        },
        "budget": {
          "memory_limit_bytes": 10331239219,
          "cpu_limit": 11,
          "reserved_host_memory_bytes": 1000000000,
          "coordinator_memory_bytes": 430718976
        },
        "estimate": {
          "allocation_model": "engine-session-metadata-v1",
          "request_count": 8,
          "input_token_bytes": 1024,
          "lower_bound_bytes": 256,
          "estimated_peak_bytes": 536939520,
          "reason": "",
          "api_version": 1
        },
        "requested_parallelism": 1,
        "effective_parallelism": 1,
        "reason": ""
      },
      "measured_wall_fields": {
        "base": {
          "processed_output_tokens_per_s": 46840.70427926428,
          "processed_tokens_per_s": 796291.9727474927,
          "wall_time_ms": 1.366333
        },
        "head": {
          "processed_output_tokens_per_s": 49597.985081856044,
          "processed_tokens_per_s": 843165.7463915528,
          "wall_time_ms": 1.290375
        }
      },
      "prediction_summaries_identical": true
    },
    {
      "concurrency": 4,
      "requests_per_load_unit": 8,
      "input_tokens": 512,
      "output_tokens": 8,
      "base": {
        "exit_status": 0,
        "externally_sampled_owned_peak_rss_bytes": 128434176,
        "maximum_observed_processes": 1,
        "completed_requests": 32
      },
      "head": {
        "exit_status": 0,
        "externally_sampled_owned_peak_rss_bytes": 170393600,
        "maximum_observed_processes": 2,
        "completed_requests": 32
      },
      "plan": {
        "schema_version": 1,
        "status": "admitted",
        "stack": "engine",
        "host": {
          "total_memory_bytes": 38654705664,
          "available_memory_bytes": 11498733568,
          "cpu_count": 12.0,
          "process_memory_bytes": 121962496
        },
        "budget": {
          "memory_limit_bytes": 10322819481,
          "cpu_limit": 11,
          "reserved_host_memory_bytes": 1000000000,
          "coordinator_memory_bytes": 431095808
        },
        "estimate": {
          "allocation_model": "engine-session-metadata-v1",
          "request_count": 32,
          "input_token_bytes": 8192,
          "lower_bound_bytes": 1024,
          "estimated_peak_bytes": 537542656,
          "reason": "",
          "api_version": 1
        },
        "requested_parallelism": 1,
        "effective_parallelism": 1,
        "reason": ""
      },
      "measured_wall_fields": {
        "base": {
          "processed_output_tokens_per_s": 55360.83151968942,
          "processed_tokens_per_s": 3598454.0487798126,
          "wall_time_ms": 4.624208
        },
        "head": {
          "processed_output_tokens_per_s": 51427.13307594321,
          "processed_tokens_per_s": 3342763.649936309,
          "wall_time_ms": 4.977917
        }
      },
      "prediction_summaries_identical": true
    }
  ],
  "qualification": "bounded synthetic native CLI comparison; not RSS calibration or production qualification",
  "anomaly_checks": {
    "finite_values": true,
    "nonnegative_ordered_bounds": true,
    "estimate_monotonicity": true,
    "request_count_conservation": true,
    "unchanged_prediction_summaries": true
  }
}
```

</details>

<details><summary>Reproduction script</summary>

```python
"""Bounded native CLI comparison; run with the PR's locked development Python."""
import io
import hashlib
import json
import math
import os
from pathlib import Path
import platform
import signal
import subprocess
import sys
import tarfile
import tempfile
import time

import psutil
import yaml

ROOT = Path.cwd()
BASE = "1615509afe61c2072a49769b9e5fee082f753836"
HEAD = "fda69bd4bce4dace3a43566ef6b25e51c8c13e65"
assert subprocess.check_output(["git", "rev-parse", "HEAD"], text=True).strip() == HEAD
OUT = Path(tempfile.mkdtemp(prefix="pr169-resource-comparison-"))
archive = subprocess.check_output(["git", "archive", BASE, "python/aisimulate/src/aisimulate"])
with tarfile.open(fileobj=io.BytesIO(archive)) as bundle:
    bundle.extractall(OUT / "base", filter="data")
assert not subprocess.check_output(["git", "diff", "--name-only", f"{BASE}..{HEAD}", "crates", "Cargo.toml", "Cargo.lock"])
runtime = next((ROOT / "python/aisimulate/src/aisimulate").glob("_runtime*.so"))
(OUT / "base/python/aisimulate/src/aisimulate" / runtime.name).symlink_to(runtime)
template = yaml.safe_load((ROOT / "tests/e2e/configs/unified_cli/predict/engine/01-synthetic-default.yaml").read_text())
report = {"base": BASE, "head": HEAD, "platform": platform.platform(), "python": platform.python_version(),
          "native_extension_shared": True, "native_extension_sha256": hashlib.sha256(runtime.read_bytes()).hexdigest(),
          "sampling_interval_seconds": 0.01, "external_stop_bytes": 1500000000,
          "cases": [], "qualification": "bounded synthetic native CLI comparison; not RSS calibration or production qualification"}

def finite(value):
    if isinstance(value, dict):
        return all(finite(v) for v in value.values())
    if isinstance(value, list):
        return all(finite(v) for v in value)
    return not isinstance(value, float) or math.isfinite(value)

for concurrency, per_unit, isl in [(1, 1, 16), (2, 4, 128), (4, 8, 512)]:
    case = {"concurrency": concurrency, "requests_per_load_unit": per_unit, "input_tokens": isl, "output_tokens": 8}
    config = dict(template)
    config["traffic"] = {"source": {"type": "synthetic", "input_tokens": isl, "output_tokens": 8},
                         "load": {"type": "concurrency", "concurrency": concurrency},
                         "stop": {"requests_per_load_unit": per_unit}}
    path = OUT / f"case-{concurrency}.yaml"
    path.write_text(yaml.safe_dump(config))
    summaries = {}
    for label in ("base", "head"):
        source = OUT / "base/python/aisimulate/src" if label == "base" else ROOT / "python/aisimulate/src"
        env = dict(os.environ, PYTHONPATH=str(source), OMP_NUM_THREADS="1", OPENBLAS_NUM_THREADS="1", MKL_NUM_THREADS="1")
        output = OUT / f"{label}-{concurrency}"
        command = [sys.executable, "-m", "aisimulate", "predict", "--stack", "engine", "--config", str(path),
                   "--output-dir", str(output), "--format", "json"]
        with (OUT / f"{label}-{concurrency}.stdout").open("w+") as stdout, (OUT / f"{label}-{concurrency}.stderr").open("w+") as stderr:
            child = subprocess.Popen(command, env=env, stdout=stdout, stderr=stderr, start_new_session=True)
            peak, max_processes = 0, 0
            started = time.monotonic()
            try:
                while child.poll() is None:
                    try:
                        parent = psutil.Process(child.pid)
                        owned = [parent, *parent.children(recursive=True)]
                        sizes = []
                        for process in owned:
                            try:
                                sizes.append(process.memory_info().rss)
                            except psutil.NoSuchProcess:
                                pass
                        peak = max(peak, sum(sizes))
                        max_processes = max(max_processes, len(sizes))
                    except psutil.NoSuchProcess:
                        pass
                    if peak > 1500000000 or time.monotonic() - started > 60:
                        raise RuntimeError("bounded comparison exceeded external guard")
                    time.sleep(0.01)
                stdout.seek(0); stderr.seek(0)
                assert child.returncode == 0, (label, child.returncode, stderr.read())
                summaries[label] = json.loads(stdout.read())
            finally:
                if child.poll() is None:
                    os.killpg(child.pid, signal.SIGKILL)
                    child.wait(timeout=10)
        case[label] = {"exit_status": child.returncode, "externally_sampled_owned_peak_rss_bytes": peak,
                       "maximum_observed_processes": max_processes, "completed_requests": summaries[label]["completed_requests"]}
        assert peak > 0 and finite(summaries[label])
        if label == "head":
            events = [json.loads(line) for line in (output / "execution-events.jsonl").read_text().splitlines()]
            plan = next(event["value"] for event in events if event["event"] == "resource_plan")
            case["plan"] = plan
            assert finite(plan)
            assert plan["estimate"]["estimated_peak_bytes"] >= plan["estimate"]["lower_bound_bytes"] >= 0
    wall_fields = {"wall_time_ms", "processed_tokens_per_s", "processed_output_tokens_per_s"}
    case["measured_wall_fields"] = {label: {key: value for key, value in summary.items() if key in wall_fields}
                                    for label, summary in summaries.items()}
    predictions = {label: {key: value for key, value in summary.items() if key not in wall_fields}
                   for label, summary in summaries.items()}
    case["prediction_summaries_identical"] = predictions["base"] == predictions["head"]
    assert case["prediction_summaries_identical"], summaries
    assert case["head"]["completed_requests"] == concurrency * per_unit
    report["cases"].append(case)
peaks = [c["plan"]["estimate"]["estimated_peak_bytes"] for c in report["cases"]]
assert peaks == sorted(peaks)
report["anomaly_checks"] = {"finite_values": True, "nonnegative_ordered_bounds": True, "estimate_monotonicity": True,
                             "request_count_conservation": True, "unchanged_prediction_summaries": True}
(OUT / "comparison.json").write_text(json.dumps(report, indent=2, allow_nan=False) + "\n")
print(OUT / "comparison.json")
print(json.dumps({"cases": [{k: v for k, v in c.items() if k != "plan"} for c in report["cases"]], "anomaly_checks": report["anomaly_checks"]}, indent=2))

```

</details>

docs/local-resources.md
```diff
@@ -0,0 +1,168 @@
...
+**No manual resource configuration is required.** AISimulate automatically
+detects the local machine's available RAM and CPU capacity, reserves host
+headroom, and limits parallel simulations to fit the estimated resource budget.
+You can omit the entire `execution.resources` section. Set explicit limits only
+when you want to override the automatic budget.
+
+Normal package installation includes the required `psutil` and `ijson`
+dependencies. Downstream images that install an AISimulate wheel with
+`--no-deps` must explicitly install both dependencies before adopting a release
+containing these controls.
```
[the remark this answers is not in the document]
jasonqinzhou: This is a downstream adoption requirement, already documented by the lines attached to this comment, rather than a break in the current pinned runtime. I verified that Dynamo still pins the published aisimulate==0.12.0 in container/deps/requirements.aisimulate.txt. This PR does not publish or replace that artifact, and the new resources.py module is not added to its existing wheel by merging source here. When Dynamo updates its pin to a release containing this change, its --no-deps runtime must include ijson (and psutil), as the new documentation explicitly requires. The exact package dependencies are declared and tested here. Your current-head review response already confirmed this release boundary: https://github.com/ai-dynamo/aisimulate/pull/169#issuecomment-5722932776 . Resolving the duplicate current-breakage finding on that basis; no downstream release adoption is being represented as qualified by this PR.

jasonqinzhou:
Final assessment of the automatic report on `fda69bd4`:

- The linked-repository dependency comment repeats the release-scoped finding already answered and accepted in [the current-head review](https://github.com/ai-dynamo/aisimulate/pull/169#issuecomment-5722932776). The new dependency requirement is documented; merging this PR does not alter Dynamo's pinned published wheel. The inline thread has been answered and resolved.
- The outside-diff suggestion to add shell guards to the recommendation examples is optional documentation hardening, not a new correctness defect in this change. The migration guide explicitly asks the reader to inspect selected results before evaluating the first result; both guides document empty/resource-limited results and instruct the reader to inspect them before proceeding. The commands themselves are unchanged from base. If a reader nonetheless invokes `predict` with a missing YAML, the existing loader rejects it with a clear error before simulation; this PR also preserves existing outputs on that path. The tested successful-path example is not presented as an unattended shell script or automatic fallback. No behavior-changing fix is needed for this suggestion.
- The modeling-evidence warning was generated without the just-added [bounded native base/head comparison](https://github.com/ai-dynamo/aisimulate/pull/169#issuecomment-5722982880). That comment includes the full reproduction script, machine-readable inputs/host snapshots/estimates/observed RSS/admission results, anomaly assertions, and an explained comparison. The distinction between this synthetic evidence and production/RSS calibration is explicit. Recommendation selection and partial-result evidence is separately linked in the independent human review and regression table. No unverified calibration or unchanged-prediction claim is being substituted for measurement.

Codex finds no remaining actionable issue in the submitted change. Full CI is running on the unchanged reviewed head: https://github.com/ai-dynamo/aisimulate/actions/runs/35289797545 .

PR #173, opened 2026-09-11
fix(replay): reporting and lifecycle cleanup

jasonqinzhou:
## Why and what changed

This PR is scoped to **replay reporting and request-lifecycle cleanup**. Large offline replays retain completed request state and allocate additional latency arrays at report time. The change releases state after its final consumers are done and stores exact summary latency samples on disk when their in-memory buffer fills.

The implementation covers three related parts:

1. **Reporting cleanup:** fold finished requests into aggregate statistics and spill exact latency samples to replay-owned anonymous temporary files after 4,096 samples per distribution. All samples contribute to the report; counts and rounded-rank percentiles remain exact. Mean/std can differ at floating-point roundoff.
2. **Lifecycle cleanup:** retire completed batch-summary records after completion callbacks, and release disaggregated request state after both pipelines and the handoff coordinator are quiescent. A cancellation-accounting guard ensures that the same request is retired exactly once.
3. **Supporting reporting correctness:** configure SLA thresholds before incremental aggregation, propagate actual temporary-storage failures through Rust/Python, and cover these paths with parity, lifecycle, and boundary tests.

Active requests and steppable SDK completed-request queries retain their existing behavior. Detailed reports keep their complete, ordered in-memory list API. Temporary files are released when reporting finishes or fails; existing user files are untouched.

Scope excludes workload caps, host-budget enforcement, sampling, truncation, a new streaming-output API, and changes to performance models or search policy. The 4,096-sample threshold only chooses storage location. Temporary-disk use grows with the workload, and actual storage failures stop the replay rather than produce a successful partial report.

## Review map

- Risk level: medium (request lifecycle and reporting).
- Scope audit: all 12 changed files at `cc7ddf09810e848d5b3c8989550074224c6aee3f` belong to the reporting/lifecycle implementation, its cross-language error propagation, tests, or documentation. No unrelated change was identified.

| Files | Purpose within this scope |
| --- | --- |
| `crates/core/src/replay/report.rs`, `report/bounded.rs` | Completed-record retirement, incremental statistics, exact disk-backed latency samples, and preservation regressions. |
| `crates/core/src/replay/agg.rs`, `disagg.rs`, `replayer.rs` | Apply cleanup only to batch execution, wait for lifecycle completion, and initialize SLA thresholds before incremental reporting. |
| `crates/core/src/engine/scheduler/rank.rs` | Guard against publishing duplicate retirement for a canceled request; this is a lifecycle-accounting fix. |
[...]

PR #170, opened 2026-09-10
fix(sweeper): bound worker cleanup before pool replacement

jasonqinzhou:
Direct `Sweeper` callers can start a replacement pool while timed-out workers are still alive, or hang while shutting down a completed sweep. This change uses the bounded worker cleanup introduced by #169 for those callers and fixes a shutdown-lock deadlock in the shared helper, so stuck finalizers cannot prevent escalation to kill.

#169 has merged. This PR now targets `main` and contains only four files:

- Use `terminate_pool` before replacement and after orchestration failures, and `close_pool` after successful direct sweeps. Failed cleanup prevents replacement.
- Avoid calling executor shutdown again while its manager may hold the shutdown lock joining a hung finalizer. The manager remains the sole worker-join owner; survivors are killed before final executor shutdown.
- Exercise real spawned workers that ignore SIGTERM, normal finalizers, and hung finalizers. Preserve completed candidate results and verify that workers have been reaped.
- Document the cleanup guarantee for direct Sweeper callers, including custom factories without resource admission. Sequential Runner execution remains outside this process-pool guarantee.

## Validation

Head: `ef372a936c2fcddefdc80be0babcc798d2ad4db5`, based on `main` at `2832e6b8756b8b83f18aee2f8e8fcaf23eea8675`.

All four changed files are byte-for-byte identical to the previously approved head `c56b38e5ca192aa0b4dafc4f0a77c0c8fe52ad89`. The rebase removes duplicated parent history after #169's squash merge.

- Configured Ruff lint/format, documentation links, strict CODEOWNERS coverage, and whitespace checks pass.
- Fresh Codex self-review of the complete diff and cleanup callers found no remaining actionable issue.
- **721 passed, 5 skipped on the current head**, covering the complete Sweeper suite plus host resources, resource scheduler, supervision, and unified CLI. The native extension was rebuilt from this commit. Tests use the repository's warnings-as-errors configuration on macOS/Python 3.12; five optional cross-repository Dynamo packaging contracts are skipped.
- [Fast CI passed on the current head](https://github.com/ai-dynamo/aisimulate/actions/runs/35291350013).
- Historical regression evidence: the worker-replacement test fails against #169's original Sweeper implementation because replacement overlaps live workers; the hung-finalizer test reproduced the inherited shutdown-lock deadlock before its correction.
- [CodeRabbit reviewed this exact head with no actionable findings](https://github.com/ai-dynamo/aisimulate/pull/170#issuecomment-5723188111).
[...]

PR #174, opened 2026-09-11
test(resources): cover safe admission of the reported Mac sweep

jasonqinzhou:
A user reported a reproducible Mac kernel panic after running `aisimulate recommend --stack dynamo` with concurrency up to 64,512, 10,240-token prompts, and 100 requests per load unit. The largest candidate represents **6,451,200 requests**; eager u32 prompt storage alone would require **264.24 GB (246.09 GiB)**. This is a calculated allocation, not measured crash RSS or a confirmed OS-panic diagnosis.

The safety implementation is already on `main` in #169, #170, and #173. This PR adds only a regression for the reported workload: reject the oversized candidate before replay while preserving fitting work, regardless of candidate order.

## Scope

**Two files, +94/-2 lines; no production changes.**

- Preserve the original 40-line YAML fixture, including all seven concurrency choices, 72 GPUs, scheduler/memory domains, token lengths, stop rule, optimizer settings, and routing. A normalized checksum prevents silently shrinking the reproducer.
- Add one parametrized test to the existing resource-scheduler suite: largest candidate alone, small then largest, and largest then small.

The test derives candidate bounds from the public configuration model and calls the production `GuardedRunnerFactory` / `evaluate_waves` path with a 2 GB budget and one allowed CPU. It verifies a `resource_limited` outcome for exactly 6,451,200 requests and preserves the concurrency-8 result. An allocation-free replay sentinel fails immediately if the oversized candidate reaches execution. The fitting candidate runs in a real spawned worker.

The test intentionally covers the conservative `dynamo-eager-u32-v1` estimate used without the optional newer Dynamo capability. It does not require Dynamo to be installed. It no longer asserts blanket refusal of the full search domain, which would contradict #169's partial-sweep behavior.

## Validation

Reviewed head: `00b8c986b9d10da6fee9c4ddefe128d8755a3e1c`, incorporating `main` at `4acab657e054b0ac135599b3a5206899cf831c3e`.

On macOS / Python 3.12.11, using the existing frozen development environment and this checkout's Python sources:

```sh
PYTHONPATH=python/aisimulate/src python -m pytest -c pytest.ini -p no:timeout \
  tests/test_resource_scheduler.py tests/test_host_resources.py -q
```

- **47 passed**, including all three new regression cases; repository warnings-as-errors settings retained.
- Negative control: bypassing admission makes **all three new cases fail** at the oversized-replay sentinel. No large allocation is attempted.
- Package-configured Ruff lint/format, strict CODEOWNERS coverage, and whitespace checks pass.
- Fresh self-review confirms the complete diff contains only the fixture and test addition.
- [Fast CI passed on this exact head](https://github.com/ai-dynamo/aisimulate/actions/runs/35296203066), including Repository Policy, Python Static Checks, Rust Format, and the aggregate gate. CODEOWNERS and DCO also passed. CodeRabbit skipped this draft; Full CI and human review remain outstanding.

## Safety protection versus full workload support

**#174 verifies safety protection; it does not demonstrate completion of the user's entire sweep.** The Dynamo change is an important enabling step, but merging it alone does not establish full support.

| Milestone | Current evidence |
| --- | --- |
| Stop oversized candidates before replay and preserve fitting work | Implemented on AISimulate main by [#169](https://github.com/ai-dynamo/aisimulate/pull/169), [#170](https://github.com/ai-dynamo/aisimulate/pull/170), and [#173](https://github.com/ai-dynamo/aisimulate/pull/173). [#174](https://github.com/ai-dynamo/aisimulate/pull/174) adds a small regression using production admission/scheduling and an allocation-free replay sentinel; it does not execute the full native Dynamo workload. |
| Reduce Dynamo's upfront prompt allocation | [Dynamo #14679](https://github.com/ai-dynamo/dynamo/pull/14679), head `81c3909f0704d41e12b3157da06032c9985fc136`, is still open as of September 17, 2026. It generates prompts as requests enter the active concurrency window. The matching native binary must be installed to use this path; older binaries retain the conservative eager estimate. |
| Complete the unchanged original sweep on the user's Mac | Not yet demonstrated. Current combined-build native qualification, full-workload completion, and reporter-confirmed incident closure remain outstanding. |

For the largest candidate (6,451,200 requests, concurrency 64,512, 10,240 input / 1,024 output tokens), #14679 reduces **calculated prompt-token storage from 246.09 GiB to 2.46 GiB**. This excludes other runtime memory. Its current conservative **total estimate remains approximately 128.47 GiB**, including reporting/evidence allowances. These are calculations and policy estimates, not measured full-scale RSS or a proven machine-size requirement. The candidate therefore still fails admission under the 2 GB budget used by #174.

To claim full workload support, validate the exact integrated AISimulate/Dynamo builds through the real public CLI with the unchanged workload on a specified host and declared resource budget; record completion, peak process-tree memory, runtime, and cleanup. Further reporting/estimator work or a larger host may be needed. A safe refusal or partial run establishes safety handling, while complete execution and the reporter's confirmation require their own evidence.

The earlier [Mac native manifest and build log](https://github.com/ai-dynamo/aisimulate/tree/dafeae296a0cff039e68da188dc5cf51de740a39/tests/e2e/configs/resource_safety/evidence) remain historical evidence at their immutable revision: an 8-request smoke replay completed, and both large CLI cases refused under 2 GiB. They do not qualify this head. The runner, README, manifest, and build log are removed from this PR's current diff; incident context and validation details live here.

Tracks [AIC-1956](https://linear.app/nvidia/issue/AIC-1956). **Keep in draft.**


scripts/verify_reported_resource_safety.py
```diff
@@ -0,0 +1,202 @@
...
+    # Require a real native module and complete a small native replay first.
+    # An import error or placeholder runtime must never count as safe handling.
+    from dynamo import _core
+    from dynamo.mocker import MockEngineArgs
+    from dynamo.replay import run_synthetic_trace_replay
+    from aisimulate.config.common import ResourceConfig
+    from aisimulate.supervision import run_process
+
+    native_hash = digest(Path(_core.__file__).read_bytes())
+    with zipfile.ZipFile(args.dynamo_wheel) as wheel:
```
jasonqinzhou, line 103: The evidence runner hashes the native library and wheel with `read_bytes()`/`wheel.read()`. Debug native libraries can be hundreds of MiB or larger, and this runs before the supervisor budget. Hash files and archive members in bounded chunks so the safety verifier cannot create its own large allocation.

scripts/verify_reported_resource_safety.py
```diff
@@ -0,0 +1,202 @@
...
+            block_size=4, num_gpu_blocks=128, speedup_ratio=1000.0
+        ),
+    )
+    summary = smoke.summary
+    if summary["num_requests"] != 8 or summary["completed_requests"] != 8:
+        raise RuntimeError(f"native smoke replay did not complete: {summary}")
+
+    original = yaml.safe_load(CASE.read_text())
+    identity = {
+        "fixture_sha256": digest(CASE.read_bytes()),
```
jasonqinzhou, line 163: The manifest records this checkout's AISimulate HEAD, but imports can resolve to a different installed/editable AISimulate tree. Verify the loaded Python source against the recorded checkout and record the actually loaded source/native identities. Otherwise the runner can associate a newer revision with older CLI/preflight code.

scripts/verify_reported_resource_safety.py
jasonqinzhou: Fixed in 8c226ae2c48826e1b47fb7845a398cc85d12b75d: native libraries, wheels and archive members are hashed with 64 KiB reads. The regression reader rejects unbounded or oversized reads.

scripts/verify_reported_resource_safety.py
jasonqinzhou: Fixed in 8c226ae2c48826e1b47fb7845a398cc85d12b75d: compare loaded AISimulate and Dynamo replay Python trees with the recorded checkouts, record both native binary identities, and test that differing source trees fail.

.github/codeowners/areas.yaml
```diff
@@ -55,6 +55,8 @@ areas:
       - tests/test_runner.py
       - tests/test_traffic.py
       - tests/e2e/
+      - tests/test_reported_resource_regression.py
```
jasonqinzhou, line 63: The new regression test is assigned only to the replay team. Repository policy requires maintainer coownership, and exact-head Full CI fails test_subsystem_teams_retain_maintainer_coownership on this path. Add the corresponding shared ownership rule and regenerate CODEOWNERS.

tests/test_reported_resource_regression.py
```diff
@@ -0,0 +1,181 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Deterministic preflight regression; native CLI evidence has a separate runner."""
+
+from copy import deepcopy
```
jasonqinzhou: Exact-head Python Static Checks fail I001 because CI uses the package Ruff configuration for root tests. Normalize this new file with that configuration and run the same CI lint command locally.

.github/codeowners/areas.yaml
jasonqinzhou: Fixed in 912c7745, now on the remote head. Added shared maintainer/replay/sweeper rules and regenerated CODEOWNERS. All 105 ownership policy tests pass.

tests/test_reported_resource_regression.py
jasonqinzhou: Fixed in 912c7745, now on the remote head. The full package-configured CI lint command and all 44 resource regression tests pass locally.

tests/test_reported_resource_regression.py
```diff
@@ -0,0 +1,165 @@
...
+        requested_parallelism=8,
+    )
+    assert plan["status"] == "resource_limited"
+    assert plan["estimate"]["input_token_bytes"] == 264_241_152_000
+    assert plan["effective_parallelism"] == 0
+    assert raw == original
+
+
+@pytest.mark.parametrize("largest_only", [False, True])
+def test_sweep_refuses_before_candidate_sampling_or_native_allocation(tmp_path, monkeypatch, largest_only):
```
jasonqinzhou: The full-sweep assertion now conflicts with the behavior merged in #169: oversized candidates must be refused individually while fitting candidates remain available. Replace blanket refusal-before-sampling with a focused regression for the reported largest candidate and retained fitting results, using current main interfaces.

jasonqinzhou (commented):
Refreshing this draft into a minimal regression-only PR on current main, as requested. The implementation has landed in #169/#170/#173; incident context and evidence will move into the description.

scripts/verify_reported_resource_safety.py
```diff
@@ -0,0 +1,248 @@
...
+    manifest = args.output_dir / "manifest.json"
+    manifest.write_text(json.dumps(identity, indent=2) + "\n")
+    runs = []
+    for name, largest_only in [("full-sweep", False), ("largest-candidate", True)]:
+        raw = yaml.safe_load(CASE.read_text())
+        if largest_only:
+            raw["traffic"]["load"]["concurrency"] = 64_512
+        # Only the execution budget changes. Model, topology, token lengths,
+        # stop rule, and scheduler/search domains remain the reported values.
+        raw["execution"] = {"resources": {"memory_limit_gib": 2.0, "cpu_limit": 1}}
```
jasonqinzhou: The evidence runner still uses memory_limit_gib, which current main replaced with decimal memory_limit_gb. Together with the obsolete GIB/cli._main test references, this prevents reusing the qualification additions on the integrated code. Remove the standalone runner and committed incident artifacts from the code diff; retain current reproducible validation and its limitations in the PR description.

tests/test_reported_resource_regression.py
jasonqinzhou: Fixed in 00b8c986. The original fixture now drives the existing resource-scheduler test with largest-only, small-then-largest, and largest-then-small cases. Oversized work must be resource_limited before the replay sentinel; fitting results survive in either order. All 47 resource tests pass, and bypassing admission fails all three new cases.

scripts/verify_reported_resource_safety.py
jasonqinzhou: Fixed in 00b8c986. Removed the standalone runner, separate regression module, README, manifest, build log, and associated ownership additions from the current diff. The two-file regression uses current-main APIs and memory_limit_gb; incident details, historical evidence links, validation, and remaining native-qualification limits are in the rewritten PR description.

PR #182, opened 2026-09-11
fix(replay): enforce total event ordering and reject non-finite timing

ajcasagrande:
## Why and what changed

`SimulationEvent` used `partial_cmp(...).unwrap_or(Equal)` for timestamps, which violates the total ordering required by `BinaryHeap` for NaN and disagrees with the existing bitwise equality for signed zero. Use `f64::total_cmp`, retaining kind-rank and sequence tie-breaks.

Also reject non-finite prefill/decode timing results before the historical timing floors. Previously, `f64::max` silently converted NaN or negative infinity to 0 ms (prefill) or 1 ms (decode), so an invalid external timing provider could produce a successful but misleading replay report. Invalid timing now propagates as `ReplayError::Engine` through aggregate and disaggregate replay.

The original review's NaN-completion hang was not reproduced through the public native runtime: the generalized engine already validates rank pass-end timestamps before scheduling completions. The runtime regression instead exposed the normalization defect above, which this change fixes at its source.

## Review map

- Risk level: medium.
- Start with `crates/core/src/replay/event.rs` for the heap ordering, `crates/core/src/engine/common/perf_model.rs` for validation before timing normalization, and `crates/core/tests/replay_non_finite_timing.rs` for public-API coverage.
- Public or serialized contract changed: no declarations or schemas changed; non-finite timing-provider results now fail explicitly.
- Compatibility or rollback concern: finite timing floors, event tie-breakers, and valid replay behavior are preserved. Rolling back restores inconsistent heap ordering and silent non-finite timing normalization.

## Evidence

- Tests: `cargo test --locked -p aisimulate-core --test replay_non_finite_timing --lib -j 4 -- --test-threads=4` — 1,215 library tests passed, 1 ignored; public runtime regression passed all 96 cases.
- Current-main compatibility: clean merge with `68d3e1b7`; all 251 replay tests and the 96-case public runtime regression passed on the combined tree.
- Formatting: `cargo fmt --all -- --check` and `git diff --check` — passed.
- Fast CI: [passed on 1d393c9a](https://github.com/ai-dynamo/aisimulate/actions/runs/34652058922).
- Full CI: [passed on 1d393c9a](https://github.com/ai-dynamo/aisimulate/actions/runs/34652440737).
- CodeRabbit reviewed commit: `1d393c9aca2d3b02df4eb4eb14e7ce9c501f2c84` — completed with no actionable comments.
- Codex reviewed commit: 1d393c9aca2d3b02df4eb4eb14e7ce9c501f2c84; full diff and the engine-to-runtime error propagation reviewed.
- Negative or boundary cases: heap NaN/signed-zero regressions; 72 invalid-timing cases covering NaN and both infinities, all three backends, aggregate/disaggregate, prefill/decode, and DP sizes 1/2; 24 finite controls complete normally.
- Expected-value derivation: heap tests assert ascending finite timestamps and bitwise equality consistency. Runtime tests require the actual invalid provider to be called and its failure to emerge as an engine error; each finite control must complete exactly one request.
- Before/after output: an aggregate decode provider returning NaN previously completed the request with finite reported timing; after the fix it returns an engine error identifying a non-finite decode duration. No prediction goldens were refreshed.

Synthetic timing-provider fixtures exercise input validation and runtime error propagation. No measured data, performance tables, formulas, or finite-input prediction changes.

## Tracking

- Closes: N/A.
- Related PRs or issues: [review discussion](https://github.com/ai-dynamo/aisimulate/pull/182#discussion_r3993679817).


PeaBrane (approved):
✅

dreamtalen (approved):

crates/core/src/replay/event.rs
```diff
@@ -124,11 +124,70 @@ impl<Events: EngineEventBatch> PartialOrd for SimulationEvent<Events> {
...
+        // `total_cmp`, not `partial_cmp`: this feeds a `BinaryHeap`, which requires a
+        // total order agreeing with `Eq`. `partial_cmp(..).unwrap_or(Equal)` makes a NaN
+        // timestamp compare equal to everything, silently corrupting heap ordering
+        // instead of surfacing the bad input, and disagrees with the bitwise `PartialEq`
+        // above on signed zero.
         other
             .at_ms
-            .partial_cmp(&self.at_ms)
-            .unwrap_or(Ordering::Equal)
+            .total_cmp(&self.at_ms)
```
jasonqinzhou, line 134: Changing this to `total_cmp` restores a total heap order, but it still leaves NaN as an accepted event timestamp. In the real consumer, `next_timestamps()` advances `now_ms` to the heap head and every `pop_ready_*` helper requires `event.at_ms == now_ms`; for NaN that comparison is always false. Once a NaN worker completion reaches the head, no handler removes it, the request remains in flight, and the outer replay loop repeats forever. The new heap-only test stops before exercising that path, so it passes while the liveness failure remains. Please reject non-finite event timestamps at the scheduling boundary with an error propagated by both aggregate and disaggregate runtimes, and add a runtime-level regression showing that a NaN completion fails instead of hanging.

jasonqinzhou (changes requested):

## What this PR does

This PR replaces the replay event heap's partial floating-point timestamp comparison with f64::total_cmp and adds focused tests for NaN, signed zero, and heap ordering.

The change correctly recognizes that BinaryHeap needs a total Ord implementation, preserves the existing kind and sequence tie-breakers, and adds a compact regression that demonstrates the prior non-transitive heap behavior.

## Why this score

The comparator now satisfies a total ordering for floating-point bit patterns, but the proposed NaN behavior is not safe in the real replay consumer. A NaN event still becomes the heap head and can never pass the event.at_ms == now_ms guards, so aggregate or disaggregate replay can loop indefinitely; this P1 liveness issue requires explicit non-finite timestamp rejection and a runtime-level regression before approval.

crates/core/src/replay/event.rs
jasonqinzhou: I rechecked this finding on 814fc727618f854a40605299fc4baeb705df83b9 through the public run_engine_replay_with_timing API. The claimed NaN-completion hang is not reachable through the current native engine: Engine::execute_pass already validates every rank pass end before returning a scheduled completion. However, the runtime test exposes a related concrete input-validation defect: PerfModel::predict_prefill_time and predict_decode_time apply max(0.0)/max(1.0) before modeled_duration_ms validates the value. A NaN external timing result is therefore silently converted into a plausible finite duration; an aggregate decode test completed successfully despite exercising the NaN provider.

I am correcting that boundary by rejecting non-finite provider results before the historical timing floors, and adding public replay regressions covering both topologies, all three backends, prefill/decode, and attention-DP. This preserves finite timing behavior and the existing engine pass-end validation while making invalid timing fail explicitly. I will resolve this thread after the fix is pushed and validated.

jasonqinzhou: Addressed in 1d393c9aca2d3b02df4eb4eb14e7ce9c501f2c84. Non-finite prefill/decode predictions now fail before max(0.0)/max(1.0) can conceal them. The existing generalized-engine rank pass-end validation remains the scheduling guard.

The new public replay regression exercises 72 invalid cases (NaN, positive/negative infinity; vLLM/SGLang/TRT-LLM; aggregate/disaggregate; prefill/decode; DP 1/2), verifies that each invalid provider is actually called, and requires ReplayError::Engine. All 24 finite controls complete one request. Local validation passed: 1,215 library tests, 1 ignored; the public runtime regression; Rust formatting and diff checks. Full CI will be rerun on this exact commit before approval/merge.

Resolving with this correction to my original finding: the demonstrated failure was silent timing normalization, not a reachable NaN-completion hang in the native runtime.

jasonqinzhou (approved):
Reviewed 1d393c9aca2d3b02df4eb4eb14e7ce9c501f2c84. The review thread is resolved: non-finite timing is rejected before normalization, and the public runtime regression verifies error propagation across both topologies, all three backends, prefill/decode, and DP sizes 1/2. The total_cmp change preserves finite event ordering and existing tie-breakers. No remaining actionable findings.

Validation passed: 1,215 library tests (1 ignored), the 96-case public runtime regression, 251 replay tests plus the runtime regression on the clean current-main merge, Fast CI, CodeRabbit, and Full CI on this exact commit: https://github.com/ai-dynamo/aisimulate/actions/runs/34652440737.

PR #122, opened 2026-09-07
feat(replay): report deterministic AgentX play outcomes

Harrilee:
Implements [AIC-1816](https://linear.app/nvidia/issue/AIC-1816). Base: main; risk: medium.

AgentX plays report explicit `completed`, `failed`, or `incomplete` outcomes. The primary failure is selected by earliest causal-terminal time, then graph ordinal, so callback order and status severity do not choose the result. Outcomes include the failure request/status and actual resource-settlement time, and remain in authored play order when later plays settle first.

Client scheduling follows merged #121: a lane advances after all required client requests terminate, while the play outcome remains `incomplete` until server cleanup finishes. Failed plays skip undispatched work and wait for dispatched siblings; `causal_terminal_ms` identifies the primary failure and may precede lane release. This retains the strict replay failure policy and does not claim full AIPerf error-policy parity.

The Weka timing tests cover completion gaps, overlapping and post-completion spawns, blocking joins, parent resumption, background requests, missing `api_time`, and zero-output requests. Outcomes propagate through both replay runtimes into report JSON. Public aggregated qualification remains in #190.

## Validation

- Integrated main `24ec3028092245c370ac830ad18043e2bbb1ed67` in `32df941955588c0328824ebe1bbddd2e0e8b2c2f`, preserving #121's atomic feedback validation and mixed-batch regressions.
- The resulting diff against main has the same stable patch ID as the original #122 feature: 10 files, 467 additions, 11 deletions.
- Fresh full Rust package tests: **1,376 passed, 1 existing ignore** (`cargo test -p aisimulate-core --locked`).
- The original mixed-feedback reproducer now confirms unchanged lifecycle/dependencies after rejection and monotonic timestamps after recovery.
- Rust formatting and whitespace checks pass. [Full CI](https://github.com/ai-dynamo/aisimulate/actions/runs/35008645029) passed on this exact head: 53 successful jobs and the expected release-staging skip. Standalone Fast CI, DCO, and CODEOWNERS checks also passed.

Review order: `loadgen/driver.rs` and `loadgen/types.rs`, runtime/report propagation, then driver/P/D/Weka regression tests.

PR #195, opened 2026-09-14
fix(ci): compare rebased copies against their PR base

Harrilee:
## Why and what changed

Trusted `pull-request/*` copies trigger Full CI through push events. After a rebase, `push.before` can point to an unreachable previous copy head. The whitespace gate then fails with `Invalid symmetric difference expression`, preventing the runtime suites from starting.

Resolve the originating PR's head and base together, retain exact-head verification, and pass the validated PR base to Fast CI. Fetch missing base history using a temporary, read-only GitHub-token header before checking the complete PR range. The token is not persisted. Other triggers keep their existing whitespace ranges: ordinary main/release pushes use `event.before`; manual runs outside trusted-copy refs and new release branches retain the last-commit fallback. Collector and prediction comparison bases are unchanged.

## Review map

- Risk level: **medium** — changes CI execution and comparison-base contracts.
- Start with: `.github/workflows/ci.yml` (`verify-target` → Fast CI input), then `.github/workflows/fast-ci.yml` (`BASE_SHA` → authenticated fetch → whitespace check). Tests are in `tests/test_ci_workflow_contracts.py` and the existing `.github/codeowners/test_aisimulate_policy.py` mock.
- Public or serialized contract changed: no product API/schema change; Fast CI adds the optional `base_sha` reusable-workflow input.
- Compatibility or rollback concern: retain exact-head checks, read-only permissions, `persist-credentials: false`, and full-range whitespace validation. Reverting this change restores the rebased-copy failure. Changes are limited to two workflows and two CI test files.

## Evidence

- Tested commit: `c08ebd5b8a338f8d6ed85b4117b213c092d3a642`.
- Tests: `python -m pytest -c /dev/null .github/codeowners/test_*.py tests/test_ci_workflow_contracts.py tests/test_ci_qualification.py -q -p no:cacheprovider --override-ini='addopts=' --override-ini='filterwarnings='` — **209 passed**.
- Static/policy checks: `ruff check --config python/aisimulate/pyproject.toml tests/test_ci_workflow_contracts.py`, `ruff format --check --config python/aisimulate/pyproject.toml tests/test_ci_workflow_contracts.py`, `git diff --check`, `python3 scripts/check_copyright.py`, and `python3 scripts/check_packaged_legal_files.py` passed.
- Fast CI: [passed on `c08ebd5b`](https://github.com/ai-dynamo/aisimulate/actions/runs/34796495550); [CODEOWNERS also passed](https://github.com/ai-dynamo/aisimulate/actions/runs/34796495552).
- Full CI: [queued on `c08ebd5b`](https://github.com/ai-dynamo/aisimulate/actions/runs/34796499581). Previous revision `68433bad` [passed on its trusted copy](https://github.com/ai-dynamo/aisimulate/actions/runs/34792677350); that result does not certify this follow-up commit.
- CodeRabbit reviewed commit: `68433bad14cde57a8cf1ab0ccae0aab29b6be1fe`; new-head review pending.
- Codex reviewed commit: `c08ebd5b8a338f8d6ed85b4117b213c092d3a642` — author-side call-chain and regression review; reviewer recheck of the follow-up remains pending.
- Negative or boundary cases: changed PR head; malformed/missing base; API failure; unreachable old copy head; stacked parent changes excluded; whitespace in an earlier PR commit detected; missing history fetched without truncation/tags or persisted credentials; fetch failure terminates before diff; empty/all-zero bases preserve the last-commit fallback. Extracted whitespace and shared workflow scripts use GitHub's Bash error-handling flags.
- Expected-value derivation: temporary Git histories define the parent base, all commits belonging to the PR, and the last-commit fallback directly; no numerical model assertions.
- Before/after evidence: [original copied-head failure](https://github.com/ai-dynamo/aisimulate/actions/runs/34788614489); [#122 using its stacked PR base](https://github.com/ai-dynamo/aisimulate/actions/runs/34792674185); [#190 using its stacked PR base](https://github.com/ai-dynamo/aisimulate/actions/runs/34792673388). The missing-base transport test uses a local file origin plus an authentication contract check; it is not a hosted missing-base reproduction.

N/A — no simulation model, performance data, or product-output changes.

## Tracking

- Closes: N/A — CI prerequisite extracted from the runtime stack.
- Related PRs: [#121](https://github.com/ai-dynamo/aisimulate/pull/121), [#122](https://github.com/ai-dynamo/aisimulate/pull/122), [#190](https://github.com/ai-dynamo/aisimulate/pull/190).
- Merge this prerequisite first, then ensure #121 targets `main`. #121 currently targets this branch; #122 and #190 remain stacked above it.


.github/workflows/fast-ci.yml
```diff
@@ -127,6 +132,9 @@ jobs:
         shell: bash
         run: |
           if [[ -n "${BASE_SHA}" && ! "${BASE_SHA}" =~ ^0+$ ]]; then
+            if ! git cat-file -e "${BASE_SHA}^{commit}" 2>/dev/null; then
+              git fetch --no-tags origin "${BASE_SHA}"
+            fi
```
Arsene12358, line 143: **F1 (P2)** — this fetch has no credentials on the hosted runner: the repository is internal and the checkout above ran with `persist-credentials: false` (the job log shows `Removing auth` unsetting the extraheader before this step), so it fails whenever it is actually reached. `collector-check.yml` solves the same problem by injecting `github.token`; mirroring it keeps the stale-base case working. Alternatively drop the fetch, the `fetch-base-history` test case, and the sentence in the description.

```suggestion
        shell: bash
        env:
          GH_TOKEN: ${{ github.token }}
        run: |
          if [[ -n "${BASE_SHA}" && ! "${BASE_SHA}" =~ ^0+$ ]]; then
            if ! git cat-file -e "${BASE_SHA}^{commit}" 2>/dev/null; then
              # persist-credentials: false dropped checkout's auth header and the
              # repository is internal; authenticate like collector-check.yml.
              basic_token="$(printf 'x-access-token:%s' "${GH_TOKEN}" | base64 | tr -d '\n')"
              git -c "http.https://github.com/.extraheader=AUTHORIZATION: basic ${basic_token}" \
                fetch --no-tags origin "${BASE_SHA}"
            fi
```

tests/test_ci_workflow_contracts.py
```diff
@@ -463,6 +463,181 @@ def _run_comparison_base(
...
+
+    assert (result.returncode != 0) is earlier_whitespace, result.stdout + result.stderr
+    assert "parent-pr.txt" not in result.stdout
+    if earlier_whitespace:
+        assert "first-pr.txt" in result.stdout
+        assert "trailing whitespace" in result.stdout
+
+
+@pytest.mark.parametrize("available", [True, False], ids=["fetch-base-history", "missing-base-fails-closed"])
+def test_fast_ci_whitespace_fetches_an_absent_base_without_truncating_history(tmp_path: Path, available: bool) -> None:
```
Arsene12358, line 639: **F1 (test)** — this proves the fetch against a `file://` origin, which never asks for credentials, so it stays green while the hosted fetch fails (see the fast-ci.yml comment). If you keep the fetch, a contract pin next to `test_migrated_workflows_keep_reviewed_safety_fixes` would make the credential non-optional:

```python
def test_fast_ci_whitespace_fetch_authenticates_like_collector_check() -> None:
    job = _workflow("fast-ci.yml")["jobs"]["python-static"]
    step = next(step for step in job["steps"] if step.get("name") == "Check changed-line whitespace")
    assert step["env"]["GH_TOKEN"] == "${{ github.token }}"
    assert 'http.https://github.com/.extraheader=AUTHORIZATION: basic' in step["run"]
    assert '--no-tags origin "${BASE_SHA}"' in step["run"]
```

`_run_whitespace_step` can pass a dummy `GH_TOKEN`; the extra header is ignored by the `file://` transport, so this test keeps working.

.github/workflows/ci.yml
```diff
@@ -31,6 +31,7 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       comparison-base: ${{ steps.comparison-base.outputs.sha }}
+      whitespace-base: ${{ steps.pr-target.outputs.base_sha || steps.comparison-base.outputs.sha }}
```
Arsene12358: **S1 (non-blocking)** — for anything other than a `main`/`release/*` push, `comparison-base` is `main`'s tip (L95-98), so `workflow_dispatch` runs and the first push of a new `release/*` branch now whitespace-check `main...target` instead of `TARGET^..TARGET`; a dispatched stacked head will include its parent PR's commits. Probably fine, but undocumented. If the wider range isn't intended, `${{ steps.pr-target.outputs.base_sha }}` alone keeps the previous behavior for every other trigger (Fast CI still falls back to `pull_request.base.sha || event.before`).

tests/test_ci_workflow_contracts.py
```diff
@@ -463,6 +463,181 @@ def _run_comparison_base(
...
+    _git(repository, "add", name)
+    _git(repository, "commit", "--quiet", "-m", name)
+    return _git(repository, "rev-parse", "HEAD").stdout.strip()
+
+
+def _run_whitespace_step(repository: Path, base: str, target: str) -> subprocess.CompletedProcess[str]:
+    job = _workflow("fast-ci.yml")["jobs"]["python-static"]
+    script = next(step["run"] for step in job["steps"] if step.get("name") == "Check changed-line whitespace")
+    return subprocess.run(
+        ["bash", "-c", script],
```
Arsene12358: **S2 (nit)** — GitHub runs `shell: bash` steps as `bash --noprofile --norc -eo pipefail`; running the extracted script with plain `bash -c` means `missing-base-fails-closed` passes here because `git diff` fails after the ignored fetch error, not because the fetch aborts the step. Same verdict today; `["bash", "-eo", "pipefail", "-c", script]` would make the harness fail for the CI reason.

Arsene12358 (approved):
**Verdict: approve — no hard blockers.** One P2 (F1) I'd like fixed or explicitly deferred before merge; S1–S4 are non-blocking notes.

### Verification

I traced the change from the copy-branch push through `verify-target` → `whitespace-base` → the reusable Fast CI input → `BASE_SHA` → `git diff --check`, and re-ran the suites on `68433bad`.

- Hosted evidence for the bug: run [34788614489](https://github.com/ai-dynamo/aisimulate/actions/runs/34788614489) (`pull-request/121` @ `69901699`) ran the whitespace step with `BASE_SHA=e1dff098…` (the previous copy head) and failed with `fatal: Invalid symmetric difference expression …` (exit 128), which took `Fast CI Success` and `Full CI Success` down with it.
- Hosted evidence for the fix: the stacked copies used their own PR base, a value neither the old expression nor the `comparison-base` fallback could produce — [34792674185](https://github.com/ai-dynamo/aisimulate/actions/runs/34792674185) (`pull-request/122`) ran with `BASE_SHA=7252deba…` (#121's head) and [34792673388](https://github.com/ai-dynamo/aisimulate/actions/runs/34792673388) (`pull-request/190`) with `BASE_SHA=56ac9f86…` (#122's head). The `pull-request/121` run used `main`'s tip because #121 still targeted `main` when it started (it was retargeted 45 s later). This PR's own copy, [34792677350](https://github.com/ai-dynamo/aisimulate/actions/runs/34792677350), finished green.
- Local: the policy + contract + qualification suites collect 206 cases and all 11 new cases pass (also under bash 3.2). Ruff check/format, `git diff --check origin/main...HEAD`, copyright and packaged-legal checks are clean; actionlint reports only the pre-existing SC2016 info in the readiness step (same on `main`).
- No regressions on the other entry points: `main`/`release/*` pushes still compare against `event.before` (via `comparison-base`), standalone `pull_request` runs still use `pull_request.base.sha` (`inputs.*` is empty outside `workflow_call`), and the only new value that reaches a shell is `pr_base`, validated as 40-hex.

### F1 (P2): the absent-base fetch cannot authenticate on the hosted runner

The description says "fetch absent base history before checking the complete PR range", and `test_fast_ci_whitespace_fetches_an_absent_base_without_truncating_history[fetch-base-history]` covers it. On GitHub that path fails every time it is reached: the repository is internal, the `python-static` checkout runs with `persist-credentials: false` (its log shows `Removing auth` unsetting `http.https://github.com/.extraheader` at the end of the checkout step), and `git fetch --no-tags origin "${BASE_SHA}"` carries no credentials. `collector-check.yml` runs the identical fetch and injects `github.token` as a basic-auth extraheader for exactly this reason (and that fetch succeeds in the same Full CI run). The new test passes only because its origin is a `file://` path, which never asks for credentials.

This is not a regression — an absent base already failed with exit 128 — and in every hosted run so far the base was a branch tip already present in the `fetch-depth: 0` clone, so the fetch never executed. But it is a shipped path that a test certifies and the description advertises, and the case where it matters (an orphaned `.base.sha`) is exactly when a clear outcome is needed. Two acceptable fixes: mirror `collector-check.yml`'s token injection (inline suggestion on fast-ci.yml, plus a contract test pinning the header so it cannot silently disappear), or drop the fetch, the `fetch-base-history` parameterization, and the sentence in the description — for trusted copies the base is always reachable from a branch head, so the check stays correct without it.

### Non-blocking notes

- **S1** — `whitespace-base` falls back to `comparison-base`, which is `main`'s tip for non-lifecycle runs (ci.yml L95-98). `workflow_dispatch` runs and the first push of a new `release/*` branch therefore now check `main...target` instead of `TARGET^..TARGET`; for a dispatched stacked head that includes the parent PR's commits. Probably fine (tree-level diff, and every PR already passes this gate on its `pull_request` event), but it is undocumented — either note it or narrow the output to `steps.pr-target.outputs.base_sha` so the other triggers keep their previous behavior.
- **S2** — the new helpers execute step scripts with `bash -c`; GitHub runs `shell: bash` steps with `-eo pipefail`. Verdicts coincide today (a failed fetch is followed by a failing `git diff`), so this is a fidelity nit.
- **S3** — the body doesn't follow the PR template (Review map / risk level / Evidence fields / Tracking). REVIEW.md ties the Codex-review requirement to the declared tier and CodeRabbit rated this Medium; maintainers' call. The hosted runs above are the end-to-end evidence CodeRabbit said was missing and would be worth citing.
- **S4** — after this merges, #121 needs to go back to `main` as planned (GitHub does it automatically if this branch is deleted on merge).

.github/workflows/fast-ci.yml
Harrilee: Agreed; fixed in c08ebd5b8a338f8d6ed85b4117b213c092d3a642. The missing-base fetch now uses the same command-scoped GitHub-token Basic-auth header as collector-check.yml. `contents: read` and `persist-credentials: false` remain in place, and the fetch still preserves full history with `--no-tags`. Added an authentication contract check and an assertion that the fetch leaves no extraheader in local Git config. The 209 CI/policy tests pass locally; hosted checks are rerunning on this commit.

tests/test_ci_workflow_contracts.py
Harrilee: Fixed in c08ebd5b. Added `test_fast_ci_missing_base_fetch_uses_temporary_checkout_authentication`, covering the read-only token binding, disabled credential persistence, and the command-scoped authenticated fetch. The file-origin tests now use a dummy token and additionally verify that no auth header persists in local Git config. As you pointed out, these local transport tests do not establish hosted authentication; the separate contract check covers that wiring. All 209 CI/policy cases pass.

.github/workflows/ci.yml
Harrilee: Agreed; narrowed `whitespace-base` to `steps.pr-target.outputs.base_sha` in c08ebd5b. Fast CI retains the existing event-base fallback for other triggers, so manual runs outside trusted-copy refs and newly created release branches continue checking the last commit. Added empty/all-zero-base regressions, and retained the independent collector/prediction comparison-base behavior.

tests/test_ci_workflow_contracts.py
Harrilee: Fixed in c08ebd5b. `_run_whitespace_step` and the shared `_run_workflow_script` now run with `bash --noprofile --norc -eo pipefail`. The missing-base failure case additionally checks the upload-pack error and confirms that the later symmetric-difference error is absent, so the test now verifies termination at the failed fetch. All 209 CI/policy tests pass with these flags.

Harrilee:
@Arsene12358 Addressed F1 and S1–S2 in [c08ebd5b](https://github.com/ai-dynamo/aisimulate/commit/c08ebd5b8a338f8d6ed85b4117b213c092d3a642), with details in each inline thread. Local CI/policy suites: **209 passed**. [Fast CI](https://github.com/ai-dynamo/aisimulate/actions/runs/34796495550) and [CODEOWNERS](https://github.com/ai-dynamo/aisimulate/actions/runs/34796495552) passed on this exact commit; [Full CI](https://github.com/ai-dynamo/aisimulate/actions/runs/34796499581) is queued.

For S3, the description now follows the repository template, declares medium risk, and distinguishes current-head results from the earlier hosted runs. It also states the local missing-base test's transport limitation. For S4, the tracking section explicitly retains the post-merge step to ensure #121 targets main. This update only changes #195; the runtime stack has not been rebased or retargeted here.

PR #219, opened 2026-09-15
fix(dsv4): correct GEMM graph timing and MoE workspace accounting

YijiaZhao:
## Summary

Consolidate the DeepSeek-V4 alignment fixes and version-matched collection work in one PR, superseding #220. No performance-table column/unit or producer/consumer data-contract change is introduced.

1. **GEMM timing:** require CUDA Graph timing for every vLLM GEMM dtype, including `fp8_block`; fail rather than publish eager or unproven graph timing.
2. **MoE workspace:** use the DeepSeek-V4 residual hidden width (Pro: 7168), not expanded attention width (65536), while preserving the separate attention activation term.
3. **Collection and publication:** qualify the vLLM 0.25.0 collector paths, include the exact Pro shared-expert TP GEMM shapes, preserve checkpoints/failures, and publish **17 production Parquet tables / 419,538 rows** with original per-shard provenance. This remains a **partial all-model publication**.

#### A. Retained targeted-table accuracy comparison — not a new full-table replay

- **Scope:** DeepSeek-V4-Pro, B200, TP8 / attention-DP1 / EP1, config909; ISL/OSL = 1024/1024 and 8192/1024, CC = 1, 2, 4, 8, 16, 32, 64 (14 matched operating points).
- **Reference:** historical SemiAnalysis InferenceX measurements, as used by the retained dashboard snapshot generated on 2026-08-28. All latency comparisons below use **means, not medians**; MAPE is the unweighted mean of the pointwise absolute percentage errors (7 points per workload, 14 overall).
- **Old AISim:** the retained dashboard prediction using 0.24.0 performance tables. **Targeted AISim:** retained synthetic request-replay job `1961399`, using targeted vLLM **0.25.0** measurements and the workspace fix `2fce86f`; 38 eager communication controls were excluded, and zero 0.24.0 rows were admitted.
- **Important limitation:** this replay predates the newly uploaded all-model tables and the recovered historical server logs. KV-capacity / graph-reservation settings were partly inferred. It is **not** a new serving benchmark, a captured-server-timeline replay, or a latest-corpus E2E qualification.

| Workload | Old AISim TTFT MAPE | Targeted AISim TTFT MAPE | Old AISim TPOT MAPE | Targeted AISim TPOT MAPE |
|---|---:|---:|---:|---:|
| 1k/1k | 26.67% | 4.91% | 122.64% | 9.92% |
| 8k/1k | 8.93% | 9.59% | 113.94% | 9.69% |
| All 14 points | 17.80% | 7.25% | 118.29% | 9.81% |

**Residual discrepancy:** 8k/1k TTFT does not improve overall (8.93% → 9.59% MAPE). At CC1 and CC2, the targeted prediction is still 36.9% and 21.8% faster than the historical reference. Those reference points have not been remeasured; this PR does not claim that all E2E discrepancies are solved.

<details>
<summary>All 14 matched points — mean latency in milliseconds</summary>

Positive signed error means AISim predicts a slower latency; negative means faster.

| ISL/OSL | CC | Measured TTFT | Targeted AISim TTFT | TTFT error | Measured TPOT | Targeted AISim TPOT | TPOT error |
|---|---:|---:|---:|---:|---:|---:|---:|
| 1024/1024 | 1 | 76.17 | 78.77 | +3.4% | 10.42 | 9.51 | -8.8% |
| 1024/1024 | 2 | 92.31 | 81.28 | -12.0% | 10.67 | 9.80 | -8.1% |
| 1024/1024 | 4 | 92.05 | 88.48 | -3.9% | 11.73 | 11.38 | -2.9% |
| 1024/1024 | 8 | 101.70 | 101.63 | -0.1% | 12.92 | 12.98 | +0.5% |
| 1024/1024 | 16 | 123.73 | 127.20 | +2.8% | 14.69 | 17.77 | +21.0% |
| 1024/1024 | 32 | 167.60 | 174.02 | +3.8% | 19.55 | 22.20 | +13.6% |
| 1024/1024 | 64 | 246.33 | 267.15 | +8.5% | 24.13 | 27.66 | +14.6% |
| 8192/1024 | 1 | 517.80 | 326.69 | -36.9% | 10.39 | 9.68 | -6.9% |
| 8192/1024 | 2 | 428.84 | 335.22 | -21.8% | 11.11 | 10.32 | -7.1% |
| 8192/1024 | 4 | 374.32 | 368.09 | -1.7% | 12.47 | 12.32 | -1.2% |
| 8192/1024 | 8 | 465.97 | 445.66 | -4.4% | 14.52 | 14.83 | +2.2% |
| 8192/1024 | 16 | 583.95 | 573.83 | -1.7% | 17.92 | 21.81 | +21.7% |
| 8192/1024 | 32 | 826.60 | 830.18 | +0.4% | 26.17 | 30.34 | +16.0% |
| 8192/1024 | 64 | 1348.40 | 1350.94 | +0.2% | 38.89 | 43.91 | +12.9% |

Evidence: retained `job-1961399/results.json`, SHA-256 `fd372c134681e411f0a45bf7bc62e1275dab3ede3f0d9c9888399dbf4c97ce52`. The displayed means and MAPE were rechecked against the retained result and comparison artifacts; no new replay is implied.

</details>

#### B. Latest collector-code regression and published DSV4 data

Collector-code revision **`5515203`**, exact vLLM **0.25.0** runtime (`dd10e03f95f94edbea1975c67ace3a35ec9a8a40`), B200:

- **DSV4-specific smoke:** six attention/helper module groups plus mHC, **35/35 cases passed**; exact shared-down GEMM smoke **4/4 passed**. General GEMM smoke also passed **32/32** across bf16, FP8, FP8-block and NVFP4.
- **Full DSV4-family module/mHC coverage already published:** **22,744 successful cases, zero recorded failures** across the seven tables below. This is DSV4-family coverage used by Pro, not a claim that every case is Pro-only.

| Published DSV4-related table | Successful cases / rows | Failed cases |
|---|---:|---:|
| CSA context | 5,840 | 0 |
| HCA context | 5,840 | 0 |
| CSA generation | 1,544 | 0 |
| HCA generation | 1,544 | 0 |
| Paged MQA logits | 3,918 | 0 |
| HCA sparse attention | 3,918 | 0 |
| mHC | 140 | 0 |

- **Across all op families:** GPU smoke validated **25/26 batches / 148 cases**; **87 focused unit tests passed**. The remaining `compute_scale` batch measured 5/5 callbacks but failed finalization because the existing `scale_matrix_perf.txt` side output is not registered. Its output is withheld, not silently skipped or marked green.
- **Post-main-sync verification (`77122a7`):** the combined local collector/workspace suite passed **105 tests, with 1 native-runtime test skipped** (87 collector + 18 workspace tests). GitHub **Repository Policy, Python Static Checks, Rust Format, Fast CI Success, codeowners and DCO all passed** for this commit. This is Fast CI evidence, not a new E2E accuracy run.
- Clock policy: Slurm **1965 MHz requested and actual clocks sampled**; **hard clock locking was not verified**.
- The packaged `b200_sxm/vllm-0.25.0-collection-report.json`, `vllm-0.25.0-collection.md`, failure ledger, and per-table `collection_meta.yaml` contain coverage, hashes, jobs, limitations and original source revisions. Smoke rows are not merged into production tables.

### Current scope: optional campaign tooling removed (`bdc9eec9`)

- Per requester direction, remove `collector/campaigns/` and the three test modules dedicated to its runner/hook. The matching tools, configuration, instructions and tests are retained in a **verified local-only archive**, not moved to another GitHub directory, branch or package. This does not remove operator collection fixes or change the stock `collect.py` executor.
- Preserve the generic hash-checked external runtime-manifest interface. Its retained tests use an explicitly synthetic fixture rather than depend on the removed campaign directory. A separate real CPU preflight using the unchanged locally archived runtime declaration still resolves all **22 collected op families** to 0.25.0 and keeps KDA on its preview pin. The fleet default is unchanged; the current tree does not automatically select the removed campaign's runtime.
- Keep all **17 Parquet payloads / 419,538 rows**, sidecars, failure counts, original collector refs and evidence hashes unchanged. The report now explicitly resolves its historical runtime/patch paths against immutable Git tree `54b073e140219b6e923a57a09acd05e96797a901`, not the current checkout. All five evidence artifacts were checked against both that Git tree and the local archive.
- Current-revision verification: **372 targeted collector/runtime/provenance tests passed**, **105 CODEOWNERS tooling tests passed**, plus Ruff, syntax, documentation/legal/copyright and whitespace checks. No GPU serving run, recollection, latest-corpus E2E replay or Full CI completion is claimed by this scope cleanup.

### Prior review evidence (historical revisions)

- **Retained data/NCCL fixes from `c62eb14`:** unsupported sidecar fields were moved to the report without widening the public schema; the real complete-tree validator passed **R1–R7**, and three public-loader regressions passed at that revision. NCCL power sampling stops in `finally` without masking the original benchmark failure. These fixes remain in the current tree.
- **Retired orchestration fixes:** `c62eb14` and `54b073e` fixed the optional runner/hook's clean-source attestation, import isolation, snapshot replacement, completeness validation and process cleanup. They were validated before archiving (134 and 155 targeted tests respectively; 34 runner/lifecycle cases passed five repeated runs at `54b073e`). Those helpers and their dedicated tests are now **out of this PR's shipped scope**, not advertised as current-tree functionality.
- Earlier Fast CI and public-loader results remain evidence for their recorded revisions, not automatic approval of the current head. Required exact-head checks and human approval remain separate.

## Why: FP8-block GEMM timing

The dtype-specific `use_cuda_graph=gemm_type != "fp8_block"` exception measured FP8-block GEMMs eagerly but published them under the same performance-table contract as graph-timed GEMMs. Host launch gaps can dominate small-M measurements and are then multiplied across model layers.

An isolated B200 A/B with 1965 MHz requested using the same `DeepGemmFp8BlockScaledMMKernel` instance/input/weights measured:

| Local shape (M,N,K) | Eager, 6 iterations (us) | One-op graph replay (us) | 32-op graph, per op (us) |
|---|---:|---:|---:|
| (1,768,7168) | 148.55 | 12.32 | 10.64 |
| (1,7168,384) | 152.77 | 8.21 | 5.64 |

All graph outputs were bitwise equal to eager. Generic FP8 and DeepSeek-V4 E8M0-scale configurations selected the same kernel and gave similar graph timings. This mechanism A/B used a vLLM 0.25.1 build; it is **not** a full supported-version collector sweep or a reproduction of every historical 0.24.0 value.

## Why: MoE workspace width

DeepSeek-V4-Pro has `hidden_size=7168`, while `num_attention_heads * head_dim = 128 * 512 = 65536`. MoE token workspace carries residual features, not the expanded attention tensor. Using 65536 inflated that workspace component by 9.14x and could consume the model's entire KV budget.

CPU single-variable A/B with the retained model configuration (8192 scheduled tokens):

| Topology | Original activation/workspace (GiB) | Corrected (GiB) |
|---|---:|---:|
| TP4 / ADP1 / MoE-TP4 / EP1 | 46.0000 | 13.9375 |
| TP1 / ADP8 / MoE-TP1 / EP8 | 58.0000 | 25.9375 |
| TP1 / ADP4 / MoE-TP4 / EP1 | 166.0000 | 37.7500 |

Only workspace width changed in this A/B; weights, base attention activation, NCCL and other memory were held fixed. The ADP8 false no-KV-budget failure disappeared. ADP4 still needs reconciliation of the scheduled-token limit and other runtime settings; this does not establish that all memory or TTFT discrepancies are solved.

## Validation on the combined branch

- Latest collector-code validation: **87 focused unit tests passed**, covering graph-only GEMM publication, FP8-block prepack, 0.25.0 metadata/inference state, MSA layout/getters, MLA contracts, exact shared-GEMM coverage, completion hooks, NCCL output validation and provenance.
- Source-isolated workspace validation, rechecked after syncing main: **18 passed, 1 native-runtime test skipped**. Together with the 87 collector tests, the current local regression run is **105 passed, 1 skipped**.
- Restoring each original GEMM timing / MoE workspace bug caused its behavioral regression test to fail; restoring the fixes passed.
- GPU results and the unresolved compute-scale finalization case are reported explicitly in Summary above. Do not interpret the partial smoke pass as a completed all-model campaign.
- CI tooling compatibility: carry the documentation checker/tests already present on main into this older PR branch and repair two stale documentation destinations. Collector code and performance data are unchanged by that CI repair; no policy check is bypassed.

## Data and rollout status

**This PR now includes newly recollected production performance tables:** 17 Parquet files / **419,538 rows**, published by data commit `c64ba870` (the DCO-signed version of the original data commit). Columns, units and logical schemas follow the previous packaged format; multi-event metadata retains each original collection's commit/hash/date/plan rather than relabeling it as the smoke-test revision.

- DSV4 attention/helper and mHC tables are included, as are the successful shared-expert GEMM supplement measurements.
- The all-model corpus is still **partial**: DSA context/generation and MSA generation were continued from their original checkpoints; compute-scale finalization and communication publication remain outstanding in this published batch. The committed coverage report is a frozen publication snapshot, not a live scheduler report.
- KDA explicitly reuses existing **`0.1.dev19262`** data by requester approval. It is not newly collected or relabeled as vLLM 0.25.0.
- No vLLM 0.25.1 mechanism-probe measurement is relabeled as 0.24.0 or 0.25.0 production data. Failed points are preserved; no synthetic replacements are inserted.
- The retained targeted-table replay is documented above. **A new E2E replay using the newly uploaded all-model corpus, or a new GPU serving benchmark, has not been completed.**
- `aisim-e2e-gym` code and the published dashboard have not been updated by this PR. No claim is made that the public E2E accuracy dashboard has been fixed.


YijiaZhao:
Collection follow-up: commit 5b6f57f fixes the inverse-CDF singularity at power-law exponent alpha=1 (the valid x^-1 case uses the logarithmic CDF), with five new CPU regression cases; other exponents retain the existing calculation. Combined focused validation: 51 passed, 1 native-runtime-dependent smoke test skipped locally; Ruff and diff checks pass. A bounded B200/vLLM 0.25.0 MoE retry completed all 28 remaining alpha=1/2 cases, alongside the 14 earlier alpha=0 cases. This does not yet constitute completed E2E validation or publication of production performance tables.

YijiaZhao:
# Full B200 / vLLM 0.25.0 recollection preflight — 2026-09-15

Status: **blocked before locked performance timing**. No full recollection is running and no new production tables are uploaded.

## Completed

- Full retained model plan: 28 architecture files, 20 ops, **418,628 tasks**.
- Registry-only auxiliary ops: **7,976 tasks**; communication is separate and not counted. Task count is not output-row count.
- Plan generation and module import/enumeration ran inside the actual B200 / vLLM 0.25.0 container. These are not GPU timing or compatibility qualification passes.
- Runtime reports vLLM 0.25.0 (`gdd10e03f9`), Torch 2.11.0+cu130, CUDA 13.0.
- Cached runtime artifact SHA256: `9ae776fd614406ea136d29aeb5ac05d9980865cd4a10eb8ec6956437da7d4ab7`.

## Clock gate — failed

- Direct `nvidia-smi -lgc 1965,1965`: permission denied.
- Host-side `sudo -n nvidia-smi -lgc 1965,1965`: requires a password; none was supplied.
- Independent Slurm probe requested 1800 MHz then 1965 MHz. All eight GPUs reported application graphics clocks of 1965 MHz in both steps. 1800 MHz is listed as supported.
- Therefore requesting `--gpu-freq=1965` and seeing the default/max 1965 MHz is **not evidence that clock control or a hard clock lock was applied**.
- Existing DSV4 diagnostic samples at active 1965 MHz must not be presented as proof of a hard-locked all-model campaign. They remain unpromoted candidate data.

## Remaining work

- Obtain administrator-enabled clock control or an authorized clock-controllable B200 allocation. Do not bypass privilege restrictions or substitute unlocked timings.
- Finish source/dispatch and representative runtime qualification for vLLM 0.25.0; existing compatibility declarations intentionally remain in force meanwhile.
- Then run the complete immutable plan in exclusive, resumable multi-node shards, record actual clocks, preserve failures, finalize provenance and publish validated data in this same PR.

## Per-op enumeration

| Op | Tasks | Scope |
|---|---:|---|
| `gemm` | 149,776 | full plan |
| `compute_scale` | 1,628 | full plan |
| `attention_context` | 54,468 | full plan |
| `attention_generation` | 67,636 | full plan |
| `encoder_attention` | 7,679 | full plan |
| `moe` | 3,069 | full plan |
| `mla_bmm_gen_pre` | 300 | full plan |
| `mla_bmm_gen_post` | 336 | full plan |
| `mla_context_module` | 13,176 | full plan |
| `mla_generation_module` | 13,248 | full plan |
| `dsa_context_module` | 35,136 | full plan |
| `dsa_generation_module` | 26,496 | full plan |
| `msa_context_module` | 17,568 | full plan |
| `msa_generation_module` | 13,248 | full plan |
| `dsv4_csa_context_module` | 5,840 | full plan |
| `dsv4_hca_context_module` | 5,840 | full plan |
| `dsv4_csa_generation_module` | 1,544 | full plan |
| `dsv4_hca_generation_module` | 1,544 | full plan |
| `dsv4_paged_mqa_logits_module` | 3,918 | auxiliary |
| `dsv4_hca_attn_module` | 3,918 | auxiliary |
| `mhc_module` | 140 | auxiliary |
| `gdn` | 84 | full plan |
| `kda` | 12 | full plan |

YijiaZhao:
Published the first verified vLLM 0.25.0 data batch in `a654f33`:

- 17 Parquet tables, **419,538 production rows**, retaining the prior directory/column/unit conventions.
- Original per-shard collector refs, hashes, case-plan hashes, dates, and observed failure counts are preserved in the existing multi-event metadata format. Smoke rows are not merged into production data.
- Existing KDA `0.1.dev19262` data are reused via an explicit 0.25.0 `reuse.yaml`; no KDA measurements were relabeled as 0.25.0.
- 87 focused unit tests passed. GPU smoke on collector-code commit `5515203` validated **25/26 batches, 148 cases**. The MLA campaign manifest was corrected from an inherited 0.27.0 family pin to the actual 0.25.0 image and both MLA smokes passed.
- The fresh `compute_scale` smoke measured 5/5 cases but exposed an existing finalization defect: its `scale_matrix_perf.txt` side output is not registered as an owned output. That family's outputs are withheld, not silently discarded or marked as passed.
- DSA context/generation and MSA generation are continuing only unattempted cases from existing checkpoints on 12 nodes. Existing successes and failures remain recorded. CAR/NCCL publication is still pending.

This is a **partial release**, not a claim of all-case success or completion. The packaged collection report, failure ledger, and README document the remaining scope and the approved frequency-request policy (`hard_lock_verified=false`).

python/aisimulate/src/aiconfigurator_core/systems/data/b200_sxm/gemm/vllm/0.25.0/collection_meta.yaml
```diff
@@ -0,0 +1,134 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+schema_version: 2
+provenance: local
+runtime:
+  framework: vllm
+  version: 0.25.0
+  image: aisim-b200-vllm-0250-sqsh-9ae776fd614406ea136d29aeb5ac05d9980865cd4a10eb8ec6956437da7d4ab7
+  source_commit: dd10e03f95f94edbea1975c67ace3a35ec9a8a40
```
jasonqinzhou: The repository's fail-closed data validator rejects this field in every one of the 10 new schema-v2 sidecars: `runtime has unsupported key(s): source_commit`. I reproduced this with `PYTHONPATH=python/aisimulate/src:python/aisimulate python/aisimulate/.venv/bin/python python/aisimulate/tools/perf_database/check_collector_data.py`, which exits 1 with ten R1 failures. Full CI's collector-data gate will therefore fail, and strict provenance loads cannot accept these tables. Please either encode the upstream revision in a schema-supported location or extend the schema and its tests deliberately, then rerun the validator for the complete data tree.

python/aisimulate/collector/campaigns/run_shard.py
```diff
@@ -0,0 +1,218 @@
...
+    command = [
+        sys.executable,
+        "-u",
+        str(root / "source/python/aisimulate/collector/collect.py"),
+        "--backend",
+        "vllm",
+        "--ops",
+        spec["op"],
+        "--sm",
+        "100",
```
jasonqinzhou, line 397: This runner invokes the stock `collect.py`, but the exact recorded checkout's `framework_manifest.yaml` rejects vLLM 0.25.0 for every published family I sampled: gemm/attention/moe/msa resolve to 0.24.0, MLA to 0.27.0, and KDA to 0.1.dev19262. Calling `require_collector_runtime(..., '0.25.0', requested_ops={op})` raises for each, so a clean checkout at the recorded `collector_ref` cannot reproduce these jobs. Please make the runner consume a committed, hash-recorded 0.25.0 runtime declaration (and verify the source tree is clean and at `manifest['source_commit']`) so provenance does not depend on an unrecorded campaign-side patch.

python/aisimulate/collector/campaigns/run_shard.py
```diff
@@ -0,0 +1,218 @@
...
+    spec = next(s for s in manifest["smokes"] + manifest["shards"] if s["id"] == args.shard_id)
+    base = root / "results" / args.shard_id
+    job = base / ("job-" + os.environ["SLURM_JOB_ID"])
+    job.mkdir(parents=True, exist_ok=True)
+    data = Path(manifest["local_root"]) / args.shard_id / "data"
+    data.mkdir(parents=True, exist_ok=True)
+    previous = base / "latest_snapshot.json"
+    if previous.exists():
+        snapshot = root / json.loads(previous.read_text())["relative_path"]
+        shutil.copytree(snapshot, data, dirs_exist_ok=True)
```
jasonqinzhou: `dirs_exist_ok=True` overlays the last canonical snapshot onto whatever an interrupted attempt left in node-local `data`. Stale checkpoints, Parquet files, or transaction markers that are absent from the snapshot therefore survive and can be resumed or republished as though they belonged to the canonical state. Replace or clear the shard data directory before copying the selected snapshot so the resume input is exactly that snapshot.

python/aisimulate/collector/campaigns/slurm_collection_hook.py
```diff
@@ -0,0 +1,319 @@
...
+    meta = yaml.safe_load((data / "collection_meta.yaml").read_text())
+    if meta["runtime"]["version"] != "0.25.0" or meta["runtime"]["framework"] != "vllm":
+        raise ValueError("Performance provenance belongs to another framework/version")
+    if table_reader is None:
+        import pyarrow.parquet as pq
+
+        table_reader = lambda path: pq.read_table(path).to_pylist()
+    files = {}
+    row_count = 0
+    for path in sorted(data.glob("*_perf.parquet")):
```
jasonqinzhou: Validation iterates only Parquet files that are present. If `collection_meta.yaml` declares two completed tables and one file is absent, the remaining file still makes `files` and `row_count` truthy, so the shard validates while silently dropping a declared artifact. Compare the declared table-name set with the observed Parquet stem set before validating rows and fail on any mismatch.

python/aisimulate/collector/campaigns/slurm_collection_hook.py
```diff
@@ -0,0 +1,319 @@
...
+        rows = table_reader(path)
+        entry = meta["tables"][path.stem]
+        if spec["mode"] == "full":
+            expected_ref = status.get("source_commit")
+            if not expected_ref or expected_ref == "unknown" or entry.get("collector_ref") != expected_ref:
+                raise ValueError("Full-data provenance does not match the recorded source commit")
+        if entry.get("status") != "complete" or entry["rows"] != len(rows):
+            raise ValueError("Parquet row count/status does not match its provenance")
+        for row in rows:
+            if row["version"] != "0.25.0" or not math.isfinite(row["latency"]):
```
jasonqinzhou: A null or nonnumeric `latency` raises `TypeError` inside `math.isfinite`, but the per-shard handler catches only `OSError`, `ValueError`, and `KeyError`. One malformed shard therefore aborts the whole scan before unrelated ready shards are processed. Validate the value's numeric type and raise `ValueError` so this shard becomes `validation_failed` without stopping the campaign observer.

python/aisimulate/collector/network/collect_nccl.py
```diff
@@ -83,17 +105,19 @@ def nccl_benchmark(
         if power_monitor:
             power_monitor.start_sampling()
 
-        result = subprocess.run(cmd_args, capture_output=True, text=True)
+        result = subprocess.run(
```
jasonqinzhou: With `check=True`, a benchmark failure raises before `PowerMonitor.stop_sampling()` runs, leaving its background sampling thread active until process teardown. Put the stop call in a `finally` block around `subprocess.run` while preserving the original subprocess error.

jasonqinzhou (changes requested):
**REQUEST_CHANGES · 5.6/10 · high confidence**

## What this PR does

This PR qualifies vLLM 0.25.0 collector paths, corrects CUDA-graph GEMM timing and DeepSeek-V4 MoE workspace sizing, and publishes a documented partial B200 corpus of 17 Parquet tables totaling 419,538 rows.

The publication report is unusually candid and useful: it preserves per-shard hashes and failures, labels the corpus as partial, separates retained accuracy evidence from new collection evidence, and explicitly withholds compute-scale output after finalization failed.

## Why this score

The core modeling and timing fixes are well tested, but the exact head cannot pass the repository's packaged-data validator because every new schema-v2 sidecar uses an unsupported runtime field. The committed campaign runner also cannot reproduce its declared vLLM 0.25.0 jobs from the pinned source checkout, and the resume/validation helpers retain several reachable data-integrity and availability gaps. Those issues make the publication unsafe to accept despite 131 focused tests and internal table consistency checks passing.

python/aisimulate/src/aiconfigurator_core/systems/data/b200_sxm/gemm/vllm/0.25.0/collection_meta.yaml
YijiaZhao: Fixed in c62eb14 without extending the consumer schema. Removed the unsupported runtime.source_commit and per-event classified_failures fields from the ten v2 sidecars; the upstream runtime revision and per-shard failure counts are retained in the adjacent collection report, with the original sidecar hashes. All 17 Parquet SHA-256 values are unchanged.

I reproduced the ten R1 failures using the real public checker against the complete data tree, then reran it after the fix: R1–R7 all pass. Three public-loader regressions pass, including discovery of all ten directories and assertions that unsupported fields are still rejected. No SDK/parser relaxation was made.

python/aisimulate/collector/campaigns/run_shard.py
YijiaZhao: Addressed in c62eb14. The runner now requires a committed runtime-manifest path plus SHA-256, checks the exact source HEAD and a clean worktree (including untracked patches), verifies the declaration against its Git blob, and places that source first on the child's PYTHONPATH. The child receives a hash-checked runtime declaration through an explicit loader interface; the fleet default is unchanged. The committed B200 declaration resolves all 22 collected op families to 0.25.0 (including MLA), while KDA remains pinned to its preview version.

A real clean-checkout attestation and child-process resolution passed for all 22 ops at this commit. Negative tests cover dirty/wrong sources, missing declarations, hash mismatches and tampering.

Historical-data qualification: I did not relabel old measurements or claim those old checkouts were clean. Their exact original runtime/link patches are now committed under historical_runtime_overrides/ and hash-bound to each original source ref in the publication report. The new hardened runner does not apply those patches; new runs require a clean revision with the declaration interface. This distinction and migration requirement are explicit in the README, for your re-review.

python/aisimulate/collector/campaigns/run_shard.py
YijiaZhao: Fixed in c62eb14. Restore now copies the selected canonical snapshot into a fresh temporary directory, quarantines existing local data, and replaces the destination rather than overlaying it. A fresh attempt also cannot inherit unpublished local state. Snapshot publication uses the same exact-copy rule. Regression tests cover stale checkpoints/Parquet/transaction files, preservation of orphan evidence, and preserving the old directory when copying fails.

python/aisimulate/collector/campaigns/slurm_collection_hook.py
YijiaZhao: Fixed in c62eb14. Validation compares the complete declared table-name set with observed Parquet stems before reading rows and fails on missing or undeclared artifacts. Malformed table metadata is also converted to a shard validation error. Tests cover both missing declared tables and extra unclaimed files.

python/aisimulate/collector/campaigns/slurm_collection_hook.py
YijiaZhao: Fixed in c62eb14. Latency must be a real numeric value (booleans excluded) before math.isfinite is called. Nulls, strings, containers, NaN and infinity fail as ValueError. Added a tick-level regression proving that an invalid shard becomes validation_failed while a later unrelated ready shard is still submitted; malformed YAML/metadata are likewise isolated.

python/aisimulate/collector/network/collect_nccl.py
YijiaZhao: Fixed in c62eb14. PowerMonitor.stop_sampling() is now in finally around subprocess.run(check=True). The original benchmark exception is preserved even if sampling cleanup also raises, and no performance row is published on that path. Both ordinary subprocess failure and simultaneous cleanup failure are covered by behavioral tests.

python/aisimulate/collector/campaigns/run_shard.py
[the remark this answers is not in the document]
YijiaZhao: Fixed in c62eb14. Restore now copies the selected canonical snapshot into a fresh temporary directory, quarantines existing local data, and replaces the destination rather than overlaying it. A fresh attempt also cannot inherit unpublished local state. Snapshot publication uses the same exact-copy rule. Regression tests cover stale checkpoints/Parquet/transaction files, preservation of orphan evidence, and preserving the old directory when copying fails.

python/aisimulate/collector/campaigns/slurm_collection_hook.py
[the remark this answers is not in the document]
YijiaZhao: Fixed in c62eb14. Validation compares the complete declared table-name set with observed Parquet stems before reading rows and fails on missing or undeclared artifacts. Malformed table metadata is also converted to a shard validation error. Tests cover both missing declared tables and extra unclaimed files.

python/aisimulate/collector/campaigns/slurm_collection_hook.py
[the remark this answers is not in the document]
YijiaZhao: Fixed in c62eb14. Latency must be a real numeric value (booleans excluded) before math.isfinite is called. Nulls, strings, containers, NaN and infinity fail as ValueError. Added a tick-level regression proving that an invalid shard becomes validation_failed while a later unrelated ready shard is still submitted; malformed YAML/metadata are likewise isolated.

python/aisimulate/collector/network/collect_nccl.py
```diff
@@ -83,17 +105,19 @@ def nccl_benchmark(
...
+            cmd_args,
+            capture_output=True,
+            text=True,
+            check=True,
+            env={**os.environ, "NCCL_DEBUG": "VERSION"},
+        )
```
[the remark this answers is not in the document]
YijiaZhao: Fixed in c62eb14. PowerMonitor.stop_sampling() is now in finally around subprocess.run(check=True). The original benchmark exception is preserved even if sampling cleanup also raises, and no performance row is published on that path. Both ordinary subprocess failure and simultaneous cleanup failure are covered by behavioral tests.

python/aisimulate/collector/campaigns/run_shard.py
```diff
@@ -0,0 +1,312 @@
...
+def collector_environment(root, runtime_declaration):
+    """Bind helper imports and runtime selection to the attested source."""
+    source_python = str((Path(root) / "source/python/aisimulate").resolve())
+    previous_pythonpath = os.environ.get("PYTHONPATH", "")
+    return {
+        **os.environ,
+        "AISIM_COLLECTOR_RUNTIME_MANIFEST": runtime_declaration["path"],
+        "AISIM_COLLECTOR_RUNTIME_MANIFEST_SHA256": runtime_declaration["sha256"],
+        "PYTHONDONTWRITEBYTECODE": "1",
+        "PYTHONPATH": os.pathsep.join(part for part in (source_python, previous_pythonpath) if part),
```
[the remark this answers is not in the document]
YijiaZhao: Fixed in 54b073e140219b6e923a57a09acd05e96797a901.

`collector_environment()` now gives the child only the resolved attested `source/python/aisimulate` in `PYTHONPATH`; parent entries are not appended. It also removes inherited `PYTHONHOME`, `PYTHONUSERBASE`, and `PYTHONSTARTUP` and sets `PYTHONNOUSERSITE=1`. Runtime dependencies must be installed in the selected trusted interpreter/venv/image; this is documented, not represented as an OS sandbox.

The updated environment assertion and a real CPU child-process regression verify that a parent `sitecustomize.py` is not executed, a parent-only module cannot be imported, and the attested-source module remains importable. The combined runner/hook/runtime/NCCL suite passes **155 tests** on Python 3.12/macOS; the 34 runner/lifecycle cases also passed five consecutive runs. Ruff check/format and local policy checks pass.

No collector executor, case IDs, performance schemas, or measured values changed; all 17 published Parquet SHA-256 digests are unchanged. This is not a new GPU or E2E qualification.

python/aisimulate/collector/campaigns/run_shard.py
```diff
@@ -0,0 +1,312 @@
...
+    save(status_path, status)
+    review = manifest.get("reviewed_failure_continuation", {})
+    reviewed = spec["op"] in review
+    status["failure_review"] = review.get(spec["op"])
+    finalized_since = None
+    env = collector_environment(root, runtime_declaration)
+    with (job / "collector.log").open("w") as log:
+        proc = subprocess.Popen(
+            command, cwd=data, stdout=log, stderr=subprocess.STDOUT, start_new_session=True, env=env
+        )
```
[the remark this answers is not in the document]
YijiaZhao: Fixed in 54b073e140219b6e923a57a09acd05e96797a901.

The collector is now owned by a context manager covering startup and the entire monitoring lifecycle. Its `finally` path sends SIGTERM to the task-owned process group, waits with a deadline, escalates to SIGKILL when necessary, and reaps the direct child. It checks for live group members even after the leader exits and fails closed if writers remain; snapshots are published only after this returns successfully. No global process-name kill is used.

SIGTERM/SIGINT during startup are deferred until the child handle is owned, then propagated through cleanup. Subsequent signals cannot interrupt cleanup. Checkpoint, worker-inspection, or status-write failures record `runner_failed` after cleanup and re-raise the original exception; secondary cleanup/reporting failures are retained without masking it. Uncatchable SIGKILL/host loss remains a scheduler-level cleanup responsibility.

Added **21 tests** (including parametrized cases), using real CPU subprocesses for ordinary/stubborn children, an exited leader with a surviving worker, runner SIGTERM, startup signal races, normal completion and walltime yield. Failure-injection cases cover checkpoint/inspection/status/reporting failures and a group that remains active after SIGKILL. A macOS exit-time EPERM race is handled only after checking that no live group members remain, not by swallowing permission errors.

Validation: **155 targeted tests passed** on Python 3.12/macOS; the **34 runner/lifecycle cases passed five consecutive runs**; **105 CODEOWNERS tooling tests passed**, plus Ruff and documentation/legal/copyright checks. All 17 Parquet payload hashes are unchanged. This does not claim a new GPU run, Full CI completion, or human approval.

YijiaZhao:
@jasonqinzhou Scope update at requester direction, committed in bdc9eec9e8d4bcb9b9f2b06b239450739a7b4f78:

- Removed the optional `collector/campaigns/` runner/hook and their three dedicated test modules from this PR. They have been backed up locally with byte-for-byte verification for future reuse; **no new GitHub archive directory, branch or package** is published. Runner/hook-specific review findings are now outside the current PR scope, rather than being claimed as current-tree functionality.
- Retained the data-schema compatibility fix, NCCL cleanup, operator fixes, and generic hash-checked external runtime-manifest interface. Runtime-interface tests no longer import/read the removed campaign directory.
- All **17 Parquet payloads / 419,538 rows**, sidecars, original collector refs, failure counts and evidence digests are unchanged. Historical runtime/patch paths in the report are explicitly qualified with immutable Git tree `54b073e140219b6e923a57a09acd05e96797a901`; all five evidence artifacts were verified against that tree and the local archive. They are not presented as paths in the current checkout.
- **372 targeted collector/runtime/provenance tests passed**, **105 CODEOWNERS tooling tests passed**, plus Ruff/syntax/documentation/legal/copyright/whitespace checks. A CPU preflight with the real archived runtime declaration still resolves all 22 collected ops to 0.25.0 and leaves KDA on its preview pin. No fresh GPU measurement or E2E replay is claimed.

Summary updated to distinguish retained product fixes, historical orchestration validation and the current scope. The DSV4 accuracy report and its qualifications are unchanged. This is not a claim of Full CI completion or human approval.

Arsene12358 (approved):

jasonqinzhou (approved):
Re-reviewed bdc9eec9e8d4bcb9b9f2b06b239450739a7b4f78. The previous metadata-schema and NCCL cleanup findings are fixed, and the optional campaign tooling has been removed from scope. All review threads are resolved; no remaining actionable code finding was identified.

Validation: 153 focused tests passed; the complete packaged-data validator passed R1–R7; all 17 Parquet tables (419,538 rows) passed hash, version, identity-uniqueness, and positive-finite-latency checks; all five historical evidence hashes matched. The public loader and native table view returned the exact published shared-expert GEMM measurement. Local validation reused a native extension from a checkout with an identical Rust crates tree.

Approval is for the revised code and documented partial data publication. Full CI must pass for this exact head before merge; only Fast CI is currently complete. The retained accuracy comparison does not establish end-to-end qualification of the newly published full corpus.

PR #220, opened 2026-09-15
fix(memory): use residual width for DeepSeek-V4 MoE workspace

YijiaZhao:
> **Superseded by #219.** Both fixes now ship in one PR at the requester's direction. This PR is closed without merging; its implementation is included in #219.

## Summary

Use the residual hidden width for DeepSeek-V4 MoE dispatch workspace accounting in the shared TRT-LLM hook. VLLMBackend aliases this hook, so both backends receive the correction.

The change is deliberately narrow:

- `DEEPSEEKV4` uses the model's `_hidden_size` rather than its expanded attention width.
- The existing attention activation term is unchanged.
- Legacy `DEEPSEEK` behavior and the existing GEMMA4MIX/STEP3P7 exceptions are preserved.
- No measured performance tables, collector logic, or scheduler behavior are modified.

## Why

DeepSeek-V4-Pro has `hidden_size=7168`, while `num_attention_heads * head_dim = 128 * 512 = 65536`. The MoE token workspace carries residual features, not the expanded attention tensor. Using 65536 inflated that workspace component by 9.14x and could consume the model's entire KV budget.

CPU single-variable A/B with the retained model configuration (8192 scheduled tokens):

| Topology | Original activation/workspace (GiB) | Corrected (GiB) |
|---|---:|---:|
| TP4 / ADP1 / MoE-TP4 / EP1 | 46.0000 | 13.9375 |
| TP1 / ADP8 / MoE-TP1 / EP8 | 58.0000 | 25.9375 |
| TP1 / ADP4 / MoE-TP4 / EP1 | 166.0000 | 37.7500 |

Only the workspace width changed; weights, base attention activation, NCCL and other memory were held fixed. The ADP8 false no-KV-budget failure disappeared. ADP4 can still require reconciliation of the scheduled-token limit and other runtime settings, so this is not a claim that all memory or TTFT discrepancies are solved.

## Validation

- 18 CPU source-isolated regression cases passed; 1 installed-native-backend smoke test skipped locally because the native runtime is not installed.
- The source-isolated tests execute the real memory method, class constants and vLLM hook alias, without reimplementing the memory formula. Native-installed coverage is also included for CI.
- Restoring the old exception set makes the new tests fail; restoring the fix passes.
- Pro and Flash geometries are covered; attention activation and non-workspace memory terms are checked separately.
- Ruff check/format and `git diff --check` pass.

Local command:

```bash
python -m pytest -q -p no:timeout \
  --confcutdir=python/aisimulate/tests/unit/sdk/backends \
  python/aisimulate/tests/unit/sdk/backends/test_deepseek_v4_workspace_memory.py
```

## Scope

SDK/modeling change only. No new GPU workload or full E2E replay was run for this patch. The impact on a complete request-level TTFT replay remains to be measured. This change has been consolidated with the GEMM collection fix in #219 at the requester's explicit direction.


YijiaZhao:
Consolidated into #219 at the requester’s explicit direction so the GEMM CUDA Graph timing and DeepSeek-V4 MoE workspace corrections are reviewed and landed together. The memory fix is included in #219 (commit 2fce86f); the combined branch passed 46 focused tests with 1 native-runtime-dependent smoke test skipped. Closing this duplicate without merging; no branch deletion.

PR #234, opened 2026-09-16
test(replay): preserve Weka future-parent rekey regression

Harrilee:
## Summary

Restore the Weka regression removed when Dynamo delegated ingestion to AISimulate (ai-dynamo/dynamo#14355). When seam resolution chooses a later request with a deeper prefix overlap as a continuation, an earlier fork must keep its original parent; reparenting it to that future continuation creates a cycle.

The test uses AISimulate's public graph loader, verifies the continuation was selected, and checks that the earlier fork retains its original dependency. Source attribution records the immutable Dynamo revision in the test file and both packaged notice copies.

## Validation

- `cargo test --locked -p aisimulate-core --lib replay::loadgen::weka::tests::` — 27 passed.
- Mutation check: removing the future-parent guard makes the regression fail with a cycle between the earlier fork and later continuation; the guard was restored afterward.
- `rustfmt --check --edition 2024 crates/core/src/replay/loadgen/weka.rs`
- `python3 scripts/check_packaged_legal_files.py`
- `git diff --check`


Harrilee:
Checked the failed run [35149065556](https://github.com/ai-dynamo/aisimulate/actions/runs/35149065556) against the current merged head `93356ae0`.

The old failures were collector registry drift (49 differences) and stale GEMM case-count, DeepSeek-V4 FPM capability, and B200/SGLang GDN expectations, in both architecture lanes. They were unrelated to the added Weka test. The main merge already brought in the relevant fixes; no additional code change was necessary.

Fresh local validation at `93356ae0`, with its native extension rebuilt:
- Previously failing Python test selections: **5 passed** (including the current parameterized FPM cases).
- Collector data invariants: **all rules passed**.
- Backend facts: **2,139 fact slices match**.
- Weka Rust tests: **27 passed**, including the new regression.
- Packaged legal files match.

[New Full CI](https://github.com/ai-dynamo/aisimulate/actions/runs/35947986473) is still running, with no failures observed at this check. This is not a claim that hosted CI has completed.

Scope clarification for ai-dynamo/dynamo#14355: this PR only restores a regression test and its attribution. The runtime future-parent guard already exists in the pinned nightly; there is no runtime fix here that Dynamo must wait to consume in a new release.

PR #189, opened 2026-09-11
ci(aic): add advisory forward prediction gate

jasonqinzhou:
Adds an advisory workflow that benchmarks forward-prediction performance for relevant PR changes and publishes comparison results and artifacts. It runs against the trusted canonical PR commit, using the base revision's controller and thresholds with separate workers, wheels, and environments for each revision.

This replaces #75, which was automatically closed when its stacked base branch was deleted after #71 merged. The three original commits retain jthomson04's authorship and are cherry-picked onto current `main`; `.github/workflows/performance.yml` is byte-for-byte identical to #75 at `c453ebe7`.

The workflow includes the four reviewed fixes: protocol compatibility skips, a shared push/manual concurrency group per PR, guarded cleanup, and narrow path filters. The performance check remains advisory. Build caching and shared build-action refactoring remain outside this change.

Validation for head `885df8291369d0e5daa3eadde11cca119b02c0ae`, based on `main` at `d1f4a474288525ab99890477c9e5b42ddb4e481e`:

- One workflow file added; clean diff and no conflict with current `main`.
- actionlint 1.7.12, YAML parsing, and all eight embedded shell syntax checks pass.
- 29 existing CI workflow contract tests pass.
- The actual workflow preparation step selected the expected base and head, created both detached worktrees, and the actual cleanup step removed both successfully.
- Fresh self-review of the complete diff and its benchmark consumers found no actionable issue.
- Direct Fast CI passed on the replacement head.
- [Advisory benchmark](https://github.com/ai-dynamo/aisimulate/actions/runs/34644496577) passed against current `main`: all 36 cases and five paired rounds, 72 comparisons, 53 stable, 19 noisy, and zero confirmed regressions, invalid/skipped comparisons, or run errors. [Full artifact](https://github.com/ai-dynamo/aisimulate/actions/runs/34644496577/artifacts/10282121931).
- [Full CI](https://github.com/ai-dynamo/aisimulate/actions/runs/34644496969) is still running on this exact head; 25 jobs have passed with no failures at the latest check.
- CodeRabbit review is in progress.

Original review discussion and historical benchmark evidence remain linked from #75. This replacement will require its own independent review approval.

PR #236, opened 2026-09-16
ci(perf): expand coverage and fix automatic PR runs

jthomson04:
## Summary

Expand advisory prediction-performance CI from 36 to 64 cases (128 cold/warm comparisons), and fix automatic runs when copy-pr-bot creates a trusted branch with an empty push commit list. A hosted selector reads the complete PR diff before allocating a benchmark runner.

- Keep the original 36 Qwen requests, IDs, values, order, and cache groups. Add five model families, prefix cases, DeepSeek attention-DP cases, and Qwen3.5 on SGLang 0.5.14. The other new profiles use vLLM 0.24.0; Nemotron uses H100 and the other models use B200.
- Keep protocol v1, fixed priming, five paired rounds, ten warmups, 100 warm calls, and current regression thresholds. No public API, request-schema, or CLI changes.
- Select from all PR files, including rename sources, so documentation-only follow-up commits still run when the PR contains relevant code. Gate README-only PRs skip benchmarking. Manual runs must match the current PR head; API errors fail selection explicitly.
- When gate Python files or the shared grid change, run both the normal base controller and the PR controller against the same builds. This validates the expanded matrix before merge.
- Release the outgoing database when system/backend/version changes, including transitions after setup or priming failures. Preserve the incoming reset and behavior for consecutive groups using the same database.
- Use named DeepSeek V3.2 and Qwen3.5 configurations for special scenarios; check priming exclusion against the full matrix. Clarify profile labels and prefix worker-group separation.

## Review map

- Risk: low; matrix, benchmark-worker cache lifecycle, workflow selection, tests, and documentation.
- Start with `cases.py` and `worker.py`, then `scripts/select_forward_perf.py` and `.github/workflows/performance.yml`.
- The normal workflow uses the base controller. This PR therefore runs 36 base cases plus a separate 64-case PR-controller validation. The normal matrix becomes 64 cases once the merge base contains this change.
- Keep the agreed scope: no shared Full CI helper, report-grouping redesign, selector-security redesign, threshold changes, or minor diagnostic cleanup.

## Validation

Current head: `a9867cf03fd49e617d7b0ca17a9dde12f43af339`; merge base: `dfa8a1fdac8374058b91d797e8a8bc433b05f406`.

Rebased onto current main, including the license-file fix from #241 and the nightly validation fix from #250. All four PR patches are unchanged (`git range-diff`); all 64 serialized requests, IDs, values, and ordering are unchanged. No conflict resolution or extra fix commit was needed.

- **277 focused tests passed**: worker/comparison tests and workflow contracts.
- **207 workflow contracts passed with warnings treated as errors**, including the nightly-license tests that failed on the previous base. The test environment includes the new `pip-licenses==5.5.5` dependency from main.
- Ruff lint/format, actionlint, and whitespace checks passed. All four commits retain DCO sign-offs and valid signatures.
- Behavioral coverage includes database transitions after success/setup failure/priming failure, shared-grid and gate-Python triggers, README-only skipping, relevant full-PR selection after a documentation-only follow-up, scenario model identity, and full-matrix priming exclusion.

[Fast CI](https://github.com/ai-dynamo/aisimulate/actions/runs/35254117728), [Full CI](https://github.com/ai-dynamo/aisimulate/actions/runs/35254113366), and [automatic performance CI](https://github.com/ai-dynamo/aisimulate/actions/runs/35254112836) all passed on this exact head. Both previously failing contract shards passed: 1,851 tests and 132 subtests per architecture, with five expected skips. The old nightly-license file-warning failure is cleared by the fix in main.

Both controller result sets were checked from the [uploaded artifact](https://github.com/ai-dynamo/aisimulate/actions/runs/35254112836/artifacts/10512099403):

| Controller | Cases | Comparisons | Rounds | Regressions | Invalid / skipped | Cold / warm unstable | Benchmark time |
|---|---:|---:|---:|---:|---|---|---:|
| Base controller | 36 | 72 | 5 | 0 | 0 / 0 | 14 / 0 | 53 s |
| PR controller | 64 | 128 | 5 | 0 | 0 / 0 | 29 / 0 | 135 s |

Every availability probe and measured response was `OK`. The expanded matrix took 82 seconds more than the base matrix; the separate pre-merge controller pass added 135 seconds. Both are below the five-minute limit. Artifact upload and cleanup passed.

Prior qualification of the unchanged controller and worker at `15744612` used two separate installations of the same CI wheel from `52daaef0`. Dependencies and native-library hashes matched. Runs were interleaved with the original 36-case matrix on CPU 10; times exclude builds and installation. These are pre-rebase measurements; the current-head CI evidence above is separate.

| Run | Original time | Expanded time | Cold unstable | Warm unstable | Regression flags |
|---|---:|---:|---:|---:|---:|
| 1 | 33.88 s | 83.67 s | 37 | 0 | 0 |
| 2 | 32.36 s | 83.84 s | 9 | 0 | 0 |
| 3 | 32.55 s | 84.00 s | 17 | 0 | 0 |

Every expanded run completed 64/64 cases, five paired rounds, and 128 valid comparisons, with no missing data, skipped cases, invalid comparisons, or regression flags. No repeated regression flag required investigation. Median added benchmark time was 51.29 seconds, below the five-minute limit. Cold noise remains visible; same-wheel timing improvements are not evidence of a product speedup.

Uses bundled model configurations and checked-in performance data. Nemotron uses `nvidia/NVIDIA-Nemotron-3-Super-120B-A12B-FP8`. SGLang 0.5.14 is used because 0.5.16 lacks the required B200 custom-all-reduce data. Measurements cover CPU prediction-call cost.

## Tracking

Related: #189.


jthomson04:
@coderabbitai full review

scripts/select_forward_perf.py
```diff
@@ -0,0 +1,165 @@
...
+            raise ValueError(f"Invalid trusted PR ref: {ref}")
+        pr_number = match[1]
+    elif event != "workflow_dispatch" or not re.fullmatch(r"[1-9][0-9]*", pr_number):
+        raise ValueError("Expected a trusted PR push or manual dispatch with a positive pr_number")
+
+    endpoint = f"repos/{repository}/pulls/{pr_number}"
+    pull = api(endpoint)[0]
+    if event == "workflow_dispatch":
+        sha = api(f"repos/{repository}/git/ref/heads/pull-request/{pr_number}")[0]["object"]["sha"]
+    if not re.fullmatch(r"[0-9a-f]{40}", sha) or sha != pull["head"]["sha"]:
```
jthomson04, line 99: **Manual dispatch now fails whenever the trusted copy lags the PR head.**

The old `actions/github-script` path read `git/ref/heads/pull-request/N` and benchmarked that SHA, with no comparison against `pull.head.sha`. copy-pr-bot does not copy new commits from outside contributors until a maintainer re-approves the new SHA, so the copied branch routinely lags.

Dispatching for a PR in that state: `git/ref` returns commit N, `pulls/N` returns head N+1, this check raises, `main()` returns 1, `select` fails and `compare` is skipped. The manual escape hatch that previously worked now produces a red workflow and no benchmark. The same assertion also turns a superseded push into a failed advisory run if the author pushes again during the selection window.

Consider scoping the head-match check to the `push` event and letting `workflow_dispatch` benchmark the trusted copy as before.

.github/workflows/performance.yml
```diff
@@ -178,6 +148,13 @@ jobs:
             exit 0
           fi
 
+          controller_changes="$(git diff --name-only "${base_sha}" "${PR_HEAD_SHA}" -- "${gate_path}")"
```
jthomson04: **Head-matrix validation misses `grid.py`.**

`gate_path` covers only `python/aisimulate/tools/forward_perf_gate`, but `cases.py` imports `prediction_regression_gate.grid` and reads `PREFILL_POINTS`, `DECODE_POINTS`, `CTX_OSL`, `GEN_OSL` and `STRIDE` -- every coordinate of all 64 head cases.

A PR that edits `grid.py` rewrites the case IDs and coordinates, yet this `git diff` comes back empty, `validate_head_controller=false`, and the proposed matrix never runs before merge. The next PR then hits it as the normal comparison and discovers a DATA_MISS/INVALID matrix post-merge -- exactly the failure mode this feature exists to prevent.

Note that `select_forward_perf.py` already lists `python/aisimulate/tools/prediction_regression_gate/grid.py` in `PATH_PATTERNS`; `gate_path` should match.

python/aisimulate/tools/forward_perf_gate/cases.py
```diff
@@ -58,4 +115,16 @@ def expand_cases() -> list[dict]:
                         "stride": grid.STRIDE,
                     }
                 )
+
+    for model in ADDITIONAL_MODELS:
+        result.extend(_silicon_cases(model, ADDITIONAL_MODEL_POINTS))
+    for model in MODELS:
+        result.extend(_silicon_cases({**model, "model_id": f"{model['model_id']}-prefix"}, PREFIX_POINTS))
+    result.extend(_silicon_cases({**ADDITIONAL_MODELS[0], "tp_size": 1, "attention_dp_size": 8}, ATTENTION_DP_POINTS))
```
jthomson04: **Positional model selection is unpinned by any test.**

`ADDITIONAL_MODELS[0]` here and `ADDITIONAL_MODELS[2]` on line 126 bind by index. Insert a sixth model at the front of the tuple, or alphabetize it, and the attention-DP cases silently retarget from DeepSeek V3.2 and the SGLang cases from Qwen3.5-397B-A17B to whatever now sits at index 2 -- likely a model with no B200 SGLang 0.5.14 data, i.e. DATA_MISS.

Nothing catches this: `test_additional_profiles_preserve_original_cache_groups` only asserts `(batch_size, isl, tp_size, moe_ep_size)` for the DP cases, and the frozen sha256 covers `expanded[:36]`. Select by name instead, e.g. `next(m for m in ADDITIONAL_MODELS if m["model_id"] == "deepseek-v3.2")`.

.github/workflows/performance.yml
jthomson04: **Doc-only edits under the gate directory trigger a second full 64-case run.**

A PR touching only `forward_perf_gate/README.md` matches both `matches_path` (`.../forward_perf_gate/**`) and this diff, so the job runs the 36-case base comparison *and* the 64-case head-controller validation. Per this PR's own data that is ~134 s of extra benchmark time plus a second prewarm pass over 13 database loads, on the scarce `prod-aisimulate-default-amd64-v1` runner, to validate a matrix that did not change.

Narrow `gate_path` to the files that actually define the matrix and protocol (`cases.py`, `worker.py`, `run.py`, `compare.py`, `__init__.py`), or exclude `*.md`.

scripts/select_forward_perf.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+def github_api(endpoint: str) -> list:
+    """Read all pages; an API error must not become a successful skip."""
+    result = subprocess.run(
+        ["gh", "api", "--paginate", "--slurp", endpoint],
+        capture_output=True,
+        text=True,
+        check=True,
+        timeout=60,
```
jthomson04, line 69: **An API timeout or transient error fails selection instead of running conservatively.**

`docs/ci.md` promises that empty, incomplete, or oversized file lists run the benchmark conservatively, and lines 105-113 implement exactly that. But a PR with ~2,900 changed files needs ~29 paginated requests; if this `timeout=60` fires part-way through, `subprocess.TimeoutExpired` (a `SubprocessError`) is caught in `main()`, which returns 1. `select` fails, `compare`'s `needs.select.outputs.run_comparison == 'true'` guard evaluates false, and the benchmark is silently skipped.

The conservative fallback should cover file-list retrieval failure too, not just short lists.

python/aisimulate/tools/forward_perf_gate/cases.py
```diff
@@ -32,6 +32,63 @@
...
+    {"model_id": "gpt-oss-120b", "model_path": "openai/gpt-oss-120b"},
+    {
+        "model_id": "nemotron-3-super-120b-fp8",
+        "model_path": "nvidia/NVIDIA-Nemotron-3-Super-120B-A12B-FP8",
+        "system_name": "h100_sxm",
+    },
+)
+
+# (phase, batch size, input length, cached prefix length).
+ADDITIONAL_MODEL_POINTS = (
```
jthomson04, line 51: **These point tuples re-type coordinates that already live in the shared grid.**

Every coordinate here already exists: `(1,1024)`, `(1,32768)`, `(1,8192)` are `grid.PREFILL_POINTS[0]`, `[2]`, `[1]`; `(1,1024)`, `(128,1024)`, `(32,1024)`, `(8,32768)` are `grid.DECODE_POINTS[0]`, `[2]`, `[1]`, `[4]`.

That creates a second source of truth. If `grid.PREFILL_POINTS` later replaces `(1,32768)` with `(1,16384)`, the original 36 cases move but these 28 hand-written ones stay on the old coordinate, so the expanded matrix silently stops mirroring the regression grid it is documented to follow -- and nothing flags it. Derive these from the grid entries instead.

python/aisimulate/tools/forward_perf_gate/cases.py
```diff
@@ -32,6 +32,63 @@
...
+        "backend_name": BACKEND,
+        "backend_version": BACKEND_VERSION,
+        "tp_size": 8,
+        "pp_size": 1,
+        "attention_dp_size": 1,
+        "moe_tp_size": 1,
+        "moe_ep_size": 8,
+        **model,
+    }
+    # model_id also separates report cells and worker cache-preparation groups.
```
jthomson04: **Mangling the profile into `model_id` overloads the field, and this comment is inaccurate.**

`compare._cell_summaries` keys cells on `(model_id, database_mode, phase, metric)` (`compare.py:251`), so the profile gets encoded into the model name, yielding `deepseek-v3.2-b200_sxm-vllm-0.24.0-tp8-pp1-adp1-mtp1-ep8` alongside a bare `qwen3-32b`. The same field now means two different things across the matrix, and the summary table becomes hard to read.

The claim about cache groups is also wrong: `worker.GROUP_KEYS` already contains `system_name`, `backend_name`, `backend_version`, `tp_size`, `pp_size`, `attention_dp_size`, `moe_tp_size` and `moe_ep_size`, so only the `-prefix` suffix on line 122 is actually load-bearing for grouping. The cleaner fix is to extend the cell key in `compare.py` with the profile fields and leave `model_id` as the model name.

scripts/select_forward_perf.py
```diff
@@ -0,0 +1,165 @@
...
+    if not re.fullmatch(r"[0-9a-f]{40}", sha) or sha != pull["head"]["sha"]:
+        raise ValueError(f"Trusted copy {sha} does not match PR #{pr_number} head {pull['head']['sha']}")
+
+    run = True
+    if event == "workflow_dispatch":
+        reason = "Manual dispatch requests a comparison."
+    elif pull["changed_files"] > 3000:
+        reason = "PR exceeds the file API limit; run conservatively."
+    else:
+        pages = api(f"{endpoint}/files?per_page=100")
```
jthomson04, line 108: **This duplicates the trusted-change-set routine already inlined in `ci.yml`.**

`.github/workflows/ci.yml` (`select-full-ci`, L150-166) already does the same four steps: read `.changed_files`, bail to conservative mode above 3000, `gh api --paginate .../files?per_page=100` emitting `filename` plus `previous_filename // empty`, then re-read `.head.sha` and fail if it moved. This is a second implementation of a security-relevant routine.

The two copies now have to stay in sync. If GitHub raises the 3000-file cap, or the rename handling needs a fix, only one gets updated and the two CI selectors disagree about what a PR changed. Worth extracting the resolution into one shared helper and keeping only the per-workflow path rules separate.

jthomson04 (commented):
### Code review

15 findings, ranked roughly by severity. 14 are inline; the one below is on a file this PR does not modify.

**`forward_perf_gate/worker.py:253` -- cache isolation no longer holds across the expanded matrix.**

Group order is now: groups 1-8 `b200_sxm`/vllm/0.24.0, group 9 `h100_sxm`/vllm/0.24.0 (nemotron), groups 10-12 `b200_sxm`/vllm/0.24.0, group 13 `b200_sxm`/sglang/0.5.14 -- three distinct `(system, backend, version)` keys where there was previously one.

`clear_caches(representative)` calls `perf_database.unload_database(system, backend, version)` for the *incoming* group only, so entering group 9 leaves the b200 vLLM database resident, and by group 13 all three are loaded simultaneously in one worker process. The README states "Before each group, the worker releases the prior objects, clears the prediction caches" and "The harness treats `clear_caches()` as the complete cache-isolation contract"; neither holds any more. Groups 9-13 -- including both prefix groups for the original models -- measure `session_setup_ms` and cold samples with foreign databases and a larger resident heap than groups 1-8.

The three worth resolving before merge are that one, the manual-dispatch hard-fail in `select_forward_perf.py:99`, and the `grid.py` gap in the head-matrix validation trigger. The rest are maintainability and robustness notes.

Checked against the PR head: 64 cases, 31/33 phase split, frozen 36-case hash matches, 4 original + 9 new cache groups, all referenced model configs and perf-data directories present.

python/aisimulate/tests/unit/tools/test_forward_perf_gate_worker.py
```diff
@@ -40,13 +42,53 @@ def _request() -> dict:
...
     assert len({case["case_id"] for case in expanded}) == len(expanded)
-    assert sum(case["phase"] == "context" for case in expanded) == expected_context
-    assert sum(case["phase"] == "generation" for case in expanded) == expected_generation
+    assert sum(case["phase"] == "context" for case in expanded) == 31
+    assert sum(case["phase"] == "generation" for case in expanded) == 33
     assert {case["database_mode"] for case in expanded} == {"SILICON", "EMPIRICAL"}
+    assert all(case["database_mode"] == "SILICON" for case in expanded[36:])
+    # Freeze the original 36 requests, including their IDs, values, and order.
+    legacy = json.dumps(expanded[:36], sort_keys=True, separators=(",", ":")).encode()
+    assert hashlib.sha256(legacy).hexdigest() == "17b8247fc587711a5e7b00b9cb3dba86e093b9e210e8272112eb33ddea39eca2"
```
jthomson04, line 53: **The frozen sha256 couples this test to another tool's grid and fails opaquely.**

`grid.PREFILL_POINTS` / `DECODE_POINTS` / `CTX_OSL` / `GEN_OSL` / `STRIDE` belong to `prediction_regression_gate`. A legitimate change there -- adding a decode point, say -- makes this fail with `AssertionError: '<hex>' == '17b8247f...'` and no indication of which field moved, in a test file the author was not touching. The assertions this replaced (`len(cases.MODELS) * len(cases.DATABASE_MODES) * len(grid.PREFILL_POINTS)`) tracked the grid automatically.

If freezing is intentional, compare against an explicit expected list or a small committed JSON fixture so the failure diff shows what actually changed.

python/aisimulate/tests/unit/tools/test_forward_perf_gate_worker.py
```diff
@@ -40,13 +42,53 @@ def _request() -> dict:
...
+        assert case["system_name"] in case["model_id"]
+        assert f"{case['backend_name']}-{case['backend_version']}" in case["model_id"]
+        assert f"-tp{case['tp_size']}-pp{case['pp_size']}-adp{case['attention_dp_size']}" in case["model_id"]
+        assert f"-mtp{case['moe_tp_size']}-ep{case['moe_ep_size']}" in case["model_id"]
+    assert [case["prefix"] for case in expanded if case["prefix"]] == [4096, 7168, 4096, 7168]
+    dp_cases = [case for case in expanded if case["attention_dp_size"] == 8]
+    assert [(case["batch_size"], case["isl"], case["tp_size"], case["moe_ep_size"]) for case in dp_cases] == [
+        (32, 1024, 1, 8),
+        (8, 32768, 1, 8),
+    ]
```
jthomson04, line 73: **The priming-outside-the-matrix guard was not extended to the new point sets.**

`test_phase_priming_query_is_outside_the_measured_matrix` (L339-341) asserts `(2, 2048) not in points` for `grid.PREFILL_POINTS` and `grid.DECODE_POINTS` only. `ADDITIONAL_MODEL_POINTS`, `PREFIX_POINTS`, `ATTENTION_DP_POINTS` and `SGLANG_POINTS` are unchecked, so that invariant is now verified for only the original 36 coordinates.

No live bug today -- none of the new sets contains `(2, 2048)`. But if a future point set adds it, or the protocol-v1 priming coordinates change, the cold sample for that case silently becomes a warm one (the engine already saw the shape during priming) and the cold-vs-warm comparison reports a bogus improvement with no test failure.

tests/test_ci_workflow_contracts.py
```diff
@@ -35,6 +36,196 @@
...
+    api, _ = _forward_api([[{"filename": "docs/ci.md"}]])
+
+    def run(command, **kwargs):
+        assert command[:4] == ["gh", "api", "--paginate", "--slurp"]
+        assert kwargs["check"] is True
+        if fail_api and "/files?" in command[-1]:
+            raise subprocess.CalledProcessError(1, command)
+        return SimpleNamespace(stdout=json.dumps(api(command[-1])))
+
+    monkeypatch.setattr(forward_perf.subprocess, "run", run)
```
jthomson04, line 245: **This patches the shared stdlib `subprocess` module, not a seam on the module under test.**

`forward_perf.subprocess` *is* the global `subprocess` module object, so this replaces `subprocess.run` process-wide for the duration of the test. The stub asserts `command[:4] == ["gh", "api", "--paginate", "--slurp"]`, so anything else reaching `subprocess.run` inside this test body (a pytest plugin, a fixture finalizer) fails with a confusing assertion about `gh api`.

Patch the seam instead -- `monkeypatch.setattr(forward_perf, "github_api", fake)` -- or use the `api=` injection point that `select_comparison` already exposes.

scripts/select_forward_perf.py
```diff
@@ -0,0 +1,165 @@
...
+        KeyError,
+        IndexError,
+        TypeError,
+        ValueError,
+        OSError,
+        subprocess.SubprocessError,
+    ) as error:
+        print(f"Cannot select forward performance CI: {error}", file=sys.stderr)
+        return 1
+    with Path(os.environ["GITHUB_OUTPUT"]).open("a") as output:
```
jthomson04, line 150: **These two env reads sit outside the try/except that guards every other one.**

`select_comparison`'s `os.environ["GITHUB_REPOSITORY"]` and friends are inside the `try` above, with `KeyError` caught and turned into a clean `Cannot select forward performance CI: ...` plus exit 1. `GITHUB_OUTPUT` here and `GITHUB_STEP_SUMMARY` on L154 are not.

Running outside Actions, or in a composite step that strips the env, produces a raw `KeyError: 'GITHUB_OUTPUT'` traceback instead of the intended diagnostic. And if the summary write fails after the output write, the job outputs are already committed while the step reports an unexplained crash.

python/aisimulate/tools/forward_perf_gate/README.md
```diff
@@ -39,10 +45,55 @@ array. Run one request with:
 python tools/forward_perf_gate/worker.py --request request.json --pretty
 ```
 
-The initial matrix uses Qwen3-32B and Qwen3-235B-A22B on B200/vLLM 0.24.0,
-with SILICON and EMPIRICAL database modes and the nine points from the existing
-prediction regression grid. The isolated worker opts into this pinned raw data
-version instead of resolving a moving backend-version alias.
+The matrix has 64 cases and reports 128 cold/warm comparisons. The first 36
```
jthomson04, line 52: **128 comparisons at unchanged thresholds raises the false-red rate on this advisory check.**

Comparisons go from 72 to 128 while the 4-of-5-round / 10% / 2 us rule stays the same and is applied independently to each one. This PR's own control data already shows the *unchanged* 36-case matrix producing a spurious REGRESSION (`qwen3-235b-a22b/silicon/context/bs4-isl4096`, +60.8%) and 37-44 of 128 cold comparisons classified noisy per run.

With ~78% more comparisons per PR, unrelated PRs will see red advisory checks roughly 1.8x as often, which erodes the signal the gate exists to provide. Either widen the cold-query threshold, or report cold regressions at cell level only (the report already computes per-cell geomeans) rather than flagging every individual cold point.

.github/workflows/performance.yml
```diff
@@ -47,7 +25,33 @@ concurrency:
...
+    name: Select Forward Performance
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    outputs:
+      number: ${{ steps.scope.outputs.number }}
+      head_sha: ${{ steps.scope.outputs.head_sha }}
+      base_ref: ${{ steps.scope.outputs.base_ref }}
+      run_comparison: ${{ steps.scope.outputs.run_comparison }}
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
```
jthomson04, line 38: **The selector now runs from a PR-controlled checkout with `GH_TOKEN` in scope.**

For a `push` to `pull-request/N`, `ref: ${{ github.sha }}` is the contributor's code, and the `select` step then executes `scripts/select_forward_perf.py` from that checkout. A PR that edits the selector can emit any `head_sha` / `run_comparison` / `base_ref` it likes, bypassing every guarantee the description advertises (verify the trusted head, recheck head and base, fail selection on API errors); `compare` then checks that SHA out on the self-hosted `prod-aisimulate-default-amd64-v1` runner. Under the old design the resolve step ran `actions/github-script` before any checkout, so PR code never saw `GH_TOKEN`.

This mirrors the existing `select-full-ci` pattern, so it is not a new class of exposure -- but it is a real posture change worth making deliberately. Pinning the selector checkout to the base or default branch would restore the old guarantee.

scripts/select_forward_perf.py
jthomson04: Keeping the current-head requirement, as agreed. Manual dispatch overrides path selection but does not waive revision checks. The documentation now states that an older trusted copy fails selection rather than producing evidence for an older PR revision.

scripts/select_forward_perf.py
jthomson04: Keeping the explicit failure on API errors. The selection job fails with a diagnostic, so this does not produce a successful silent skip. A successful response with an empty, incomplete, or oversized file list remains a separate conservative-run case. We are not adding retry or fallback policy in this PR.

python/aisimulate/tools/forward_perf_gate/cases.py
jthomson04: Keeping the new scenario coordinates explicit. These are fixed, qualified workloads selected for this expansion; only the original 36 cases are defined by the shared grid. Deriving the new scenarios from positional grid entries would make unrelated grid edits change their coverage indirectly.

python/aisimulate/tools/forward_perf_gate/cases.py
jthomson04: Keeping profile labels in model_id, as specified by the coverage plan, so configurations remain separate without changing report keys. The code comment and README now distinguish those report labels from the prefix suffix, which separates otherwise equal worker groups. A report-grouping redesign is deferred.

scripts/select_forward_perf.py
jthomson04: Deferring the shared-helper extraction. It would expand this change into Full CI without fixing a demonstrated defect in this selector. The performance selector retains its focused tests for pagination, renamed paths, incomplete results, and revision changes.

python/aisimulate/tests/unit/tools/test_forward_perf_gate_worker.py
jthomson04: The freeze is intentional: preserving the original 36 requests and their order is a requirement of this expansion, so a shared-grid change should require an explicit update here. Improving the failure diff with an expected fixture is useful maintenance work, but is deferred to keep these review fixes small.

tests/test_ci_workflow_contracts.py
jthomson04: Deferring this test cleanup. This test intentionally exercises main() through the gh subprocess boundary, including command options and CLI error handling; patching select_comparison's API input would bypass that coverage. There is no observed interference in the focused suite. A narrower patch context can be considered separately.

scripts/select_forward_perf.py
jthomson04: Deferring the output-error diagnostic cleanup. GitHub Actions supplies these environment variables in the supported workflow, and a missing or unwritable output already makes the step fail. Broader handling for execution outside that environment is outside this focused fix.

python/aisimulate/tools/forward_perf_gate/README.md
jthomson04: Deferring noise-policy changes. The approved scope keeps the existing thresholds and reports cold/warm noise separately. The comparison count alone does not establish a 1.8x false-alarm rate. Threshold changes or cell-level cold reporting need a separate calibration study; we will not loosen thresholds to qualify the additional cases.

.github/workflows/performance.yml
jthomson04: Deferring the workflow-security redesign. Trusted-copy admission remains the security boundary, and the workflow token has read-only contents/PR permissions. The selector's revision checks provide consistency checks within that admitted run. Pinning only the Python file to main would not secure the PR-defined workflow itself, so that proposal needs a separate design across the CI workflows.

.github/workflows/performance.yml
jthomson04: Fixed in [15744612](https://github.com/ai-dynamo/aisimulate/commit/157446124910dab556883d1d88eda63fea99a670). PR-controller detection now includes prediction_regression_gate/grid.py as well as the gate's Python files. A behavioral test runs the workflow's detection code against a real grid-only Git change and verifies that it enables controller validation.

python/aisimulate/tools/forward_perf_gate/cases.py
jthomson04: Fixed in [15744612](https://github.com/ai-dynamo/aisimulate/commit/157446124910dab556883d1d88eda63fea99a670). DeepSeek V3.2 and Qwen3.5 now have named configurations reused by the model list and their special scenarios. Existing matrix tests assert both scenario model identities. All 64 serialized requests, IDs, values, and ordering are unchanged.

.github/workflows/performance.yml
jthomson04: Fixed in [15744612](https://github.com/ai-dynamo/aisimulate/commit/157446124910dab556883d1d88eda63fea99a670). Selection and controller-change detection now cover the gate's Python files, including measurement.py, plus the shared grid. Gate README-only changes skip benchmarking. Tests retain the behavior that a documentation-only follow-up still runs the benchmark when the full PR contains relevant code changes.

python/aisimulate/tests/unit/tools/test_forward_perf_gate_worker.py
jthomson04: Fixed in [15744612](https://github.com/ai-dynamo/aisimulate/commit/157446124910dab556883d1d88eda63fea99a670). The priming test now checks every expanded case in each phase, while retaining the checks that protocol-v1 priming coordinates stay fixed when the shared-grid OSL constants change.

jthomson04:
Addressed the five agreed fixes in [15744612](https://github.com/ai-dynamo/aisimulate/commit/157446124910dab556883d1d88eda63fea99a670).

For the [database-retention finding](https://github.com/ai-dynamo/aisimulate/pull/236#pullrequestreview-5230642670), the batch worker now evicts the outgoing database through the existing public cache-reset path when `(system, backend, version)` changes. The incoming reset and consecutive same-database behavior remain unchanged. Tests cover normal transitions and transitions after setup or priming failures. The documentation describes this lifecycle. This fixes retention; it does not establish a prior timing effect.

The other fixes cover the shared-grid controller trigger, Python-only gate selection, named scenario configurations, and full-matrix priming exclusion. All 64 requests and their order are unchanged. The agreed deferrals and declines are documented in the inline replies and remain open for reviewer agreement.

Validation: 267 focused tests passed; the 30 forward-performance workflow tests also pass with warnings treated as errors. Ruff, actionlint, and whitespace checks passed. Three matched-revision runs completed all 64 cases and five paired rounds with no missing data, invalid comparisons, or regression flags. Cold unstable counts were 37/9/17; warm counts were 0/0/0. Median added benchmark time was 51.29 seconds, below five minutes.

[Automatic performance CI](https://github.com/ai-dynamo/aisimulate/actions/runs/35181140976) passed on `15744612`. The base controller completed 36 cases/72 comparisons in 54 seconds; the PR controller completed 64 cases/128 comparisons in 136 seconds. Both completed five rounds with no regression, missing-data, skipped, or invalid results. Cold/warm unstable counts were 9/2 and 26/2. Artifact upload and cleanup passed. The PR description contains the full evidence.

Fast CI passed after adding the required review-ready label. Full CI still has the same three nightly-license file-warning failures as the prior head, on both architectures; this is recorded separately in the description.

python/aisimulate/tools/forward_perf_gate/README.md
```diff
@@ -39,10 +49,57 @@ array. Run one request with:
...
+
+- Both original Qwen models on B200/vLLM 0.24.0 retain their original parallel
+  layouts, with context batch 1, ISL 8,192, and cached prefixes 4,096 and 7,168.
+- DeepSeek V3.2 on B200/vLLM 0.24.0 uses attention DP=8, TP=1, MoE TP=1,
+  and EP=8 for generation at batch 32/ISL 1,024 and batch 8/ISL 32,768.
+- Qwen3.5 on B200/SGLang 0.5.14 uses TP=8, attention DP=1, MoE TP=1, and
+  EP=8 for context batch 1/ISL 8,192 and generation batch 32/ISL 1,024.
+
+SGLang is pinned to 0.5.14 because 0.5.16 lacks the B200 custom-all-reduce
+data required when shared-layer reuse is disabled.
```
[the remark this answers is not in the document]
jthomson04: The version choice is based on the checked-in data. At this PR head, the [B200 SGLang communication data](https://github.com/ai-dynamo/aisimulate/tree/157446124910dab556883d1d88eda63fea99a670/python/aisimulate/src/aiconfigurator_core/systems/data/b200_sxm/comm/sglang) includes `0.5.14/custom_allreduce_perf.parquet` and has no `0.5.16` communication directory. The [worker disables shared-layer reuse](https://github.com/ai-dynamo/aisimulate/blob/157446124910dab556883d1d88eda63fea99a670/python/aisimulate/tools/forward_perf_gate/worker.py#L259), so the local data is required. The [current automatic qualification](https://github.com/ai-dynamo/aisimulate/actions/runs/35181140976) completed all 64 cases with 0.5.14 and no data misses.

Leaving this new README source-link request open. This patch addresses the five fixes agreed from the original review.

jasonqinzhou (approved):
**APPROVE · 9.3/10 · high confidence**

## What this PR does

This PR expands the advisory forward-prediction performance matrix from 36 to 64 cases, selects against the complete PR change set before allocating a benchmark runner, validates changed PR controllers, and evicts outgoing performance databases across configuration transitions.

The author preserved the four PR patches exactly through the rebase and supplied focused boundary coverage for selector pagination and renames, controller-change detection, cache transitions after failures, and the original 36-case ordering.

## Why this score

9.3/10 and APPROVE for the exact reviewed head. The rebase preserves all four PR patches unchanged, and strict local checks plus exact-head Full CI and the hosted 36-case/64-case benchmark pass without an actionable defect. No review comment is recommended. The requested Claude Fable model was not observed, so this is a Codex-reviewed result with consensus unavailable, not a dual review; the 11 unresolved scheduler-snapshot threads remain a separate merge-readiness obligation.

python/aisimulate/tools/forward_perf_gate/README.md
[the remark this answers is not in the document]
jthomson04: The source evidence is linked above and the version pin is verified. We are treating the inline README citation as an optional documentation improvement and deferring it from this PR. Resolving this thread on that basis; the selected version and worker behavior remain unchanged.

jthomson04:
Regarding the [publication-guard comment](https://github.com/ai-dynamo/aisimulate/pull/236#pullrequestreview-5239435512): we are leaving this behavior unchanged.

A superseded or canceled run may retain its own diagnostic output. The [report records both measured commit SHAs](https://github.com/ai-dynamo/aisimulate/blob/a9867cf03fd49e617d7b0ca17a9dde12f43af339/python/aisimulate/tools/forward_perf_gate/compare.py#L447), and [GitHub artifacts belong to their workflow run](https://docs.github.com/en/actions/tutorials/store-and-share-data#downloading-artifacts-during-a-workflow-run). These steps write the current run's summary and artifact; they do not update a mutable latest-result record or replace the newer commit's check. Keeping those diagnostics is useful when a run fails or is canceled.

We do not consider an additional current-head publication check necessary for this PR. The existing revision checks still govern selection, and the approved head a9867cf0 has passing Fast CI, Full CI, and both benchmark controllers. The earlier decline/deferral reasons stand; the optional README citation is also deferred. No code change is being made while resolving the remaining conversations.

PR #239, opened 2026-09-16
perf(perfmodel): index PreparedGrid lookups

jthomson04:
## Summary

Repeated boundary-hold queries scan every leaf of a prepared measured-data grid. This PR uses one lazy immutable k-d tree over the existing log2 coordinates, with a bounded candidate heap and one final sort. It continues the implementation from #119.

This review pass starts from `e7ad28292d32e2982eb7d36b5246c9f606bfdc73` without a rebase. Tested candidate: `71c8e6579bbd65a18676ae554ca1068c5e865557`. It adds two separate commits: a flat scan for at most 16 leaves (also when the query's candidate limit covers all leaves), and `BinaryHeap` in place of the manual heap. Tree construction stays independent of the first query's SOL function and rank. `UtilGrid` construction is unchanged.

The distance/index ordering, query-specific SOL evaluation, validity filtering, support radius, weighting and exhaustive fallback are preserved. There are no public API, scheduler, cache, table-loading or model-data changes. The shared-tree contract and the pre-existing `NearestTwo` square-root rounding limitation are now explicit; that behavior change and wider refactors remain deferred.

## Validation

- Core release Rust library suite: **1,339 passed, 0 failed, 1 ignored**, including shared-tree and `UtilGrid` differential tests.
- **13 Dynamo prediction CLI smoke tests**, **8 Qwen3.8 engine-step parity tests**, and **6 existing Qwen compile-engine parity tests** passed. The compile suite has a Qwen3-30B case, not a Qwen3.8 case. Goldens are unchanged.
- New cutoff coverage checks 12, 13, 16 and 17 leaves and whole-table candidate limits, with bit-identical anchors and latency/power/energy. Existing indexed/equal-distance fixtures still build a tree. The collector test covers signed zero and non-finite values. NaN-clamp and infinity-fallback coverage are explicit.
- Formatting and diff checks passed. Clippy completed with existing warnings; all reported source lines also exist at the prior PR head. The retained strict `-D warnings` run fails on those existing warnings.
- All three variants have exact full prediction parity after excluding only `summary.wall_time_ms`, `summary.processed_tokens_per_s` and `summary.processed_output_tokens_per_s`. Per-request files are byte-identical, including timing. Compiled engine specifications match except `engine.systems_path`.
- All 339 recorded model/system/data file hashes, all four input/configuration hashes, 158 package versions and normalized Dynamo compatibility patches match. Timed/profiled simulated summaries match qualification.

### Forward prediction gate: accepted setup cost

The [Forward Prediction Performance (advisory) check](https://github.com/ai-dynamo/aisimulate/actions/runs/35182848679/job/105078571376) **fails at this head** on the cold 32K-token prefill queries for Qwen3-32B and Qwen3-235B-A22B (B200, vLLM 0.24.0, SILICON). Both fail in 5/5 rounds, adding about 0.14 ms and 0.19 ms per first query, respectively.

This is an **accepted setup-cost tradeoff and is not a blocker for this PR**. The gate's 2K-token priming query does not enter the nearest-anchor path, so the cold 32K query still includes lazy k-d tree construction. That cost occurs once per affected grid's first boundary-hold query. Warm queries reuse the tree: these two cases improve by 17.3% and 20.7%, and no warm case regresses under the gate's criteria. All 180 CI prediction pairs match exactly. The Dynamo results below include first-use work and retain the 19.46% AgentX CPU improvement with no material short-run regression. No gate settings or goldens were changed.

## Fresh three-way Dynamo benchmark

Baseline AISimulate: `8da29bf2b2ce91fbb12c9a2cbe3aa68a67494099`. Current PR: `e7ad28292d32e2982eb7d36b5246c9f606bfdc73`. Updated candidate: `71c8e6579bbd65a18676ae554ca1068c5e865557`. All use Dynamo `99224822bdbbd45a520d2ec1a4c9e38ca967e8e7`, matching AISimulate Rust/Python code, separate build directories and environments, and the same measured data.

Intel Core Ultra 9 285K, serial performance CPU4, existing powersave governor. Rust 1.96.1, Python 3.12.12, release/thin-LTO, debug=2, strip=false, frame pointers and x86-64-v3. Builds completed before measurement. Odd rounds run baseline/current/candidate; even rounds reverse the order. Process CPU is user + system. Values are median [min, max].

**AgentX:** Saved 128 sessions, Qwen/Qwen3.8-2.4T-A95B-FP8, GB300, SGLang 0.5.17, TP16, MoE TP4/EP4; 33,101 requests and 28,854,680 output tokens. 10 runs per variant. All requests and output tokens complete.

| Metric | Baseline | Current PR | Updated | vs baseline | vs current |
|---|---:|---:|---:|---:|---:|
| Wall, s | 70.125 [69.290, 71.960] | 56.055 [54.900, 58.330] | 56.475 [54.850, 58.500] | -19.47% | +0.75% |
| Process CPU, s | 70.065 [69.210, 71.930] | 56.005 [54.840, 58.310] | 56.430 [54.780, 58.470] | -19.46% | +0.76% |
| CPU us/output token | 2.4282 [2.3986, 2.4928] | 1.9409 [1.9006, 2.0208] | 1.9557 [1.8985, 2.0264] | -19.46% | +0.76% |
| Internal replay wall, s | 66.8377 [66.0157, 68.6254] | 52.7256 [51.5484, 54.8761] | 53.1882 [51.5666, 55.1190] | -20.42% | +0.88% |
| Max RSS, KiB | 1,110,714 [1,106,184, 1,116,412] | 1,112,664 [1,104,556, 1,117,472] | 1,109,772 [1,106,300, 1,114,624] | -0.08% | -0.26% |

**Short vLLM:** Same model/hardware, vLLM 0.27.1, TP16, MoE TP16/EP1; 1,024 input/output tokens, concurrency 32, 512 requests and 524,288 output tokens. 10 runs per variant. All requests and output tokens complete.

| Metric | Baseline | Current PR | Updated | vs baseline | vs current |
|---|---:|---:|---:|---:|---:|
| Wall, s | 1.465 [1.450, 1.500] | 1.470 [1.450, 1.500] | 1.460 [1.450, 1.480] | -0.34% | -0.68% |
| Process CPU, s | 1.460 [1.450, 1.500] | 1.460 [1.440, 1.490] | 1.460 [1.440, 1.460] | +0.00% | +0.00% |
| CPU us/output token | 2.7847 [2.7657, 2.8610] | 2.7847 [2.7466, 2.8419] | 2.7847 [2.7466, 2.7847] | +0.00% | +0.00% |
| Internal replay wall, s | 0.7292 [0.7245, 0.7441] | 0.7284 [0.7156, 0.7438] | 0.7262 [0.7183, 0.7298] | -0.41% | -0.30% |
| Max RSS, KiB | 228,968 [227,808, 229,312] | 228,492 [228,388, 228,680] | 230,928 [230,128, 231,276] | +0.86% | +1.07% |

Five rounds passed the median gates, but the observed ranges triggered five more rounds. All ten are reported. The predeclared paired bootstrap uses 20,000 resamples (seed 239) and a 95% interval for the ratio of medians. For AgentX CPU versus current PR, that interval is [-0.77%, +2.32%], below the +3% limit at its upper endpoint. All other wall/CPU limits and the -5% baseline improvement limit also clear their uncertainty checks.

The review changes cost 0.76% median AgentX CPU versus the current PR; CPU is higher in 9/10 matched rounds. This is an accepted cost for the simpler standard heap and small-grid improvement, within the agreed 3% bound.

AgentX CPU improves in **10/10 baseline comparisons**. Median acceptance gates: **PASS**. Final acceptance, including uncertainty: **PASS**. Internal replay wall includes setup/first-use work; simulated TTFT and token latency are correctness outputs.

## Lookup benchmark

An identical ignored Rust benchmark uses varying queries and `black_box`, with five cold/warm samples for each of 60 regular/ragged 2D–4D fixtures. Capacities are 12, 13, 16, 17, 19, 20, 64, 256, 1,024 and 4,096. Prime capacities have singleton axes; ragged capacities 16/19/20 exercise 13/16/17 leaves. Cold timing includes first-hold index construction and excludes fixture creation/cloning. Warm samples run for at least 100 ms.

| Selected fixture | Current cold, us | Updated cold, us | Current warm, ns | Updated warm, ns | Warm change |
|---|---:|---:|---:|---:|---:|
| 2D / 12 leaves / regular | 2.35 [1.63, 23.70] | 2.62 [1.22, 21.27] | 254.03 [252.75, 260.93] | 225.81 [223.28, 225.94] | -11.11% |
| 2D / 13 leaves / regular | 1.59 [1.32, 14.01] | 1.08 [0.99, 2.85] | 269.00 [268.82, 269.53] | 243.40 [243.08, 246.33] | -9.52% |
| 2D / 13 leaves / ragged | 1.44 [1.10, 2.11] | 1.29 [0.94, 1.59] | 261.15 [260.91, 263.39] | 224.42 [224.01, 225.15] | -14.06% |
| 2D / 16 leaves / regular | 1.67 [1.56, 2.55] | 1.49 [1.08, 1.72] | 250.99 [250.22, 251.46] | 243.04 [242.57, 243.90] | -3.17% |
| 2D / 17 leaves / regular | 2.29 [1.57, 7.88] | 2.56 [1.74, 13.12] | 276.65 [276.14, 277.10] | 285.78 [285.38, 286.02] | +3.30% |
| 3D / 4096 leaves / regular | 395.29 [393.44, 451.86] | 409.37 [391.80, 455.27] | 1412.73 [1404.20, 1442.04] | 1444.23 [1425.33, 1456.25] | +2.23% |

All fixtures, including baseline costs and ranges, are retained in `lookup-all.md` and `lookup-all.json`. The largest measured warm regression versus the current PR is 4.1% (3D ragged, 17 leaves); the 13-leaf 2D case improves 14.1%. The cutoff and standard heap are retained under the end-to-end acceptance limits, with this small-lookup trade-off recorded.

## Separate AgentX profiles

99 Hz user-cycle profiles use DWARF call stacks. The complete nearest-anchor path counts each sampled stack once, including tree traversal, candidate/heap work, sorting and allocation. Inclusive rows overlap and are not added. Absolute weights are sampled cycle periods, not exact hardware event totals. Heap/sort and tree groups are restricted to descendants of the hold selector; each group counts a stack once. These single-profile estimates do not replace the repeated unprofiled timing results.

| Attribution | Current self / inclusive | Updated self / inclusive | Current periods, self / inclusive | Updated periods, self / inclusive |
|---|---:|---:|---:|---:|
| Logical nearest-anchor selector | 0.34% / 10.36% | 0.44% / 8.87% | 1,086,996,795 / 33,207,870,598 | 1,377,779,612 / 27,826,488,887 |
| Heap collection and final sort, inline | 1.18% / 5.91% | 0.84% / 3.55% | 3,777,130,855 / 18,952,124,178 | 2,639,612,140 / 11,132,716,638 |
| Tree traversal, inline | 1.58% / 8.26% | 2.36% / 7.35% | 5,077,407,182 / 26,478,716,449 | 7,396,068,023 / 23,061,054,502 |
| Native PreparedGrid query | 1.68% / 16.02% | 1.28% / 14.49% | 5,378,815,068 / 51,342,119,321 | 4,017,438,277 / 45,433,228,743 |
| Native tree visit | 3.04% / 8.26% | 4.06% / 7.54% | 9,753,865,275 / 26,478,716,449 | 12,732,276,730 / 23,635,046,827 |
| Native candidate collector | 5.25% / 5.25% | 1.68% / 1.68% | 16,839,289,568 / 16,839,289,568 | 5,277,262,924 / 5,277,262,924 |

The complete nearest-anchor union is **33,207,870,598 → 27,826,488,887 sampled periods** (-16.21%), or **10.36% → 8.87%** of each profile. Total sampled periods are 320,461,693,764 and 313,580,536,779. Both captures have zero lost samples and matching loaded Dynamo build IDs; `qualification/profile-integrity.json` records those checks. Native and inline flamegraphs are retained for both variants.

## Scope decisions

The let-chain, cutoff, standard heap, NaN/infinity clarification and immutable-point-storage contract are addressed. `NearestTwo` ordering changes, `DeltaKdTree` unification, packed nodes, removal of `root`/`samples`, owner-specific wrappers, and shared exhaustive-selector helpers remain deferred. The unchanged exhaustive selector stays independent as the differential reference.

## Source locations

- [FlatHoldIndex cutoff](https://github.com/ai-dynamo/aisimulate/blob/71c8e6579bbd65a18676ae554ca1068c5e865557/crates/core/src/perfmodel/perf_database/perf_interp.rs#L100).
- [Candidate total ordering](https://github.com/ai-dynamo/aisimulate/blob/71c8e6579bbd65a18676ae554ca1068c5e865557/crates/core/src/perfmodel/perf_database/perf_interp.rs#L725).
- [Bounded BinaryHeap collector](https://github.com/ai-dynamo/aisimulate/blob/71c8e6579bbd65a18676ae554ca1068c5e865557/crates/core/src/perfmodel/perf_database/perf_interp.rs#L767).
- [Nearest-anchor selection](https://github.com/ai-dynamo/aisimulate/blob/71c8e6579bbd65a18676ae554ca1068c5e865557/crates/core/src/perfmodel/perf_database/perf_interp.rs#L801).
- [Shared search contract](https://github.com/ai-dynamo/aisimulate/blob/71c8e6579bbd65a18676ae554ca1068c5e865557/crates/core/src/perfmodel/kd_tree.rs#L132).
- [Retained NearestTwo rounding limitation](https://github.com/ai-dynamo/aisimulate/blob/71c8e6579bbd65a18676ae554ca1068c5e865557/crates/core/src/perfmodel/operators/util_empirical.rs#L282).

## Reproduction and artifacts

All scripts, configurations, traces, raw results, profiles and four flamegraphs remain local under:

```text
/home/jothomson/Desktop/dynamo/.worktrees/aisimulate-pr239-review-20260916/benchmark-harness/pr239-review-20260916
```

The measured prior worktree/artifacts are preserved. The benchmark overlay and large artifacts are outside the PR patch. `README.md` has setup, build and exact validation commands; `variants.json` gives all three interpreters. Each run also records its full public CLI command. Existing result tags cannot be overwritten.

```bash
cd /home/jothomson/Desktop/dynamo/.worktrees/aisimulate-pr239-review-20260916/benchmark-harness/pr239-review-20260916
bash measure.sh 11 15  # use fresh tags after the retained rounds
python3 check.py
```

The runner uses `taskset -c 4 /usr/bin/time -v` around `"$VARIANT_PYTHON" -X perf -m aisimulate predict --stack dynamo --config "$CONFIG" --output-dir "$NEW_OUTPUT_DIR" --format json`. Configs are `configs/qwen-agentx-sglang.yaml` and `configs/qwen-synthetic-vllm.yaml`. Both use one replica, 8,192 batched tokens/256 sequences, prefix caching and SILICON timing. Environment: `PYTHONHASHSEED=0`, `DYN_LOG=warn`, `HF_HUB_OFFLINE=1`, OMP/OpenBLAS/MKL threads=1. Profiling wraps that command with `perf record -e cpu_core/cycles/u -F 99 --call-graph dwarf,16384`.

Trace SHA-256: `71f74873d3b1a1ff834bc0cc76b9520908ff19688be8ecd31cc0bb3f4eb5c76f`. `build/*-provenance.json` records data/input hashes, extension SHA-256 hashes and ELF build IDs. `measurements.csv`, `summary.json`, `profile-summary.json`, and `qualification/` contain all results and checks.


jthomson04:
@coderabbitai full review

jthomson04 (commented):
### Code review

Reviewed the 3 commits (726/152 lines across 4 files) at `e7ad282`. No correctness regression found in the new selection path: I ran two throwaway differential fuzzers - `KdTree` top-k vs. a linear scan over 400 randomized 1-4D point sets with heavy duplicates and exact ties, and `hold_anchor_weights_prepared` with and without the index across 120 randomized ragged 2-4D tables with invalid/NaN/zero latencies at widths 1-13. Both were bit-identical. `cargo clippy --all-targets` is warning-free on the changed files and all 681 `perfmodel` tests pass.

The findings below are performance, reuse and test-coverage instead. The main one is the first: with `nn_leaves` hardcoded to 4, the bounded collector cannot prune until it holds 12 candidates, so any grid with <= 12 leaves builds a tree and then traverses every node anyway - strictly slower than the scan it replaced, which is what the regressions in your own benchmark table are.

One note on anchoring: the `DeltaKdTree` comment is attached to the nearest line in the diff; the code it refers to is at `util_empirical.rs:554-662`.

crates/core/src/perfmodel/perf_database/perf_interp.rs
```diff
@@ -136,12 +138,15 @@ impl FlatHoldIndex {
...
-        Self {
+        let mut index = Self {
             coords: coords.into_boxed_slice(),
             log_coords: log_coords.into_boxed_slice(),
             offsets: offsets.into_boxed_slice(),
             leaves: leaves.into_boxed_slice(),
-        }
+            kd_tree: None,
+        };
+        index.kd_tree = KdTree::build(&index);
```
jthomson04: **The index is built and traversed unconditionally, but it cannot prune at `nn_leaves = 4`.**

`nn_leaves` is 4 at both production constructions (`perf_interp.rs:360`, `perf_interp.rs:379`), so `m = nn_leaves + SLACK = 12`. `HoldCandidates::cutoff_distance_squared()` returns `None` until `values.len() == 12`, and `KdTree::search` only prunes when the cutoff is `Some`. A `FlatHoldIndex` with, say, 9 leaves therefore pays `FlatHoldIndex::build`'s `select_nth_unstable_by` recursion plus a 48-byte-per-leaf `Vec<KdNode>`, and then still visits all 9 nodes - now through recursion, node indirection and heap sift work the old flat scan did not have. That is strictly slower than what it replaced.

The PR body's own benchmark records exactly this: a ragged 13-leaf 2D case 9.4% slower warm, and first-hold cost up 2.8-3.4x.

Suggest gating the build on a leaf count that can actually exceed the collector limit (e.g. `index.len() > 32`) so small grids keep the scan.

crates/core/src/perfmodel/operators/util_empirical.rs
```diff
@@ -278,127 +279,17 @@ impl NearestTwo {
...
-            nodes[node].right = right;
-            Some(node)
-        }
-
-        let mut samples: Vec<usize> = (0..norm.len()).collect();
-        let mut nodes = Vec::with_capacity(norm.len());
-        let root = build_nodes(norm, &mut samples, 0, dims, &mut nodes)
-            .expect("non-empty util grid has a kd-tree root");
-        Some(Self { nodes, root, dims })
+impl NeighborCollector for NearestTwo {
```
jthomson04, line 282: **The PR extracts the shared tree but leaves a third copy, `DeltaKdTree`, undeleted in this same file** (`util_empirical.rs:554-662`).

It duplicates the new `kd_tree.rs` nearly line for line: same `len < 3` + finite + equal-dims guard, same `build_nodes` with `select_nth_unstable_by(mid, total_cmp.then(index))`, same `is_sign_negative` near/far split, same `delta * delta <= cutoff` far-branch rule, down to the same comment about equally distant samples with earlier indices.

Given the stated goal of sharing the private exact search code, stopping at `UtilKdTree` leaves two hand-synchronised implementations of a subtle exactness contract. A nearest-1 `NeighborCollector` over the new `KdTree` - mirroring what `NearestTwo` does right here - would delete ~100 lines and leave one implementation to fuzz.

crates/core/src/perfmodel/operators/util_empirical.rs
```diff
@@ -278,127 +279,17 @@ impl NearestTwo {
...
-                    visit(tree, norm, query, far, nearest);
-                }
-            }
-        }
-
-        debug_assert_eq!(query.len(), self.dims);
-        let mut nearest = NearestTwo::default();
-        visit(self, norm, query, self.root, &mut nearest);
-        nearest
+    fn cutoff_distance_squared(&self) -> Option<f64> {
```
jthomson04, line 291: **The pruning cutoff and the ordering key disagree for `NearestTwo`, so a far-branch tie-break can be pruned.**

`Neighbor::cmp` (`util_empirical.rs:254`) orders by `distance` (= `distance_squared.sqrt()`) with a sample-index tie-break, while this cutoff returns `second.distance_squared` and `KdTree::search` prunes on `delta * delta > cutoff`. `sqrt` is exactly rounded, so two distinct squared values can map to the same `distance`. If a far-subtree point P has `distance_squared_P > second.distance_squared` but `sqrt(distance_squared_P) == second.distance` and `sample_P < second.sample`, P should win the documented "earlier original index" tie-break, yet the branch is pruned whenever `second.distance_squared < delta*delta <= distance_squared_P`.

This is pre-existing behaviour, but `kd_tree.rs:42` now documents the shared tree as an exact nearest-neighbour index for both callers. `HoldCandidates` does not have the gap (it orders and prunes on `distance_sq`). Either order `NearestTwo` on `distance_squared` too, or note the caveat in the shared doc.

crates/core/src/perfmodel/perf_database/perf_interp.rs
```diff
@@ -696,20 +711,108 @@ struct HoldAnchor {
...
+    }
+
+    /// Consume the max-heap and return the exact nearest-first order expected
+    /// by validity filtering and support-radius selection.
+    fn into_sorted_values(mut self) -> Vec<HoldCandidate> {
+        self.values.sort_unstable_by(|left, right| left.cmp(*right));
+        self.values
+    }
+
+    fn sift_up(&mut self, mut child: usize) {
```
jthomson04: **This is `std::collections::BinaryHeap`, reimplemented.**

Lines 748-778 are ~45 lines of hand-written heap index arithmetic (`(child - 1) / 2`, `parent * 2 + 1`, worse-child selection) - a classic off-by-one surface that now needs its own unit test (`hold_candidate_heap_matches_exact_sorted_top_k` exists only to cover it).

The crate already runs `BinaryHeap` over a `total_cmp`-based `Ord` in `crates/core/src/replay/event.rs:156` and `crates/core/src/replay/disagg.rs:929`. Giving `HoldCandidate` an `Ord` that delegates to the existing `HoldCandidate::cmp` makes `peek()` the cutoff, `pop()` + `push()` the bounded replace, and `into_sorted_vec()` a drop-in for `into_sorted_values()` - same ordering, no hand-written index math.

crates/core/src/perfmodel/kd_tree.rs
```diff
@@ -0,0 +1,316 @@
...
+///
+/// Implementations choose their exact ordering rule. The squared cutoff is
+/// used only for exact far-branch pruning.
+pub(crate) trait NeighborCollector {
+    fn consider(&mut self, sample: usize, distance_squared: f64);
+    fn cutoff_distance_squared(&self) -> Option<f64>;
+}
+
+#[derive(Debug, Clone)]
+struct KdNode {
```
jthomson04, line 37: **`KdNode` is 48 bytes where 16 would do.**

`sample: usize` (8) + `axis: usize` (8) + `left`/`right`: `Option<usize>` (16 each, since `usize` has no niche) = 48 bytes. A packed `{ sample: u32, axis: u8, left: u32, right: u32 }` with a `u32::MAX` sentinel is 16.

Sample counts are already bounded by leaf counts that fit `u32` (`offsets` indexes a `Box<[u32]>` of coords) and `axis < dims <= 4`. This matters because the index is now resident for the process lifetime: every `PreparedGrid` that ever takes a boundary hold retains 48 bytes per leaf in the long-lived per-quant/per-key maps built in `gemm.rs`, `mla.rs`, `attention.rs`, `dsv4.rs` and `wideep_mla.rs`, plus the same per `UtilGrid` sample.

crates/core/src/perfmodel/kd_tree.rs
```diff
@@ -0,0 +1,316 @@
...
+    axis: usize,
+    left: Option<usize>,
+    right: Option<usize>,
+}
+
+/// Immutable exact nearest-neighbour index over finite, equal-sized points.
+#[derive(Debug, Clone)]
+pub(crate) struct KdTree {
+    nodes: Vec<KdNode>,
+    root: usize,
```
jthomson04, line 49: **`root` is always 0, and `samples` exists only to feed a `debug_assert`.**

`build_nodes` pushes the root node first into an empty `nodes` vec (`let node = nodes.len()` with `nodes.len() == 0`), so the value returned at line 109 and stored here is 0 on every build; line 171 always calls `visit(..., 0, ...)`. The `DeltaKdTree` in `util_empirical.rs` already hardcodes `visit(self, query, 0, &mut nearest)` and omits the field.

Dropping both saves two `usize`s per tree and spares a reader having to prove the invariant.

crates/core/src/perfmodel/kd_tree.rs
```diff
@@ -0,0 +1,316 @@
...
+
+    /// Return whether this tree can search the given query without changing
+    /// the caller's former fallback behavior.
+    pub(crate) fn can_query(&self, query: &[f64]) -> bool {
+        query.len() == self.dims && query.iter().all(|value| value.is_finite())
+    }
+
+    /// Search the tree and pass exact candidates to the caller's bounded
+    /// collector.
+    pub(crate) fn search<P, C>(&self, points: &P, query: &[f64], nearest: &mut C)
```
jthomson04, line 132: **The point set is a caller-supplied parameter guarded only by a `debug_assert`.**

`search` indexes `points.point(node.sample)` with indices baked in at build time; the only guard is `debug_assert_eq!(points.len(), self.samples)` on line 132, which is compiled out in release. Both current call sites pass data the tree is already glued to - `hold_anchor_weights_prepared` does `tree.search(index, ...)` where `tree` *is* `index.kd_tree`, and `UtilGrid` passes `self.norm` - so the parameter exists purely to dodge a self-referential struct.

A future caller that passes a shorter slice gets a release-mode out-of-bounds panic; one that passes a same-length but reordered/filtered set gets silently wrong neighbours. Exposing the search as a method on the owner (`FlatHoldIndex::nearest(&self, q_log, collector)`) would make the mismatch unrepresentable.

crates/core/src/perfmodel/perf_database/perf_interp.rs
```diff
@@ -2266,4 +2479,154 @@ mod tests {
...
+                            let mut query: Vec<f64> = (0..axes.len())
+                                .map(|axis| ((ordinal * 17 + axis * 11) % 31) as f64 / 2.0)
+                                .collect();
+                            // Force a boundary hold, including beyond ragged rows.
+                            *query.last_mut().unwrap() = 32.0 + ordinal as f64 / 7.0;
+                            if ordinal == 18 {
+                                query[0] = f64::INFINITY;
+                            }
+                            if ordinal == 19 {
+                                query[0] = f64::NAN;
```
jthomson04, line 2620: **This NaN case does not reach the non-finite fallback it is written to cover.**

`hold_anchor_weights_prepared` maps coords through `v.max(1e-12).log2()`, and Rust's `f64::max` returns the non-NaN argument: `f64::NAN.max(1e-12).log2() == -39.863...`, which is finite. So `q_log` is finite, `can_query` returns true, and the indexed path runs - this ordinal is a duplicate of the ordinary cases.

Only ordinal 18 (`INFINITY` -> `inf`) actually reaches the `can_query == false` linear path, so the coverage claim for non-finite queries is half met. Either drop this row or inject the non-finite value into `q_log` directly.

crates/core/src/perfmodel/perf_database/perf_interp.rs
```diff
@@ -727,27 +830,18 @@ fn hold_anchor_weights_prepared(
...
-            distance_sq += delta * delta;
-        }
-        if best.len() == m {
-            if distance_sq >= best[m - 1].distance_sq {
-                continue;
+    let mut best = HoldCandidates::new(m);
+    if let Some(tree) = index.kd_tree.as_ref().filter(|tree| tree.can_query(&q_log)) {
+        tree.search(index, &q_log, &mut best);
+    } else {
+        for leaf_idx in 0..index.len() {
```
jthomson04, line 826: **Third copy of the per-leaf squared-distance computation in this file.**

`kd_tree.rs:142-146` (tree visit), this fallback loop, and `hold_anchor_weights_exhaustive` (`perf_interp.rs:894-905`) all compute the same sum of `(log_coord - q_log[axis])^2` over a `FlatHoldIndex`, and the two copies in this file additionally share ~30 lines of identical validity-filter / support-radius / picked-anchor tail.

Any future change to the distance metric or the validity predicate has to land in three places and stay bit-identical, or the indexed, fallback and exhaustive paths silently diverge - which is the exact parity property this PR spends its test budget defending. One `fn leaf_distance_sq(index, leaf_idx, q_log)` plus one anchor-selection helper over a distance-ordered candidate iterator would collapse it.

crates/core/src/perfmodel/kd_tree.rs
```diff
@@ -0,0 +1,316 @@
...
+            // distant sample with an earlier original index. Keep this test in
+            // squared space so sqrt rounding cannot prune a valid candidate.
+            if nearest
+                .cutoff_distance_squared()
+                .is_none_or(|cutoff| delta * delta <= cutoff)
+            {
+                if let Some(far) = far {
+                    visit(tree, points, query, far, nearest);
+                }
+            }
```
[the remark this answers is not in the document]
jthomson04: Applied the let-chain cleanup in 71c8e657. The cutoff predicate still runs before the far-node match. The core tests, formatting check and Clippy run are complete; Clippy still reports existing warnings elsewhere.

crates/core/src/perfmodel/perf_database/perf_interp.rs
jthomson04: Applied a 16-leaf cutoff to FlatHoldIndex in eff4c62c. This covers the measured 13-leaf case without selecting an unmeasured 32-leaf threshold. Queries also use the flat scan when their candidate limit covers every leaf. Tree construction remains independent of the first query's SOL function or rank; UtilGrid construction is unchanged. Tests cover 12, 13, 16 and 17 leaves, whole-table limits, and exact anchors/latency/power/energy. Existing indexed and equal-distance fixtures remain above the cutoff. Fresh three-way benchmark results are in the updated PR description.

crates/core/src/perfmodel/operators/util_empirical.rs
jthomson04: Deferred for scope. DeltaKdTree is a separate lookup path and was not part of this optimization's measured cost. Sharing it would add another caller and correctness surface to validate. This PR keeps that path unchanged; a separate cleanup can evaluate it with its own differential tests.

crates/core/src/perfmodel/operators/util_empirical.rs
jthomson04: Confirmed the rounding limitation also exists in the baseline. In 71c8e657, the shared-tree documentation now states that exactness requires a cutoff consistent with the collector ordering, and NearestTwo documents its retained rounded-sqrt ordering with a squared-distance cutoff. HoldCandidates uses squared distance consistently. The NearestTwo behavior change is explicitly deferred because changing its tie ordering can change existing predictions. Leaving this thread open for that separate decision.

crates/core/src/perfmodel/perf_database/perf_interp.rs
jthomson04: Replaced the manual heap with BinaryHeap in dd5f1f40. HoldCandidate's Eq, PartialEq, Ord and PartialOrd all use the same total ordering: squared distance, then original leaf index. The bounded collector replaces the worst entry through peek_mut() and finishes with into_sorted_vec(). The existing collector test remains and now covers signed zero, infinities and NaN. Fresh correctness checks and benchmark results are in the updated PR description.

crates/core/src/perfmodel/kd_tree.rs
jthomson04: Deferred. No material memory problem was established by the measurements. The u32 coordinate values do not bound the number of samples or usize offsets, and UtilGrid supports more than four dimensions (including an existing 5D test). Packing nodes would need additional bounds, fallback behavior and validation. That is outside this measured search optimization.

crates/core/src/perfmodel/kd_tree.rs
jthomson04: Keeping these fields in this PR. The saving is two usize values per tree, and the sample-count assertion remains a useful check on the search invariant. This small storage cleanup does not address a measured cost and is not needed for the current optimization.

crates/core/src/perfmodel/kd_tree.rs
jthomson04: Documented the invariant in 71c8e657: search must receive the same immutable points, in the same order, used to build the tree. Both private callers satisfy it: FlatHoldIndex supplies its own stored coordinates, and UtilGrid supplies its unchanged normalized samples. The sample-count assertion remains. Owner-specific wrappers are deferred to keep this change small; no public API is added.

crates/core/src/perfmodel/perf_database/perf_interp.rs
jthomson04: Clarified the test in 71c8e657. The NaN row deliberately checks the existing max() clamp behavior; the infinity row exercises the non-finite linear fallback. The test now asserts can_query() for that distinction. Direct NaN rejection remains covered by the tree-level test. Both rows retain exact differential checks against the original exhaustive selector.

crates/core/src/perfmodel/perf_database/perf_interp.rs
jthomson04: Deferred to preserve an independent differential reference. The exhaustive selector is intentionally unchanged and shares neither its distance calculation nor its anchor-selection tail with the new search path. Factoring those operations together would make an error in the shared helper affect both sides of the exactness tests. This PR changes neither the distance metric nor the validity policy.

jasonqinzhou (approved):
**APPROVE · 9.7/10 · high confidence**

## What this PR does

This PR accelerates repeated multi-axis performance-model boundary holds with a shared crate-private k-d tree and deterministic bounded candidate heap while preserving linear and exhaustive fallbacks; the incremental commits keep small tables on a flat scan, replace the manual heap with BinaryHeap, and clarify the shared-search contracts.

The follow-up responds thoughtfully to review feedback: it measures and avoids small-table overhead, removes bespoke heap mechanics, documents the retained UtilGrid rounding limitation honestly, and adds cutoff-boundary and total-order regression coverage without expanding the public API.

## Why this score

Score 9.7/10, APPROVE. The current-head delta resolves the performance and maintainability concerns it targets without changing prediction contracts, and exact-head local tests plus CodeRabbit, Fast CI, and Full CI are green. No actionable P0-P3 issue was found. Claude Fable was requested at this head, but only fallback models were observed, so this remains a Codex-only recommendation rather than a dual-reviewed result.

PR #255, opened 2026-09-17
ci: unbreak the 3.11 license gate on setuptools PEP 639 metadata

pvijayakrish:
A fresh 3.11 venv bundles setuptools 79, whose license is only a PEP 639 expression that pip-licenses 5.5.5 reads as UNKNOWN, failing the gate. Install setuptools>=84 in the check venv, which reports MIT again. Verified: the gate fails on 3.11 without this and passes with it.
- Closes: OPS-8621
- Related PRs or issues:


jasonqinzhou:
Confirmed the [existing review finding](https://github.com/ai-dynamo/aisimulate/pull/255#discussion_r4040467806) on head `5d7af156681de410679e288153f3f717271a6617`: the install command now ends with `setuptools>=84`, but `tests/test_ci_workflow_contracts.py:2188` still asserts that its final argument is `pip-licenses==5.5.5`.

I independently reproduced all four failing cases of `test_python_license_gate_checks_target_environment_before_export`. The same failure is present in [Fast CI's Repository Policy job](https://github.com/ai-dynamo/aisimulate/actions/runs/35261917570/job/105339560544), which blocks Full CI.

Please update the assertion to cover both requirements while preserving the install command:

```python
assert command[-2:] == ["pip-licenses==5.5.5", "setuptools>=84"]
```

Before approval, please address the existing review thread and get Fast CI and Full CI passing on the updated head.

jasonqinzhou (approved):

PR #276, opened 2026-09-18
feat(afd): heterogeneous per-pool hardware and super-node fabric tiering

LPhgh:
Derived from https://github.com/ai-dynamo/aiconfigurator/pull/1597 , with some adaptations, bug fixes, and refactoring.

## Why and what changed

AFD deploys up to three pools — a static prefill pool (when combined with P/D),
the A (attention) pool, and the F (FFN/MoE) pool. Two gaps blocked realistic
rack-scale AFD modeling:

1. **One system for the whole deployment.** The A pool is memory-bandwidth bound
   and owns the KV cache, while the F pool is compute bound — they want opposite
   devices, but every pool was forced onto a single `system_name`.
2. **The cross-pool transfer ignored the fabric.** `P2P._query_p2p_table`
   hardcoded `inter_node_bw`, so a topology that left the scale-up domain was
   priced as if it had not; `inter_rack_latency` was declared in gb200/gb300 and
   read by nobody.

This PR closes both, layered so each piece is reviewable on its own:

- **Per-pool hardware model, resolution & routing (SDK/Task).** New per-pool
  overrides `afd_{prefill,f}_system_name`, each defaulting to `None` so an unset
  pool inherits the top-level system. The A pool **is** the top-level
  `system_name` by definition (the same anchoring hetero-disagg uses), so there is
  no `afd_a_system_name` — "A on different hardware" is spelled by moving
  `--system`. `afd_pool_triple` is the single source of truth for a pool's
  `(system, backend, version)`; each pool resolves its **own** latest perf-DB
  version and loads its **own** GPUs-per-node, so node width, the HBM check, and
  the topology budget/pinning gates are per pool. Cross-pool A2F/F2A transfers are
  priced at the **slower** endpoint (bandwidth = `min` of both sides). All pools
  share the single top-level backend.
- **CLI surface.** `--afd-prefill-system` / `--afd-f-system` (`estimate` exposes
  **F only** — it has no static prefill pool); flags passed outside AFD/all mode
  are rejected rather than silently ignored. `load_pool_database` /
  `_resolve_version_for(honor_version_pin=False)` mirror the Task-layer
  resolution; a pinned `--perf-db-version` is honored only for the top-level
  system. Adds the `afd_hetero` example config.
- **Generator guard.** AFD pools are modeling-only in the generator (no per-pool
  `NodeConfig`), so the bridge now reads `afd_pool_overrides()` and **fails fast**
  when a pool diverges from the top-level system instead of emitting artifacts
  priced on the wrong hardware.
- **Super-node fabric tiering (Rust engine + Python mirror).** Adds
  `NodeSpec.inter_rack_latency` and `SystemSpec::get_p2p_latency` (the latency
  counterpart of `get_p2p_bandwidth`), and `P2POp.span_gpus` to select the tier.
  Only the cross-pool P2P legs take the span; F-side all-gather/reduce-scatter are
  intra-node and `a_combine` is a local HBM reduce, so neither can cross a rack.
- **Docs.** Heterogeneous AFD pools documented in the legacy AIC user guide and
  `advanced_tuning`.

Intended consumer: the AFD sweep/pareto path (`Task`, `sweep`), CLI users running
`serving-mode afd`/`all`, and the Rust perf engine's cross-pool P2P pricing.

## Review map

- Risk level: **high**
- Start with:
  - `python/aisimulate/src/aiconfigurator/sdk/task_v2.py` — `afd_pool_triple`
    resolution and the per-pool invariants (the shared `prefill_*` role fields are
    only a top-level mirror for the static prefill pool's model config, never its
    hardware).
  - `crates/core/src/perfmodel/operators/communication.rs` — tier selection and
    the `KNOWN CONSERVATISM` block; and `crates/core/src/perfmodel/common/system_spec.rs`
    — `get_p2p_latency` and the load-bearing cross-rack **clamp**.
  - `python/aisimulate/src/aiconfigurator/sdk/operations/afd_transfer.py` —
    min-endpoint pricing and where the A+F span reaches only the cross-pool legs.
  - `python/aisimulate/src/aiconfigurator_core/sdk/config.py` — `AFDConfig`
    per-pool width normalization (`a_gpus_per_node` / `f_gpus_per_node`).
  - `python/aisimulate/src/aiconfigurator/cli/main.py` — `_collect_afd_pool_kwargs`
    out-of-mode rejection and `honor_version_pin=False` version resolution.
  - `python/aisimulate/src/aiconfigurator/generator/module_bridge.py` — the new
    `serving_mode == "afd"` guard.
  - Cross-layer contract: the Rust `get_p2p_latency` and its Python mirror in
    `aiconfigurator_core/sdk/system_spec.py` **must agree on the number** (Rust
    clamps silently — no logging dep in the crate; Python warns on the same yaml).
- Public or serialized contract changed: **yes**
  - New `Task` fields `afd_{prefill,f}_system_name`; new `AFDConfig` fields
    `a_gpus_per_node` / `f_gpus_per_node` — all default `None`.
  - New public CLI flags `--afd-{prefill,f}-system`; new example
    `cli/exps/afd_hetero.yaml`.
  - New `serde(default)` Rust fields `NodeSpec.inter_rack_latency` and
    `P2POp.span_gpus` (older serialized specs still load); new public
    `SystemSpec::get_p2p_latency`.
  - Behavior change: generator raises `ValueError` for hetero AFD pools.
- Compatibility or rollback concern:
  - An unset pool inherits the top-level system and **reuses the exact same
    database object**, so identity-based hetero detection and the homogeneous path
    stay byte-for-byte unchanged; `test_no_orphan_fields` whitelists the two new
    dynamically-read fields.
  - `span_gpus = None` / `num_gpus = None` reproduces the legacy flat
    `inter_node_bw + p2p_latency` pricing **bit-for-bit**, which every
    pipeline-parallel call site relies on.
  - An explicitly overridden pool that cannot resolve perf data or a DB version
    **fails fast** rather than silently falling back to the top-level database.
  - Candidate enumeration still walks the A pool's node width, so a hetero sweep
    covers a conservative subset — never mispriced, just incomplete.
  - Rollback = drop the override fields / span wiring; plain AFD and all PP paths
    are unaffected.

## Evidence

- Tests: **376 passed** across every Python test file the PR touches.
- Fast CI (local equivalent): `ruff check` → *All checks passed!*;
  `ruff format --check` → *already formatted*; `cargo fmt --all --check` → *clean*
  (rustfmt 1.8.0, exit 0); documentation link checker → *"Checked local
  destinations in 66 Markdown files"* (exit 0).
- Full CI (local equivalent): compiled Rust `--lib` tests pass offline
  (`cargo test -p aisimulate-core --offline --lib`).
- Before/after output, trace, benchmark, or golden diff: before — a single system
  for all pools and cross-pool legs always priced at `inter_node_bw`/`p2p_latency`.
  After — per-pool `(system, backend, version)` triples, min-endpoint cross-pool
  pricing, and rack-crossing spans priced on the scale-out fabric; homogeneous
  runs reuse the identical DB object and `span=None` output is bit-identical to
  before (no numeric drift).

## Tracking

- Closes:
- Related PRs or issues: https://github.com/ai-dynamo/aiconfigurator/issues/888


crates/core/src/perfmodel/operators/communication.rs
```diff
@@ -419,6 +419,15 @@ pub struct P2POp {
...
     pub seq_split: u32,
+    /// Number of GPUs the link spans, used to pick the fabric tier
+    /// (intra-node / intra-rack / inter-rack). `None` keeps the flat
+    /// `inter_node_bw` + `p2p_latency` pricing, which is what every
+    /// pipeline-parallel caller wants: adjacent PP ranks are one hop apart
+    /// and their span is not the deployment size. Cross-pool callers (AFD)
+    /// set it so a topology leaving the scale-up domain is priced on the
+    /// scale-out fabric instead. `serde(default)` keeps older specs loadable.
+    #[serde(default)]
+    pub span_gpus: Option<u32>,
```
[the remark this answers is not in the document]
LPhgh: fixed

python/aisimulate/src/aiconfigurator/sdk/pareto_analysis.py
```diff
@@ -1124,11 +1161,14 @@ def afd_pareto(
...
-                tp_f = int(n_f_nodes) * int(gpus_per_node)
+                tp_f = int(n_f_nodes) * effective_f_gpus_per_node
                 if int(f_moe_ep_size) <= 0 or tp_f % int(f_moe_ep_size) != 0:
                     continue
                 f_moe_tp = tp_f // int(f_moe_ep_size)
-                n_a_workers = (int(n_a_nodes) * int(gpus_per_node)) // int(tp_a)
+                n_a_workers = (int(n_a_nodes) * effective_a_gpus_per_node) // int(tp_a)
+                # A-pool TP cannot span more GPUs than one A node holds.
+                if int(tp_a) > effective_a_gpus_per_node or n_a_workers <= 0:
+                    continue
```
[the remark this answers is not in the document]
LPhgh: fixed

python/aisimulate/tests/unit/sdk/test_afd_transfer.py
```diff
@@ -466,3 +466,130 @@ def test_single_gpu_node_zeroes_collectives(self, engine):
...
+            topk=0,
+        )
+        kwargs.update(overrides)
+        return AFDTransfer(name=f"afd_{direction}_transfer", scale_factor=1.0, direction=direction, **kwargs)
+
+    def test_without_peer_matches_pre_hetero_behavior(self, engine2):
+        op = self._op()
+        assert float(op.query(self.FAST, x=64)) == pytest.approx(float(op.query(self.FAST, x=64)))
+        vols = engine2.volumes(self.FAST)
+        assert vols[0] == vols[-1]
```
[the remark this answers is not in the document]
LPhgh: fixed

python/aisimulate/src/aiconfigurator/sdk/inference_session.py
```diff
@@ -1240,36 +1286,55 @@ def _build_afd_comm_ops(self, a_model, f_model, *, rank_mapping: str = "one_to_o
...
+        # legs get it: on a super-node fabric (NVL72-class) a pool pair that
+        # fits one scale-up domain is priced on NVLink, one that exceeds
+        # num_gpus_per_rack on the scale-out fabric. The F-side AG/RS are
+        # intra-node and a_combine is a local HBM reduce, so neither can
+        # cross a rack and neither takes a span.
+        #
+        # ``f_moe_tp_size`` is the mirror case: only the F-node collectives
+        # take it, because it gates whether a token-dimension TP group exists
+        # at all. The cross-pool transfers move tokens regardless.
+        span_gpus = cfg.n_a_nodes * cfg.effective_a_gpus_per_node + cfg.n_f_nodes * cfg.effective_f_gpus_per_node
```
yunyang1999, line 1306: I think the key point to make explicit is: for supernode / NVL72 systems, heterogeneous pairs that resolve to different device racks should not be priced as NVLink.

On GB200/GB300 NVL72, NVLink/NVSwitch is the **scale-up fabric within a rack**, while communication **between racks** goes over the scale-out fabric (InfiniBand). 

So if two pools resolve to different systems (for example `gb200` vs `gb300`), they are not in the same NVLink domain and should skip the rack-local tiers entirely. In that case we should always price them as scale-out (`inter_rack_bw` if present, otherwise `inter_node_bw`), regardless of `span`. 

The current issue is that we sum GPU counts across the two pools, use that as `span`, and then let each side independently tier against its own spec. That allows heterogeneous pairs on rack-tier specs to incorrectly hit the intra-rack NVLink bandwidth, which is optimistic for cross-system communication.

python/aisimulate/src/aiconfigurator/sdk/pareto_analysis.py
yunyang1999, line 1208: The F pool inherits quant modes resolved for the A system. f_model_config = copy.deepcopy(base_model_config) overwrites only parallelism fields, and there is no per-pool quant re-resolution anywhere in pareto_analysis.py. Upstream, Task._resolve_quant_modes uses roles = ["agg"] for AFD, and _apply_fmha_data_fallback's nvfp4→nvfp4_wo downgrade keys on _role_attr("agg", "system_name").

So for the headline h200_sxm A + b200_sxm F pairing, the B200 pool is modeled with H200-resolved quant (no NVFP4 promotion) — understating exactly the pool the feature exists to accelerate.

The single-point path gets this right (cli/api.py:2521-2524 resolves a_model_config against system_name and f_model_config against f_system), so estimate and default-mode sweep disagree on the same hetero config. These two paths need to be reconciled.

python/aisimulate/src/aiconfigurator/sdk/task_v2.py
yunyang1999, line 2098: A hetero prefill pool is enumerated against the top-level system. build_disagg_parallel_lists(prefill_system=self.system_name, decode_system=self.system_name, ...) passes the top-level system, not afd_prefill_system_name. That argument is load-bearing (task_v2.py:250: dense models on gb200/gb300 get num_gpu_per_worker/tp_list extended to 16 and pp_list pinned to [1]), so --system h200_sxm --afd-prefill-system gb200 enumerates H200-shaped prefill candidates.

The new prefill_gpus_per_node doesn't compensate: _enumerate_afd_prefill_options only consults it when prefill_parallel_config_list is None (pareto_analysis.py:439), and that list is always built and supplied when afd_combined_with_pd is true (task_v2.py:2097) — so the new argument is dead on every Task-driven path.

Net effect: --afd-prefill-system currently only affects DB selection, not topology enumeration, so hetero prefill isn't really in effect.

crates/core/src/perfmodel/config.rs
```diff
@@ -80,7 +80,10 @@ pub const ENGINE_CONFIG_SCHEMA_VERSION: u32 = 1;
 //   batch/query widths and FpmForwardOp gained verify_width. Upstream used
 //   14/15, already occupied here; these are positional bincode layout changes.
 //   TokenScale was appended to remap draft query widths before op lookup.
-pub const ENGINE_SPEC_SCHEMA_VERSION: u32 = 18;
+// - 19 (AFD super-node fabric tiering): `P2POp` gained `span_gpus` — a
+//   positional bincode op-layout change (the serde default covers JSON
+//   only, same class as v12/v14).
+pub const ENGINE_SPEC_SCHEMA_VERSION: u32 = 19;
```
yunyang1999, line 86: 19 is already taken on main by a different schema change, so this PR should not reuse it.

Please bump this change to 20 instead, and keep both changelog entries.

python/aisimulate/src/aiconfigurator_core/sdk/operations/afd_transfer.py
```diff
@@ -131,8 +165,13 @@ def direction(self) -> str:
 
     @property
     def num_f_nodes(self) -> int:
-        """Physical F-node count: ``ceil(n_f_workers / gpus_per_node)``."""
-        return max((self._n_f_workers + self._gpus_per_node - 1) // self._gpus_per_node, 1)
+        """Physical F-node count: ``ceil(n_f_workers / f_gpus_per_node)``."""
+        return max((self._n_f_workers + self._f_gpus_per_node - 1) // self._f_gpus_per_node, 1)
```
yunyang1999, line 169: f_gpus_per_node is the right signal to add, but _afd_send_prob() still consumes it as if expert sharding were over physical nodes.
So with identical worker count / EP layout / model, changing only F-node width changes dispatch volume, which is a modeling artifact rather than a hardware effect.

This bug predates the PR, but this change makes it hardware-sensitive. Ideally _afd_send_prob() should follow EP groups, not physical nodes; otherwise please document the limitation and file a follow-up.

LPhgh: This behavior is expected. In the AFD comm model, tokens are first delivered to the F node over the cross-pool P2P link and then spread to every F-GPU inside that node by the intra-node collective (AFDFAllGather / AFDFReduceScatter). The volume charged on the link is therefore the partner node's token demand, not one EP rank's share — which is why the F-node layout legitimately affects dispatch volume.

We will refine the comm modeling in follow-up work and look into this further, but it is out of scope for this PR.

yunyang1999: Thanks — that clears it up and makes sense.

python/aisimulate/src/aiconfigurator/cli/main.py
```diff
@@ -2955,6 +3039,10 @@ def _run_estimate_mode(args):
             afd_phase=args.afd_phase,
             afd_combined_with_pd=getattr(args, "afd_combined_with_pd", True),
             afd_boundary_on_attn=not getattr(args, "boundary_on_ffn", False),
+            **_collect_afd_pool_kwargs(
```
xutizhou: This validation only runs inside elif estimate_mode == "afd", so it cannot reject AFD-specific flags passed to other modes. For example, --estimate-mode agg --afd-f-system b200_sxm is accepted by the parser, but skips this branch and reaches cli_estimate() without the F-pool override or an error.

python/aisimulate/src/aiconfigurator/sdk/task_v2.py
```diff
@@ -3334,14 +3577,23 @@ def _run_afd_single_point(self, database):
...
         n_f_nodes = self.afd_n_f_nodes
         tp_a = self.afd_tp_a
 
+        # The A pool is the top level by definition, so it uses the shared
+        # database object directly; only the F pool can sit on different
+        # hardware. Reusing the shared object keeps the homogeneous path
+        # byte-for-byte identical (hetero detection downstream is an identity
+        # check). Both pools share the single top-level backend.
+        f_system = self.afd_pool_attr("afd_f", "system_name")
+        f_database = self._afd_pool_database("afd_f", shared_database=database)
```
liz-badada: The pinned `Task.run()` path switches to `f_database` here, but `f_model_config` below is still copied from the A-resolved config. No F-specific DSV4/NVFP4 or quant resolution runs in this path, unlike `cli/api.py::_run_afd_estimate`.

A pinned heterogeneous task can therefore report different F latency and memory results from the API for the same topology.

Suggested change: resolve `f_model_config` against `f_system` before creating the session, and add a pinned-topology regression where the selected quant mode differs between pools.

liz-badada (changes requested):
One additional blocking issue in the pinned AFD Task path.

LPhgh:
Thanks for the feedback! Based on our previous discussion, this PR is on hold until the basic AFD functionality has been adapted. We can resume the work here once that is complete.

PR #284, opened 2026-09-18
feat(perfmodel): load DCP self-benchmark profiles

tedzhouhk:
## Summary

Kimi K3 TP8+DCP8 self-benchmark profiles could not be consumed through the canonical estimator API: DCP was absent from configuration and cell matching, runtime attention labels and mixed measurement policies were rejected, and the multimodal model wrapper blocked text FPM construction.

This change enables measured vLLM DCP timing through `ForwardPassPerfModel::best_available(config)` and its Python facade:

- Preserve optional recorded DCP through canonical configuration, compiled identity, SDK engine-cache keys, FPM cells, CLI/Replay handoff, and saved estimator configurations. DCP must divide TP and does not multiply GPU count. Missing DCP, explicit DCP1, and DCP8 stay distinct.
- Accept v6 per-row single-sample/median-of-three policies after validating each policy/repeat pair. Keep exact backend identity for `FLASHINFER_MLA` while mapping its compiler description to FlashInfer.
- Add typed FPM controls for text-only profiles and explicitly unrecorded FMHA/communication precision. Unrecorded values match null profile fields exactly, with conflicting explicit precision overrides rejected. Text FPM construction retains encoder weights in both Python and compiled weight accounting.
- Pass prediction worker timing precision/backend settings to the canonical constructor and use the selected KV precision for automatic transfer-byte sizing. Recommendation rejects these overrides until its feasibility preflight supports the same identity. DCP op-level/SOL-dependent timing, regression selection (including automatic fallback), and automatic DCP/hybrid KV sizing fail explicitly; Replay requires fixed KV capacity and explicit transfer bytes when offload or P/D transfer is configured.

Canonical construction and the direct compilation adapter share the Rust FPM option parser and quantization-conflict validator. Legacy migration validates the canonical configuration before returning it.

The user-facing FPM onboarding guide explains when self-collection is useful, which engine-iteration differences it captures, and which behaviors (including PP, scheduling, and cache lifecycle) require separate support. It defines seven onboarding steps and then illustrates them with the collected Kimi TP8+DCP8 profile and a MiniMax collection campaign. Backend status explicitly covers vLLM today and forthcoming SGLang/TensorRT-LLM support; new models may need benchmark and consumer adaptation. Both examples use the canonical API and request-scoped systems roots.

This covers timing consumption and configuration identity. KDA checkpoint/eviction accounting, native hybrid prefill chunk alignment, DCP operator-level performance modeling, and full hybrid-cache replay fidelity remain separate work. No benchmark values or existing golden answers are changed.

Existing custom parallelism presets retain their original serialization. `FLASHINFER_MLA` normalization applies to both native candidates so default `auto` selection also works for DCP1 and unrecorded DCP.

## Validation

- Executed the guide's published-profile download, measured prefill/decode queries, and CLI examples directly from its code blocks: Kimi **2/2 requests**, MiniMax **8/8 requests**. All **17 Bash blocks** pass syntax checks and in-page links resolve. GPU collection commands were retained and were not rerun for this documentation update.

- Python SDK, public API, CLI, and estimator suites: **495 passed**, covering custom-preset round trips, cache/precision isolation, DCP selection, direct compiler option rejection, and validated legacy migration.
- Rust library suite with Python bindings: **1,449 passed, 1 ignored**. External Rust public API: **7 passed**.
- Both engine-step and compile-engine golden parity suites: **386 passed**, no golden files changed.
- Real merged dataset at HF revision `6fad3f9a0df5a24603108dcea0d201259254b904`: all **669 primary points** return their stored latency exactly, with correction disabled. Synthetic-attention boundary and nonuniform profiles were not mixed into the primary library.
- Independent real-profile decode interpolation: **317 interior midpoints** match adjacent measured values.
- CLI Replay smoke with the real profile, TP8/DCP8, explicit capacity, text input, and prefix caching disabled: **2/2 requests completed**, 4 output tokens, 8 physical GPUs. This verifies integration, not hardware accuracy or KDA cache fidelity.
- Changed SDK-file pre-commit hooks, the full Fast CI Ruff/format scope, whitespace, documentation links, and packaged legal-file checks pass. No unrelated hook edits or golden changes.

Dataset: https://huggingface.co/datasets/nvidia/aisimulate-fpm-dataset/discussions/10


crates/core/src/python.rs
```diff
@@ -672,6 +675,10 @@ fn materialize_aic_capacity(
     if capacity_is_explicit {
         return Ok(());
     }
+    ensure!(
+        config.dcp.is_none_or(|dcp| dcp == 1),
+        "DCP FPM replay requires explicit KV block capacity; automatic DCP/hybrid sizing is unsupported"
+    );
```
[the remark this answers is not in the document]
tedzhouhk: Added direct Rust coverage in f7934d42. The test rejects inferred DCP4 capacity before calling the estimator, preserves explicitly supplied DCP4 blocks, and verifies that DCP1 still uses the estimator. This exercises `materialize_aic_capacity` directly rather than relying on the Python compiler guard. All 29 `python::tests` passed.

python/aisimulate/src/aisimulate/config/engine.py
```diff
@@ -72,6 +72,10 @@ class ParallelismPredictionConfig(StrictModel):
     moe_expert: PositiveInt = 1
 
 
+class ParallelismPredictionConfig(ParallelismPresetConfig):
+    decode_context: Annotated[int, Field(strict=True, gt=0)] | None = None
```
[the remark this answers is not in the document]
tedzhouhk: Documented the support boundary in f7934d42 in both the Core API and onboarding guide: `decode_context` is currently supported by AISimulate `--stack engine` only; Dynamo Replay/Planner does not yet support this field. The downstream configuration propagation, cache identity, and compatible dependency pin require a separate update. This PR does not claim Dynamo DCP integration support.

python/aisimulate/tests/unit/sdk/test_fpm_forward.py
```diff
@@ -603,3 +603,177 @@ def test_genonly_step_works_with_single_op(self, fpm_session):
...
+        return InferenceSession(model, database, get_backend(BACKEND)).run_static_latency_only(runtime, "static_ctx")
+
+    _engine_handle_cache_clear()
+    try:
+        assert prefill(config) == pytest.approx(7.0)  # synthetic measured prefill row
+        for different in (
+            replace(config, dcp_size=1),
+            replace(config, fpm_unrecorded_quant_modes=()),
+            replace(config, fpm_attention_backend="auto"),
+        ):
```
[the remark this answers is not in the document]
tedzhouhk: Added coverage in f7934d42 using both relevant paths. After warming the Kimi text-only cache, `fpm_text_only=False` must raise `NotImplementedError` during model construction, so it cannot be added to the existing `PerfDataNotAvailableError` loop as written. A separate test uses a supported text model with both settings, verifies that they receive distinct cached engine handles, and checks both against the fixture's measured 40 ms prefill row. The FPM/CLI suites pass: 177 tests.

python/aisimulate/docs/fpm/README.md
```diff
@@ -5,16 +5,23 @@ SPDX-License-Identifier: Apache-2.0
...
+Start with the [self-benchmark and FPM onboarding guide](self-benchmarking-and-onboarding.md).
+Its seven steps cover support checks, measurement planning, collection, profile
+validation, canonical model construction, Replay integration, and accuracy
+validation. It explains which engine-iteration differences self-collection can
+capture and which behaviors, including PP, need separate simulator support.
+
+Self-benchmark collection currently supports vLLM configurations that pass
+model/runtime validation; new architectures can require benchmark adaptation.
+SGLang and TensorRT-LLM support is coming soon. The guide includes the collected
+[Kimi K3 TP8+DCP8 profile](self-benchmarking-and-onboarding.md#example-a-onboard-the-collected-kimi-k3-tp8dcp8-profile)
```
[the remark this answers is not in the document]
tedzhouhk: Already correct in the reviewed head: the link ends in `#example-a-onboard-the-collected-kimi-k3-tp8dcp8-profile`, matching the full heading `Example A: onboard the collected Kimi K3 TP8+DCP8 profile`. Verified the target file and heading anchor, plus the repository documentation-link check. No additional source change is needed.

crates/core/src/perfmodel/fpm/model.rs
```diff
@@ -645,6 +645,14 @@ impl ForwardPassPerfModel {
 fn build_native_candidate(
     config: &ForwardPassPerfModelConfig,
 ) -> Result<(Engine, PathBuf), AicError> {
+    if config.dcp.is_some_and(|dcp| dcp > 1)
```
jasonqinzhou: The DCP restriction is still applied only inside `build_native_candidate`, while `best_available` returns `FpmRegression` before calling this helper. At this exact head, `dcp=8` with `estimation_mode="fpm_regression"` still constructs a model whose selected mode is `fpm_regression`; `auto` can likewise reach regression after native candidates miss. That contradicts the documented measured-vLLM-interpolation-only DCP contract. Please enforce the restriction before candidate dispatch (covering explicit and automatic regression) and add tests for both paths.

python/aisimulate/src/aiconfigurator_core/sdk/engine.py
```diff
@@ -456,6 +464,9 @@ def compile_engine(
             attention_backend=attention_backend,
             speculation=resolved_speculation,
         )
+        model_config.fpm_text_only = fpm_options.get("text_only", False)
+        model_config.fpm_unrecorded_quant_modes = tuple(fpm_options.get("unrecorded_quant_modes", ()))
```
jasonqinzhou, line 472: This public `compile_engine` path still copies `fpm_options` directly without the canonical typed validation. At this exact head, direct calls successfully produced compiled engines for explicit `fmha_quant_mode="bfloat16"` plus `unrecorded_quant_modes=["fmha"]`, the analogous communication conflict, and an unknown mode token. `FPMForwardOp` then blanks a conflicting identity and can match a null-profile field despite the explicit override. Please share the canonical option parsing/validation here and add direct SDK rejection tests for conflicts and unknown values.

crates/core/src/perfmodel/py.rs
```diff
@@ -1700,6 +1717,7 @@ impl PyForwardPassPerfModel {
             tp: request.tp_size,
             pp: request.pp_size,
             attention_dp: request.attention_dp_size,
+            dcp: request.dcp_size,
```
jasonqinzhou, line 1738: `migrate_legacy_config` now copies `dcp_size` into the canonical request but serializes it without calling `ForwardPassPerfModelConfig::validate`. I reproduced a legacy `tp_size=8, dcp_size=3` input returning a `ForwardPassPerfModelConfig(tp=8, dcp=3)`; only a later `normalize_config` call rejects it. Since this helper promises a canonical migrated configuration, please validate before serialization and cover valid DCP preservation plus zero/non-divisor rejection.

jasonqinzhou (changes requested):
**REQUEST_CHANGES · 6.2/10 · high confidence**

## What this PR does

This PR adds exact DCP identity and self-benchmark FPM profile consumption across the canonical Rust/Python estimator, profile loader, compiled-engine cache, Replay-facing prediction configuration, and a new end-to-end onboarding guide.

The update makes unsupported recommendation inputs explicit, closes the DCP/FPM engine-cache identity gap with focused behavioral tests, and documents the prediction-only and downstream Dynamo boundaries clearly.

## Why this score

The 6.2 score and REQUEST_CHANGES recommendation reflect two remaining public paths that can still select or encode timing under the wrong identity, plus a legacy migration helper that emits an invalid canonical DCP config. The incremental commits fixed the earlier silent recommendation lowering and DCP recommendation ambiguity, strengthened cache/capacity coverage, and improved documentation, but DCP still bypasses its measured-interpolation restriction through regression and direct compile_engine callers still bypass typed FPM-option validation. Claude Fable timed out before returning structured output, so this remains a Codex-backed result and is not dual-reviewed.

crates/core/src/perfmodel/engine/runtime.rs
```diff
@@ -1933,6 +1933,7 @@ mod tests {
             forward_model: None,
             kv_block_size: None,
             parallel: ParallelMapping {
+                dcp_size: None,
```
simone-chen, line 1936: @coderabbitai Should we do None or 1 by default?

tedzhouhk: We use `None` by default, consistent with `moe_tp_size` and `moe_ep_size` in the construction API.

Issue #285, opened 2026-09-18
Track mocker/replay correctness and resource-efficiency fixes


PeaBrane:
Track ten mocker/replay findings from a read-only review of main at `fbd465d9ac14d37a6e73371ba4736a9364371eb5`. The scope covers the simulated engine, scheduling/cache behavior, replay/load generation, and replay reporting. AIC kernel collection and timing-model work are excluded.

Fifteen focused reviewers inspected the scope; the findings below were consolidated and checked against callers, contracts, and existing tests. **No patches, regression tests, or benchmarks were run.** These are high-confidence static findings; the triggers and regression checks below still need execution during implementation. Performance magnitudes are derived examples, not measurements.

The first eight candidates have small, localized patch directions. The last two are supported liveness bugs whose fixes need more careful behavioral validation. Check off an item when a linked fix and its validation are complete.

**Small-fix candidates**

- [ ] **1. SGLang prefill uses the configured chunk budget instead of the remaining budget.**
  - Code: [prefill.rs:103](https://github.com/ai-dynamo/aisimulate/blob/fbd465d9ac14d37a6e73371ba4736a9364371eb5/crates/core/src/engine/scheduler/sglang/prefill.rs#L103).
  - Trigger: block size 4, chunk budget 8, ample KV and request slots, and queued disjoint prompts of lengths 4 and 8. The second request is rejected for this pass instead of using the remaining four-token chunk. With block size 4, chunk budget 6, and prompt length 5, the page-rounded charge is 8 and admission never progresses.
  - Proposed fix: compare page-rounded input against the remaining chunk budget before selecting full versus partial prefill.
  - Regression check: the 4+8 case admits both with four materialized tokens each; the page-four/chunk-six/prompt-five case completes.

- [ ] **2. Missing timestamps collapse real Mooncake arrival gaps.**
  - Code: [trace.rs:863](https://github.com/ai-dynamo/aisimulate/blob/fbd465d9ac14d37a6e73371ba4736a9364371eb5/crates/core/src/replay/loadgen/trace.rs#L863); public native path: [python.rs:1049](https://github.com/ai-dynamo/aisimulate/blob/fbd465d9ac14d37a6e73371ba4736a9364371eb5/crates/core/src/python.rs#L1049).
  - Trigger: one row omits its timestamp and another has timestamp 100 ms. Normalization excludes the omitted value when finding the origin, then the driver maps it to zero: both requests arrive at zero.
[...]

PR #289, opened 2026-09-18
fix(replay): correct scheduling, telemetry, and resource retention

PeaBrane:
## Summary

Fix eight mocker/replay issues tracked in #285:

1. Fit SGLang prefill chunks to the remaining page-rounded budget. For example, 4- and 8-token prompts with an 8-token budget now compute 8 tokens in the first pass instead of leaving half the budget unused.
2. Treat missing trace arrival timestamps as zero when normalizing the origin, preserving a mixed `[missing, 100]` trace as `[0, 100]`.
3. Calculate accept length from scheduler-owned decode/verification counters. Exclude prefill and rejected/retracted work; count accepted draft tokens plus the base token before output limits trim a burst. A `1 + 3 + 3` prefill/decode sequence therefore reports `6 / 2 = 3`. Ordinary non-speculative decode retains an explicit simulator baseline of 1.
4. Suppress SGLang cache events when prefix caching is disabled.
5. Skip destination headroom scans when no handoff needs them, while preserving held reservations.
6. Avoid scanning every session to determine whether replay has drained when the ready queue already proves work remains. Retain the fallback for stale entries from failed agentic plays.
7. Release cumulative token buffers when delta sessions end.
8. Shrink radix prefix vectors after splitting an edge so short prefixes do not retain the original allocation.

Add public behavioral regressions and replace two tests tied to internal prefill state. The temporary parity, resource, and adversarial loadgen harnesses are excluded from this PR. AIC collection/timing work and the speculative-allocation/G3 liveness candidates in #285 remain deferred.

## Review map

- Risk level: low; localized scheduler, reporting, and allocation fixes.
- Start with `engine/protocol.rs` and the vLLM/SGLang decode emitters for acceptance provenance, then `sglang/prefill.rs` and `replay/loadgen/driver.rs`. The handoff fast paths must preserve held capacity and the drain check must tolerate skipped queued nodes.
- The Rust engine API gains `DecodeAcceptance` and `PassCompletionEffects.decode_acceptance`; callers constructing that struct must supply the new field or use `Default`. Serialized schemas are unchanged. Observable corrections affect chunk scheduling, mixed-timestamp arrivals, accept-length telemetry, and disabled-cache events.
- Tests exercise engine passes, emitted artifacts, telemetry, trace loading, session cancellation, and prefix reuse after eviction.

## Validation

- **72 additional speculative before/after cases against `03e88673`:** three acceptance-rate patterns, three seeds, four output lengths, and aggregated/disaggregated vLLM execution. All requests complete with exact per-request and simulation-output parity; four aggregate latency statistics permit `1e-14` floating-point roundoff. The 37 baseline cases and both temporary adversarial preemption workloads also passed again after rebuilding the extension.
- **Memory regression sensitivity:** the new radix-capacity test fails when split shrinking is removed, and the strengthened session test fails when buffer release is replaced with `clear()`. Both pass with the fixes restored.
- **37 baseline parity cases:** all per-request records and simulation results match across prediction fixtures, vLLM/SGLang/TRT-LLM, aggregated/disaggregated execution, and caching on/off. Random UUIDs and measured host-throughput fields are excluded; two online standard-deviation fields permit `1e-14` roundoff. The intentionally corrected behaviors have separate regression expectations.
- **Temporary adversarial loadgen:** both burst and concurrency-capped arrivals force 23 preemptions/readmissions across 24 requests. All 192 output tokens complete in 189 ms simulated time. Guards are 1 second virtual time and a 10-second subprocess timeout; the harness is not committed.
- **Illustrative CPU resource probes against `fbd465d9`:** 40,000 sparse arrivals took 1.668s → 0.107s. A 32,768-token radix seed plus 512 shared-prefix branches used 278.4 MiB → 25.9 MiB peak RSS, completing all 513 requests. These are single-run probes, not statistical benchmarks. The delta-session probe showed no meaningful change in peak RSS.
- New chunk-budget, timestamp, accept-length, and disabled-cache regressions failed on the baseline and pass with these fixes. Boundary coverage includes cached prefixes, partial pages, zero/one-token outputs, speculative bursts, pending agentic cleanup, and interleaved delta histories.
- `cargo test --locked -p aisimulate-core` after the acceptance correction: **1,494 passed, 1 existing ignored**. The subsequently added radix-capacity regression and all four review-touched tests passed targeted runs.
- `python/aisimulate/.venv/bin/python -m pytest -q tests/test_unified_traffic_runtime.py tests/test_runner.py tests/test_traffic.py tests/e2e/test_unified_cli_engine.py -k 'not recommend'`: **200 passed**, including all 13 prediction CLI cases.
- After merging `main` at `72f3f441`, `cargo test --locked -p aisimulate-core --test runtime_contract --test engine`: **63 passed**. The conflict resolution preserves both the trace-arrival regression and main's context-limit tests. Cargo formatting, Ruff, and Git whitespace checks passed.
- CI on `4d30f73f` passed 63 checks; its only underlying failure was the [prediction regression gate](https://github.com/ai-dynamo/aisimulate/actions/runs/35388478319/job/105743786742), with 336 SGLang `OK → DATA_MISS` rows. It compared the PR with newer main (`a2eae639`), which included the MLA/profile fixes from #262 that the PR branch lacked. The merge imports those fixes, and prediction-model/data inputs now match main. This was a stale-base mismatch, not evidence of a flake; CI on the merged head must validate the result.

**IMPORTANT: acceptance must never be reconstructed from merged visible outputs.** Those outputs mix prefill with decode and have already been truncated by stopping limits. Prominent code comments pin the accounting rule to [SGLang](https://github.com/sgl-project/sglang/blob/5c8bd8b51b53b9b39eb1edec582ee43b21002106/python/sglang/srt/managers/scheduler_metrics_mixin.py#L80-L82) and [vLLM](https://github.com/vllm-project/vllm/blob/6e448d0ea9bf3d88d898b65449ca6dc2aec170ac/vllm/v1/spec_decode/metrics.py#L113-L114). Tests cover output-limit truncation, mixed prefill/rejection, multiple ranks, and P/D handoffs.

This aligns accounting with pinned upstream source; it is not a full GPU-engine parity claim. Existing simulator approximations for vLLM speculative prefill bursts and context-boundary draft width remain unchanged. No timing profiles or performance-model data changed.

## Review feedback

The three nonblocking suggestions are addressed, replied to, and resolved: retained-capacity regressions, removal of a temporary vector in a test, and explicit boolean naming (the old speculative boolean was replaced by acceptance-rate cases).

## Tracking

Addresses items 1–8 of #285; leaves the issue open for the deferred liveness candidates.


crates/core/src/replay/loadgen/driver.rs
```diff
@@ -2876,6 +2885,65 @@ mod tests {
...
+            let surviving = first
+                .iter()
+                .find(|turn| turn.request.tokens == [30])
+                .unwrap();
+            driver.on_output_token(ending.request_uuid, 11).unwrap();
+            driver.on_output_token(surviving.request_uuid, 31).unwrap();
+            driver
+                .on_terminal(ending.request_uuid, 1.0, terminal)
+                .unwrap();
+            driver.on_complete(surviving.request_uuid, 2.0).unwrap();
```
jthomson04, line 2939: [P3] Check retained capacity in the resource tests

This test checks surviving history, and `sglang_split_prefixes_remain_reusable_across_cache_pressure` checks cache behavior, but neither would detect a return of the original memory-retention problem. Add small assertions for released session capacity and bounded prefix-vector capacity after splits.

crates/core/tests/engine.rs
```diff
@@ -471,6 +471,125 @@ fn assert_interval_work(round: &IntervalRound, rank: usize, prefill: u32, decode
...
+        let mut engine = sglang_interval_engine(config, 1);
+        let mut now_ms = 0.0;
+        if cached_prefix > 0 {
+            submit_interval_request(&mut engine, 0, 0, cached_prefix, 0, now_ms);
+            let seed = step_interval_engine(&mut engine, &mut now_ms);
+            assert!(seed.ranks[0].outputs[0].completed);
+        }
+        for (index, &prompt) in prompts.iter().enumerate() {
+            let mut tokens = vec![0; cached_prefix];
+            tokens.extend(vec![index as u32 + 1; prompt]);
```
jthomson04: [P3] Remove the temporary allocation

Use `tokens.extend(std::iter::repeat_n(index as u32 + 1, prompt))` to avoid allocating a temporary vector just to append its elements.

crates/core/tests/artifacts.rs
```diff
@@ -226,6 +226,41 @@ fn native_and_normalized_kv_visibility_preserve_raw_order() {
     }
 }
 
+#[test]
+fn cache_events_advertise_reusable_prefixes_only_when_caching_is_enabled() {
+    for backend in [Backend::Vllm, Backend::Sglang] {
+        for enabled in [false, true] {
```
jthomson04: [P3] Use boolean prefixes

Rename `enabled` to `is_enabled`, and `speculative` in `crates/core/tests/runtime_contract.rs:106` to `is_speculative`, to make the boolean names explicit.

jthomson04 (approved):
Source-only review of 03e88673. No blocking correctness issues found. The three P3 comments below are non-blocking. Tests and CI results were not checked.

tedzhouhk (approved):

crates/core/src/replay/loadgen/driver.rs
PeaBrane: Addressed in 4d30f73f. The session test now checks that ended sessions retain zero cumulative-token capacity. A radix regression repeatedly splits a 512-page edge and bounds total retained key/page-vector capacity to twice the live page count. Both pass; removing split shrinking or replacing session-buffer release with clear() makes the corresponding regression fail.

crates/core/tests/engine.rs
PeaBrane: Addressed in 4d30f73f: the test now extends tokens with std::iter::repeat_n(index as u32 + 1, prompt), avoiding the temporary vector. The targeted prefill regression passes.

crates/core/tests/artifacts.rs
PeaBrane: Addressed in 4d30f73f: enabled is now is_enabled. The speculative boolean was removed when the acceptance regression became a table of explicit acceptance-rate cases, so there is no remaining boolean to rename. The affected cache-event and acceptance tests pass.

dreamtalen (approved):

PR #295, opened 2026-09-19
perf(timing): reduce native replay evidence overhead

jthomson04:
## Summary

Native replay added per-operation timing evidence in #211. Each prediction then copied and revalidated evidence, and each cache miss converted operator results through Python before constructing evidence again. This PR reduces that CPU work while preserving latency, power, energy, coverage, source attribution, diagnostics, defaults, and errors.

- Share immutable validated cache entries with `Arc`, and borrow them during accumulation.
- Add a private accumulation path for matching operation layouts, with reusable numeric scratch storage and validation before committing updates. Keep the checked public accumulator as the fallback and independent reference.
- Call `run_static_per_op` on the native engine returned by the canonical constructor. Keep the Python bridge as a test-only differential reference.
- Move native operation names and convert borrowed source tags directly. Build cached phases from the provider's already grouped rows, without repeated validation or duplicate-name searches.

The five original implementation commits remain separate, followed by three signed review-fix commits. The full diff is four Rust files, +1090/-87 lines. No public API, model data, golden, estimator selection, cache key/capacity, mutex, scheduler, or reporting changes.

**Rebase update (September 21):** head `4a95f4ea3f8208a96289698769890232dd55a328` is based on main `e8828036e3d32d8361e2abb819e0ee60e569f518`, including #264's corrected TRT-LLM MLA data. The rebase had no conflicts, and `git range-diff` confirms all eight PR patches are unchanged. The original measured revisions, builds and artifacts are preserved. No benchmarks were rerun after this rebase; all measurements below describe the explicitly pinned earlier revisions.

- Fix default-feature warnings, share the known source-tag table, strengthen debug-only builder contracts, use the exported stride, and improve native-provider coverage.
- Reuse one immutable empty phase and skip empty accumulation only for operation-backed state. Empty and phase-only accumulated states keep the checked path.
- Keep the independent checked accumulator/builder, release-time provider contract, branch-local selection, and existing error mapper. The wider refactors are explicitly deferred in the review replies.

The fresh comparison below measures the full PR against its pinned base. It does not isolate the two review-fix commits or make a separate performance claim for the zero-work changes.

## Review map

- Risk level: medium, because exact floating-point values and failure behavior must remain unchanged.
- Start with `crates/core/src/engine/timing.rs`: the immutable wrapper, private accumulator, and native-row builder. Review `crates/core/src/python.rs` for native-provider ownership, conversion, and cache publication.
- Public or serialized contract changed: no. The original checked constructors/accumulator remain unchanged; the native builder documents unique names in first-encounter order and checks uniqueness in debug builds.
- Arithmetic and error handling: retain ordered additions and totals reconciliation, missing/zero-energy behavior, source merging, row-error precedence, and atomic failure. FPM scalar, fixed, and polynomial timing paths are unchanged.
- Compatibility/rollback: the five implementation commits and three review commits remain separate; main's decode and diagnostics behavior is retained.

## Validation

Local rebase checks at `4a95f4ea`:

- Fresh native SDK build and default-feature workspace check pass.
- Embedded-Python Rust core: 1,585 passed, one existing ignored test. `AIC_REQUIRE_EMBEDDED_ROUND_TRIP=1` was set, so native-versus-Python parity executed.
- Power contracts and diagnostics: 167 passed.
- Formatting and diff whitespace checks pass. All eight commits retain DCO sign-offs.
- [Full CI was restarted for this exact head](https://github.com/ai-dynamo/aisimulate/actions/runs/35649460908); its result is separate from these local checks.

Local checks at review head `7fddbd927f760ce3122992eede314a3be91fbc5d`:

- Default-feature workspace check and embedded-Python core/test check pass, with zero compiler warnings.
- Debug Rust core: 1,468 passed with default features; 1,517 passed with `embed-python`. Each has one existing ignored test. Debug contract assertions are exercised.
- Power contracts and diagnostics: 167 passed.
- Native CLI: 13 passed.
- Timing/source/detail CLI tests from #272: 109 passed.
- Formatting and diff whitespace checks pass. Clippy including test code matches pinned main in both modes: 223 default-feature warnings and 224 embedded-Python warnings; no new warnings.
- The native-versus-bridge test covers public prefill/decode calls, repeated cache hits, changing geometry and exact accumulated evidence. Empty-result coverage exercises both native evidence and diagnostics.
- No goldens were refreshed. A separate final-head benchmark confirmation followed the review checks, as reported below. CI results are separate from these local checks.

The initial embedded-Python run used a stale library path in the relocated Python installation and failed importing `_contextvars`. After correcting the test environment, the complete suite passed without a source change.

The differential tests compare numeric fields by bits against the unchanged checked implementations. They cover repeated/changed layouts, duplicate-name folding, missing and positive energy, partial coverage, unknown sources, signed zero, invalid inputs, overflow, error precedence, and unchanged state after failure. Native-versus-bridge checks cover prefill/decode, prefixes, changing contexts, cache hits, provenance, invalid-mode errors, and missing data. No goldens were refreshed.

Fresh release-build qualification at `7fddbd92` has exact per-request results and exact summaries/evidence on AgentX and the fixed-ID synthetic fixture. Exclusions are only `wall_time_ms`, `processed_tokens_per_s`, `processed_output_tokens_per_s`, and generated request UUIDs. The original random-ID synthetic run has exact request results and evidence, with three retained aggregation-order differences: `mean_e2e_latency_ms` (147950.05294399973 vs 147950.05294399976), `std_ttft_ms` (363.6550848106251 vs 363.65508481062517), and `std_tpot_ms` (0.08880466051541563 vs 0.08880466051541555). The fixed-ID fixture matches exactly.

### Matched measurements before the September 21 rebase

Pinned main `7ae483e942b0a2593d55e4e30957f94640580441` versus PR `7fddbd927f760ce3122992eede314a3be91fbc5d`. Both fresh builds include #272 and #289. Separate worktrees, Python environments and Cargo targets use matching dependencies and measured data. Native `--stack engine`, CPU 4, optimized builds with debug symbols, warm filesystem caches, and no concurrent builds or profiles. Optional timing/source-detail diagnostics are off.

| Workload | Pairs | Main CPU | PR CPU | CPU change | Main wall | PR wall | Wall change |
|---|---:|---:|---:|---:|---:|---:|---:|
| AgentX | 2 | 35.4101 s | 22.8888 s | -35.36% | 35.4486 s | 22.9329 s | -35.31% |
| Synthetic long decode | 2 | 2.1636 s | 1.8550 s | -14.27% | 2.1889 s | 1.8980 s | -13.29% |

Values are medians. CPU includes the supervisor and descendants. Both workloads improve in both pairs; all eight timed runs are included, with no exclusions. Expected completion and output-token counts pass in every run. Revision order reverses in the second pair.

- AgentX nominal paired 95% intervals: CPU [-39.69%, -30.72%]; wall [-39.34%, -31.01%].
- Synthetic nominal paired 95% intervals: CPU [-14.29%, -14.24%]; wall [-26.01%, +1.63%]. The wall interval supports the 3% regression limit but does not establish a wall-time benefit.
- The original 25% AgentX CPU target and all 3% CPU/wall regression limits pass. No extra pairs were needed. These are small-sample Student-t intervals on paired log ratios, assuming independent, approximately normal ratios, without an adjustment for adaptive stopping. The narrow synthetic CPU interval describes these two pairs only.
- AgentX CPU/output token: 7.8355 to 5.0648 microseconds; internal replay: 31.6856 to 19.1076 s. Median max-process RSS: 430.49 to 435.40 MiB; sampled aggregate peak RSS: 467.21 to 471.69 MiB. Synthetic memory is approximately unchanged.

The local confirmation worktree retains `REPORT.md`, `RESULTS.md`, `REPRODUCE.md`, full provenance, qualification comparisons and every raw run. No new profile or microbenchmark was collected; the profiles below remain historical.

### Matched measurements before the final #272 rebase

Previous main `7cfa4f20` versus the earlier rebased PR `7c0e0e89`. These measurements do not include #272 or describe the post-#272 review head. Both builds include the same later main changes, including #289. These results measure the full PR, not only its last two commits.

| Workload | Pairs | Main CPU | PR CPU | CPU change | Main wall | PR wall | Wall change |
|---|---:|---:|---:|---:|---:|---:|---:|
| AgentX | 2 | 33.9092 s | 22.3678 s | -34.04% | 33.9776 s | 22.4080 s | -34.05% |
| Synthetic long decode | 2 | 2.1180 s | 1.8465 s | -12.82% | 2.1369 s | 1.8874 s | -11.68% |

AgentX: nominal paired 95% CPU interval [-43.70%, -22.70%]; wall interval [-43.61%, -22.86%]. CPU improves in 2/2 pairs.

Synthetic: nominal paired 95% CPU interval [-13.65%, -11.97%]; wall interval [-17.34%, -5.63%]. CPU improves in 2/2 pairs.

That pre-#272 two-pair comparison passed the CPU/wall regression checks. Its AgentX interval crosses the original 25% improvement target. That earlier sample did not establish the 25% target; its stopped third pair remains excluded. Pairs alternate revision order. Paired Student-t log-ratio intervals are nominal small-sample intervals, not adjusted for adaptive stopping. All trials, ranges, per-token costs, replay time and both memory measures are retained in `RESULTS.md`.

No new profile or microbenchmark was collected for either rebase. Profiles and diagnostic values below describe the preserved historical revisions only.

### Historical stage measurements (before rebase)

Two separate campaigns measure the stages. Do not add their percentages or treat them as one matched baseline-to-final measurement.

| Comparison | Workload | Paired rounds | Control median CPU | Candidate median CPU | Change |
|---|---|---:|---:|---:|---:|
| `a2eae639` → `92ea946a` (cache, accumulation, native bridge) | AgentX | 6 | 33.8926 s | 24.7750 s | -26.90% |
| `a2eae639` → `92ea946a` | Synthetic long decode | 6 | 2.1007 s | 1.8827 s | -10.38% |
| `92ea946a` → `0e35fae1` (source conversion and builder) | AgentX | 2 | 24.9389 s | 22.4274 s | -10.07% |
| `92ea946a` → `0e35fae1` | Synthetic long decode | 4 | 1.8848 s | 1.8251 s | -3.17% |

The first campaign's AgentX nominal paired 95% interval is a 25.42–27.51% reduction; all six pairs improve. The builder campaign's interval is an 8.28–11.83% reduction; both pairs improve. Its AgentX wall time falls 10.08%, and synthetic wall time falls 2.69%. All four synthetic pairs improve. Both campaigns meet their respective 25% and 5% CPU gates and the 3% CPU/wall regression limits.

Both campaigns start with two alternating pairs and add pairs only to resolve a threshold. In the builder campaign, only the synthetic CPU interval needed two more pairs. Intervals use paired Student-t log ratios and assume independent, approximately normal pair log ratios. These are nominal small-sample intervals, not adjusted for adaptive stopping.

Native `--stack engine`, serial performance CPU 4, warm filesystem caches, matching Python/Rust/dependencies/data, separate optimized builds with debug symbols, and no concurrent builds or profiles during measurement. Process CPU includes waited and adopted descendants, verified by accounting probes. Detailed request capture is off during timing.

- AgentX: saved 32-session Qwen3.8 trace, SGLang 0.5.17, GB300, TP16, MoE TP4/EP4; 4,265 requests / 4,519,160 output tokens.
- Synthetic: 1,024 input / 4,096 output tokens, concurrency 32, 256 requests / 1,048,576 output tokens.
- Builder campaign CPU/output token: 5.5185 → 4.9627 microseconds; internal replay: 21.2230 → 18.6979 s. Median max-process RSS: 513.62 → 514.44 MiB; sampled aggregate RSS: 550.41 → 550.21 MiB.
- Separate 99 Hz AgentX profiles: evidence-validation self weight falls from 6.497 billion cycles (4.81%) to 1.279 billion (1.05%). Leaf-only unwind weight is 12.55% / 14.13%; inclusive groups overlap and incomplete stacks limit attribution.
- The construction diagnostic falls from 2.550 → 0.873 microseconds for 24 operations and 3.457 → 1.117 microseconds for 32. Native input creation is outside the timer; conversion, construction, and destruction are inside. Both variants retain 5,957 cache hits, 2,237 provider calls, and 53,688 operations. Microbenchmark improvements are not additional end-to-end savings.

The harness, configurations, traces, complete reports, raw profiles, and flamegraphs remain in the profiling worktrees and are not part of the production diff. Exact commands for the pre-rebase confirmation:

```bash
cd /home/jothomson/Desktop/dynamo/.worktrees/aisimulate-pr295-confirm-candidate-20260918
CONFIRM_ROOT="$PWD/benchmark-harness/pr295-confirmation"
python3 "$CONFIRM_ROOT/scripts/build.py" control candidate
python3 "$CONFIRM_ROOT/scripts/provenance.py"
python3 "$CONFIRM_ROOT/scripts/qualify.py" candidate
AISIM_ROUNDS='[1,2]' python3 "$CONFIRM_ROOT/scripts/campaign.py" candidate
python3 "$CONFIRM_ROOT/scripts/analyze.py" candidate
python3 "$CONFIRM_ROOT/scripts/tables.py"
```

These commands refer to the retained local harness. It refuses to replace existing run directories; use unused round numbers to repeat measurements.

No estimation or measured-data changes. The pre-rebase comparison uses matching 132 external Python packages, Cargo/uv locks, and 1,483 measured-data files (126,134,657 bytes). Full identities and config hashes are retained in `analysis/provenance.json`.

- Data manifest SHA-256: `5b017086065f0923f7db054e2aa685276f5bad2a1df6da0bc3685732591b2410`.
- AgentX input SHA-256: `d156d09055ce272e3e490bb8490ed3379e579435947a4fedd76beaed16818abe`.
- Fixed-ID input SHA-256: `7cc9fcf593f8e7b36e9076066ccc516d97f738815c87f75f71b757bf2944c10b`.
- Pre-rebase control extension SHA-256: `c955adb4395ddc8c61c7a431643fb118a793f2d1ac5cbda9515eaa544406d4c7`.
- Pre-rebase candidate extension SHA-256: `f2d378b30231e8957e8ba2294a67ad4f55e9518827942d41ccc0963357aa98d0`.
- Pre-#272 control extension SHA-256: `c364db87b30914c885afd29c6162f1c51da71012176fe1c832544dc42e23458e`.
- Pre-#272 candidate extension SHA-256: `028c123ea7e63b8005d720157ac8aacf6d87ca9e2ddfec1f33ae67ea718d12cb`.

The historical ~60% slowdown is context for this work, not a claim that all historical overhead has been removed. The historical stage comparisons exclude later main changes, including #289; the pre-rebase comparison includes them in both variants.

## Tracking

- Related: #211.
- Focused review fixes and local rebase validation are complete. Benchmark confirmation covers the preserved pre-rebase head. Wider refactors remain deferred. Full CI has restarted for the rebased head; no issue-closing claim.


jthomson04 (commented):
## Review: perf(timing) native replay evidence — 13 findings

Posting as a comment-only review: the PR is a draft, so no approve / request-changes.

### What I actually verified (please read this first)

**The fast paths are not wrong, as far as I can tell.** The central risk in this PR is that the private accumulator and the name-folded builder must stay bit-identical to the checked implementations, so I tried to break that empirically rather than by reading alone. Everything below ran against head `2019d04c`:

| Check | Result |
|---|---|
| `cargo check --workspace --offline` (default features) | compiles, **7 new warnings** — see finding 1 |
| `cargo check -p aisimulate-core --offline --features embed-python --tests` | clean |
| `cargo test -p aisimulate-core --offline --lib engine::timing` | 19/19 pass |
| Randomized differential fuzz of `TimingEvidenceAccumulator::record` vs `TimingPhaseEvidence::try_accumulate` — 400 seeds × 40 steps | **no divergence** |
| Randomized differential fuzz of `ValidatedTimingPhase::from_name_folded_operations` vs `try_from_operations` — 3000 cases | **no divergence** |

The accumulator fuzzer compared struct equality *and* `to_bits()` on every phase and per-operation `latency_ms` / `energy_wms` / `covered_latency_ms`, plus `source`, `details` and error strings, across: `None`/`Some(0.0)`/`Some(-0.0)`/subnormal/`f64::MAX * 0.4` energies, signed-zero and overflowing latencies, all five `TimingEvidenceSource` variants including `Other`, ops carrying and not carrying `details`, and randomly varying op counts and name sets so that layout changes, duplicate-name folding and the checked-fallback branch were all hit. The builder fuzzer covered unique-name inputs up to 5 ops with the same value classes including `f64::MAX`.

So: **none of the 13 findings below is an observed numeric regression.** They are contract, hygiene and structural issues. I also independently re-derived the equivalence of the two implementations by hand (error ordering, `reduce`-vs-`sum` energy reduction, the `canonicalized()` energy filter, `operation_totals_match()`, phase-only fallback, source merge) and agree with the PR's reasoning.

I separately confirmed that `run_static_per_op`'s `PerOpFold` really does name-fold in first-encounter order, that `reset_provenance` / `parse_mode` ordering matches the pyclass method exactly, that `reset_provenance` is an atomic on a shared `PerfDatabase` (so dropping the GIL introduces no new race), and that `AicPhaseProvider::Native` and `self.engine` wrap the same `Arc<Engine>`.

### Confidence

All 13 are marked **CONFIRMED** inline — each is either verified by a command I ran or directly readable from the code. Three are flagged **latent** (findings 2, 3 and the `details` aspect of 4): the defect is real in the code but not reachable today, and would be triggered by a plausible future change rather than by current inputs. Nothing is posted as speculative.

Where I proposed a concrete change I offered a `suggestion` block only after applying it and re-running the builds and tests. Four qualified: the `cfg` gate here, the `record` early-return, the `OnceLock` empty phase, and `DEFAULT_STATIC_STRIDE`. Each was checked under default features, under `--features embed-python --tests`, against `engine::timing`'s 19 tests, against my fuzzer, and with `cargo fmt --check`. The `from_provider` delegation is prose only, because the lines it would replace fall outside this PR's diff hunks.

### Findings outside the diff

None — all 13 anchor to lines inside the changed hunks.

### Retracted during review

I had a 14th finding claiming the provenance assertion at `python.rs:2255` was vacuous in the missing-data half of `native_phase_evidence_matches_python_bridge`, on the theory that `raw` and `missing` are different engines. **That is wrong and I withdraw it:** `missing` is built with `Arc::clone(raw.database())`, and `Engine::last_provenance()` reads `self.db.worst_provenance()`, so the two engines share the atomic and the assertion is meaningful. Recording it here rather than dropping it silently.

### Two things I'd genuinely push on

Of the 13, the two worth arguing about rather than just fixing are **finding 4** (the private accumulator duplicates bit-exact arithmetic that has no compile-time tie to the original — #272's `details` field already forced a hand-patch, and the next one gets no warning; a `try_accumulate_ref(&mut self, &Self)` on `TimingPhaseEvidence` would deliver the same win from one implementation) and **finding 13** (the test helper now defaults to a `cfg(test)`-only provider, so ten existing tests no longer cover the code that ships).

The measured 34% CPU reduction on AgentX looks well-earned; none of this is a reason to hold the performance work.

crates/core/src/engine/mod.rs
```diff
@@ -39,6 +39,7 @@ pub use protocol::{
 };
 pub use runtime::{Engine, EngineFactory};
 pub use scheduler::SchedulerRank;
+pub(crate) use timing::{TimingEvidenceAccumulator, ValidatedTimingPhase};
```
jthomson04, line 43: **[CONFIRMED] The new items are compiled unconditionally but used only under `cfg(feature = "python")`, so the default-feature build gains 7 new warnings.**

`ValidatedTimingPhase`, `TimingEvidenceAccumulator`, `OperationUpdate`, `from_native` and `from_name_folded_operations` live in `engine/timing.rs`, which is always compiled, but their only non-test consumer is `mod python` — which is `#[cfg(feature = "python")]` (`lib.rs:23`).

Verified on `2019d04c` with `cargo check --workspace --offline`:

```
engine/mod.rs:42:   unused imports: `TimingEvidenceAccumulator` and `ValidatedTimingPhase`
engine/timing.rs:31:  associated function `from_native` is never used
engine/timing.rs:242: struct `ValidatedTimingPhase` is never constructed
engine/timing.rs:248: associated items `from_name_folded_operations`, `from_operations`, and `as_phase` are never used
engine/timing.rs:308: struct `TimingEvidenceAccumulator` is never constructed
engine/timing.rs:313: struct `OperationUpdate` is never constructed
engine/timing.rs:320: methods `snapshot` and `record` are never used
```

CI's `cargo test --workspace` job (`ci.yml:316`) builds without `embed-python`, so it now compiles with all 7. The PR's "no new warnings" clippy result only covers the `embed-python` variant — `cargo check -p aisimulate-core --features embed-python --tests` is indeed clean.

One wrinkle: the items themselves can't take a plain `feature = "python"` gate, because `timing.rs`'s own tests use `ValidatedTimingPhase`/`TimingEvidenceAccumulator` and run under `cargo test --workspace`. What works is `#[cfg(any(feature = "python", test))]` on the two structs, on `OperationUpdate`, and on the two `impl` blocks, plus `#[cfg(feature = "python")]` on `from_native` and `from_name_folded_operations`. I applied exactly that plus the suggestion below and re-ran: default `cargo check --workspace` clean, `--features embed-python --tests` clean, 19/19 `engine::timing` tests pass, `cargo fmt --check` clean.

```suggestion
#[cfg(feature = "python")]
pub(crate) use timing::{TimingEvidenceAccumulator, ValidatedTimingPhase};
```

crates/core/src/engine/timing.rs
```diff
@@ -225,6 +236,219 @@ pub struct TimingPhaseEvidence {
...
+pub(crate) struct ValidatedTimingPhase(TimingPhaseEvidence);
+
+impl ValidatedTimingPhase {
+    /// Native rows are already name-folded, in first-encounter order, and each
+    /// operation has passed `TimingOperationEvidence::new`. Keep this private
+    /// contract separate from the checked constructor for public field values.
+    pub(crate) fn from_name_folded_operations(
+        operations: Vec<TimingOperationEvidence>,
+    ) -> Result<Self> {
+        debug_assert!(operations.iter().enumerate().all(|(index, operation)| {
```
jthomson04, line 253: **[CONFIRMED, latent] The name-uniqueness precondition is enforced only by `debug_assert!`, so it is not checked in the shipped release build.**

`PerOpFold` (`perfmodel/engine/runtime.rs:188`) does guarantee unique, first-encounter-ordered names today, so this is not currently reachable — but the guard that documents the contract is compiled out of every release binary.

If folding ever stops (a new caller feeding raw rows, a refactor of `PerOpFold`, `mixed_step_breakdown_per_op` being wired in here), a release build silently produces `operations = [("a", 1.0), ("a", 2.0)]` where the checked builder would produce the folded `[("a", 3.0)]`. That phase then feeds `TimingEvidenceAccumulator::record`, whose fast path merges **positionally** while `try_accumulate` merges by **name lookup** (`merge_operation` finds only the first `"a"`). The two implementations then disagree, and `reconcile_operation_totals` reconciles to a different answer — with no error, and with the only guard compiled out.

Cheap fix: keep the `debug_assert!` for the O(n²) scan, but also make the invariant structural — e.g. have the native path hand over a type that can only be produced by `PerOpFold`, or fold defensively here (the linear scan over ~32 names is negligible next to the work already saved).

crates/core/src/engine/timing.rs
```diff
@@ -225,6 +236,219 @@ pub struct TimingPhaseEvidence {
...
+/// Immutable provider evidence. Construction validates the public fields once;
+/// callers can borrow the result but cannot change its validated values.
+#[derive(Debug, Default)]
+pub(crate) struct ValidatedTimingPhase(TimingPhaseEvidence);
+
+impl ValidatedTimingPhase {
+    /// Native rows are already name-folded, in first-encounter order, and each
+    /// operation has passed `TimingOperationEvidence::new`. Keep this private
+    /// contract separate from the checked constructor for public field values.
+    pub(crate) fn from_name_folded_operations(
```
jthomson04, line 250: **[CONFIRMED, latent] `from_name_folded_operations` never calls `TimingOperationEvidence::canonicalized()` on its inputs, so any field mutated after `new()` skips validation.**

The doc comment states the contract ("each operation has passed `TimingOperationEvidence::new`"), but the `debug_assert!` below only checks name uniqueness — not that the ops are still canonical.

That matters because the codebase already mutates ops after `new()`: `python.rs:490-497` builds via `TimingOperationEvidence::new(...)` and then assigns `operation.details = Some(entry.details)`. It currently routes to the checked `from_operations`, and the two constructors are four lines apart and take the same argument type. If that call (or a future provider) switches to `from_name_folded_operations`, `canonicalized()`'s details validation is silently skipped — `sol` XOR `sol_unavailable_reason`, non-empty reason, finite non-negative SOL values, valid fallback inference-phase/comm-backend/topology — and malformed diagnostics reach the serialized report.

Suggest extending the `debug_assert!` to also assert each operation round-trips through `canonicalized()` unchanged, or narrowing the parameter to a type that can only be built by the validating constructor.

crates/core/src/engine/timing.rs
```diff
@@ -225,6 +236,219 @@ pub struct TimingPhaseEvidence {
...
+    energy_wms: Option<f64>,
+    covered_latency_ms: f64,
+}
+
+impl TimingEvidenceAccumulator {
+    pub(crate) fn snapshot(&self) -> TimingEvidenceSummary {
+        self.summary.clone()
+    }
+
+    pub(crate) fn record(&mut self, incoming: &ValidatedTimingPhase, prefill: bool) -> Result<()> {
```
jthomson04, line 333: **[CONFIRMED, structural] `record` hand-copies ~100 lines of `try_accumulate` + `canonicalized` + `reconcile_operation_totals` arithmetic that must stay bit-identical forever, with no compile-time link between the two implementations.**

To be clear about what I did and did not find: I fuzzed this against the checked accumulator (400 seeds × 40 steps, including `details`, `f64::MAX` overflow, signed zero, `Other` sources, duplicate names, layout changes) and found **no** arithmetic or error-ordering divergence. The concern is maintenance, not a current bug.

The cost being avoided is `try_accumulate`'s `self.clone().canonicalized()?` + `other.canonicalized()?` — two deep clones and full revalidations per call. Duplicating the arithmetic to dodge that means every future change to `TimingOperationEvidence` or `canonicalized()` has to be mirrored here by hand, and nothing fails to compile if it isn't. That has already happened once: #272 added `details`, and this PR had to patch the guard below with `left.details.is_none() && right.details.is_none()` to stay correct. The next added field or validation rule gets no such warning.

The fix at the right depth is to make the shared accumulator cheap rather than to add a second one: a `TimingPhaseEvidence::try_accumulate_ref(&mut self, other: &Self)` that stages into a caller-provided scratch `Vec` and commits after validation, with the existing `try_accumulate(other: Self)` delegating to it. Both the public boundary and the provider then get the speedup from one implementation, and the `details` special case disappears.

crates/core/src/engine/timing.rs
```diff
@@ -28,6 +28,17 @@ pub enum TimingEvidenceSource {
 }
 
 impl TimingEvidenceSource {
+    pub(crate) fn from_native(source: &str) -> Self {
```
jthomson04, line 32: **[CONFIRMED, structural] `from_native` duplicates `from_provider`'s entire tag table ten lines below it, so the two can drift silently.**

The only real difference is that `from_native` avoids allocating a `String` for known tags — worth having, but not worth a second copy of the mapping.

The drift is not hypothetical: `python.rs:1570` already special-cases a `"transferred"` source tag in reporting that neither function maps to a typed variant. If a tag like that is ever promoted to a `TimingEvidenceSource` variant in `from_provider` and not here, the production native path emits `Other(tag)` while the bridge emits the typed variant. Because the bridge is also the oracle for `native_phase_evidence_matches_python_bridge`, that surfaces as a confusing differential-test failure rather than a clear one.

Delegating keeps one table and preserves the allocation win (no allocation on the `from_native` path at all, and `from_provider` reuses the `String` it already owns):

```rust
pub fn from_provider(source: impl Into<String>) -> Self {
    let source = source.into();
    match Self::from_native(&source) {
        Self::Other(_) => Self::Other(source),
        typed => typed,
    }
}
```

(Not offered as a `suggestion` block because `from_provider`'s body is outside this PR's diff hunks.)

crates/core/src/engine/timing.rs
```diff
@@ -225,6 +236,219 @@ pub struct TimingPhaseEvidence {
...
+        self.summary.clone()
+    }
+
+    pub(crate) fn record(&mut self, incoming: &ValidatedTimingPhase, prefill: bool) -> Result<()> {
+        let phase = if prefill {
+            &mut self.summary.prefill
+        } else {
+            &mut self.summary.decode
+        };
+        let incoming = incoming.as_phase();
```
jthomson04, line 339: **[CONFIRMED] Every zero-work prediction falls through to the expensive checked path at line 342 and deep-clones + revalidates the whole accumulated phase for an update that is provably a no-op.**

`predict_phase_evidence` returns a default `ValidatedTimingPhase` whenever `batch_size == 0` or a prefill is fully cached (`python.rs:475`). `record` then hits `incoming.operations.is_empty()` in the guard below and executes `return phase.try_accumulate(incoming.clone())`, which runs `self.clone().canonicalized()?` — cloning all 24-32 accumulated operations (their `String` names, sources, `details`) and re-running every per-op and phase-level check — only to add `0.0` and reconcile back to identical values.

I worked through the no-op argument: with a fully-default `other`, `has_phase_only_totals(other)` is false, so `complete_operations` stays true and `reconcile_operation_totals` recomputes the same sums; `latency += 0.0` cannot flip a sign (phase totals are `sum()`-derived, never `-0.0`); source merge is `(Some(s), None) => Some(s)`. The only behavioural difference is that the early return skips revalidating `self`, which is an invariant the accumulator already maintains.

I applied the suggestion below and re-ran: 19/19 `engine::timing` tests pass, my 400-seed differential fuzzer still finds no divergence, `cargo check` clean in both feature modes, `cargo fmt --check` clean.

```suggestion
        let incoming = incoming.as_phase();
        if incoming.operations.is_empty()
            && incoming.latency_ms == 0.0
            && incoming.energy_wms.is_none()
            && incoming.covered_latency_ms == 0.0
            && incoming.source.is_none()
        {
            return Ok(());
        }
```

crates/core/src/python.rs
```diff
@@ -455,10 +470,10 @@ impl AicTimingModel {
         osl: u32,
         prefix: u32,
         mode: &str,
-    ) -> Result<TimingPhaseEvidence> {
+    ) -> Result<Arc<ValidatedTimingPhase>> {
         let prefill = mode == "static_ctx";
         if batch_size == 0 || (prefill && isl <= prefix) {
-            return Ok(TimingPhaseEvidence::default());
+            return Ok(Arc::new(ValidatedTimingPhase::default()));
```
jthomson04: **[CONFIRMED] This heap-allocates a fresh empty phase on every zero-work prediction, on the path the PR exists to speed up.**

Before this PR the short circuit returned `TimingPhaseEvidence::default()` by value. Now each `batch_size == 0` or fully-cached-prefill call allocates an `Arc` whose payload is a constant and drops it again immediately.

A process-wide `OnceLock` returns a clone-of-`Arc` with no allocation. Verified: compiles under `--features embed-python --tests` and default features, `cargo fmt --check` clean.

```suggestion
            static EMPTY: std::sync::OnceLock<Arc<ValidatedTimingPhase>> =
                std::sync::OnceLock::new();
            return Ok(Arc::clone(EMPTY.get_or_init(Default::default)));
```

(Pairs with the early-return suggested on `engine/timing.rs:330` — together they make the zero-work path allocation-free and clone-free.)

crates/core/src/python.rs
```diff
@@ -483,11 +498,71 @@ impl AicTimingModel {
...
-            let phase = TimingPhaseEvidence::try_from_operations(operations)?;
-            self.phase_cache.insert(key, phase.clone());
+            let phase = Arc::new(ValidatedTimingPhase::from_operations(operations)?);
+            self.phase_cache.insert(key, Arc::clone(&phase));
             return Ok(phase);
         }
-        let (context, generation) = Python::with_gil(|py| {
+        let phase = match &self.phase_provider {
+            #[cfg(test)]
+            AicPhaseProvider::NativeEntries(entries) => {
```
jthomson04, line 512: **[CONFIRMED, test seam] The `NativeEntries` arm skips the `ensure!(!entries.is_empty(), ...)` guard that both production-shaped arms apply, so it is not a faithful stand-in for the path it stands in for.**

The real `Native` arm enforces `"AIC {mode} returned empty operation evidence for nonzero work"` at line 529, and the `Python` arm repeats it at line 541. This arm feeds rows straight to `phase_evidence_from_native_entries`.

Consequence: `native_evidence_builder_matches_checked_values_and_errors` — the test that exists specifically to exercise construction failures before cache publication — cannot observe that contract, and would stay green if the guard regressed. Adding the same two lines here costs nothing and makes the seam match production.

crates/core/src/python.rs
```diff
@@ -483,11 +498,71 @@ impl AicTimingModel {
...
+                    osl,
+                    prefix,
+                    seq_imbalance_correction_scale: 1.0,
+                    gen_seq_imbalance_correction_scale: 1.0,
+                };
+                let (context, generation) = crate::py::parse_mode(mode)
+                    .and_then(|mode| {
+                        engine.reset_provenance();
+                        engine
+                            .run_static_per_op(&runtime, mode, 32)
```
jthomson04: **[CONFIRMED] The stride is hardcoded as `32` instead of the exported `DEFAULT_STATIC_STRIDE` constant.**

`DEFAULT_STATIC_STRIDE` is declared `pub const ... : u32 = 32` at `perfmodel/engine/runtime.rs:103`, re-exported from `perfmodel::engine`, and used as the default by `run_static`, `run_generation_phase`, and the pyclass signature.

If it is ever retuned, this call site silently stays at 32 while every other caller moves, so replay's per-op decode weighting diverges from `run_static` totals with no compile error. (The removed Python bridge passed the same magic number, so this reproduces rather than introduces the issue — but it is now the only production call site.)

Verified the fully-qualified form compiles under `--features embed-python --tests` and is `cargo fmt --check` clean:

```suggestion
                            .run_static_per_op(
                                &runtime,
                                mode,
                                crate::perfmodel::engine::DEFAULT_STATIC_STRIDE,
                            )
```

crates/core/src/python.rs
```diff
@@ -483,11 +498,71 @@ impl AicTimingModel {
...
+                            .map_err(crate::py::aic_to_py)
```
jthomson04, line 538: **[CONFIRMED] The new no-Python fast path routes its errors through the PyO3 boundary mapper purely to format a string, so it still takes the GIL and still depends on importing the Python sdk error module.**

`aic_to_py` (`perfmodel/py.rs:107`) calls `Python::with_gil` and `py.import("aiconfigurator_core.sdk.errors")` to build a `PyErr` that is discarded one line later by `anyhow!("AIC {mode} evidence prediction failed: {error}")`. Two consequences on a path that is otherwise Python-free:

1. A perf-data miss acquires the GIL to produce an error message.
2. `aic_to_py` is explicitly "failure-tolerant" and degrades to a bare `PyValueError` when that import fails — so the `PerfDataNotAvailableError` prefix that `native_phase_evidence_matches_python_bridge` asserts on is environment-dependent, not a property of the Rust code.

It also required widening `aic_to_py` and `parse_mode` to `pub(crate)` inside `crates/core/src/perfmodel/`, which `perfmodel/mod.rs` documents as "a stable mirror of the former AIConfigurator Rust crate ... so upstream syncs [stay] mechanical". That is a new merge-conflict point in the mirror, created to serve the replay layer.

A pure-Rust classifier over `AicError` (returning the same type-name prefix) shared by `aic_to_py` and this call site would remove the GIL dependency, remove the import dependency, and leave the mirror untouched.

crates/core/src/python.rs
```diff
@@ -483,11 +498,71 @@ impl AicTimingModel {
...
+                    })
+                    .map_err(|error| anyhow!("AIC {mode} evidence prediction failed: {error}"))?;
+                let entries = if prefill { context } else { generation };
```
jthomson04, line 541: **[CONFIRMED] Commit 4 un-hoisted this block, duplicating the phase selection and the non-empty guard verbatim into two match arms.**

Lines 528-532 and 540-544 are byte-identical except for the builder they feed. Commit 3 had this hoisted below the `match`; commit 4 pushed it inside so the native arm could avoid materializing owned source strings.

Having each arm yield `(context, generation)` and keeping one post-match selection + guard gets the same allocation behaviour with one copy — and removes the class of bug where a future change lands in only one arm. That has already happened in this PR: the `NativeEntries` arm at line 507 has no guard at all.

crates/core/src/python.rs
```diff
@@ -506,34 +581,42 @@ impl AicTimingModel {
...
-        } else {
-            evidence.decode.try_accumulate(phase)?;
-        }
-        Ok(())
+        evidence.record(phase, prefill)
     }
 }
 
-fn phase_evidence_from_python(
+fn phase_evidence_from_native_entries(
```
jthomson04, line 602: **[CONFIRMED] `phase_evidence_from_native_entries` and `phase_evidence_from_entries` (line 617) are near-identical 20-line functions differing only in the source converter and the final constructor.**

The `ensure!(energy_wms.is_finite() && energy_wms >= 0.0, "AIC operation {name:?} returned invalid energy {energy_wms}W-ms")` guard and the `TimingOperationEvidence::new(...)` mapping are copy-pasted.

This one is worth fixing specifically because the duplicate is the **oracle**: `native_evidence_builder_matches_checked_values_and_errors` compares the native builder against `phase_evidence_from_entries`. A fix applied to only one silently invalidates the comparison — the test keeps passing while measuring the native builder against a stale reference.

One function parameterised by the source converter (`fn(&str) -> TimingEvidenceSource` or a generic) and by the final constructor keeps the differential test honest.

crates/core/src/python.rs
```diff
@@ -2125,13 +2211,147 @@ mod tests {
             engine,
             diagnostic_model: None,
             decoder_replay: false,
+            phase_provider: AicPhaseProvider::Python,
```
jthomson04, line 2221: **[CONFIRMED, coverage] The `timing_model()` test helper now selects `AicPhaseProvider::Python` — a variant that exists only under `#[cfg(test)]` — so the bulk of the `AicTimingModel` suite exercises a provider that does not ship.**

Ten call sites (`python.rs:2989, 3003, 3016, 3038, 3062, 3087, 3150, 3191`, plus the two bridge constructions in the new differential test) drive the Python bridge. They cover cache reuse, evidence accumulation, per-op power coverage semantics, error precedence and the zero-work short circuits — all against a code path that is compiled out of the shipped binary.

That leaves the production `Native` arm covered only by `native_phase_evidence_matches_python_bridge`, which needs a full estimator build plus real gb300 / sglang-0.5.17 parquet data, and by injected rows through `NativeEntries` (which, per my comment at line 507, omits one of the arm's guards). A regression in the `Native` arm's `RuntimeConfig` construction, prefill/generation selection, or non-empty guard passes the entire suite except that one heavyweight data-dependent test.

Suggest flipping the default: have `timing_model()` build a `Native` provider over a small synthetic `PerfEngine` (the test already constructs one at lines 2320-2338 for the missing-data case), and keep `Python` only where the bridge is genuinely the subject.

crates/core/src/engine/mod.rs
jthomson04: Added the feature gates in the review-safeguards commit. Python-only exports and conversion methods require `python`; the accumulator types and implementations use `any(feature = "python", test)` so default-feature unit tests still exercise them. Validation now covers both default features and embedded Python, including test code, rather than only the embedded-Python library.

Verified at `7fddbd92`; the PR description lists the exact-head validation results.

crates/core/src/engine/timing.rs
jthomson04: Deferred for this PR. `run_static_per_op` already guarantees unique names in first-encounter order through `PerOpFold`. The builder remains private, retains the uniqueness assertion, and now has explicit contract tests. Adding another release-time fold repeats the work this change removes; introducing a provider-owned wrapper type would widen the change. The checked general builder still handles duplicate names independently.

crates/core/src/engine/timing.rs
jthomson04: Added debug-only assertions that each input still round-trips through canonicalization unchanged and has no attached diagnostic details. The constructor documentation now states this contract. A compact debug test covers duplicate names and mutations to names, latency, energy, coverage, and details. Diagnostic rows continue through the checked constructor after details are attached.

Verified at `7fddbd92`; the PR description lists the exact-head validation results.

crates/core/src/engine/timing.rs
jthomson04: Deferred to keep this PR focused. A shared borrowed accumulator would change the public implementation and its validation/failure paths, and would remove the independent differential reference. The current private path retains checked fallback for changed layouts and diagnostic details. The extended differential tests cover those paths and empty updates. No numeric divergence has been found; a public-accumulator redesign can be assessed separately.

crates/core/src/engine/timing.rs
jthomson04: Both converters now use one private `known_tag(&str) -> Option<Self>` table. This avoids the extra unknown-tag allocation in the proposed delegation: `from_provider` retains its owned string, while `from_native` allocates only for an unknown borrowed tag. The test checks all known tags, unknown/case-sensitive/empty tags, and retention of the owned unknown string's allocation. `transferred` remains `Other`; no source variants or public signatures changed.

Verified at `7fddbd92`; the PR description lists the exact-head validation results.

crates/core/src/engine/timing.rs
jthomson04: Added a narrower shortcut in the separate zero-work commit: a fully empty incoming phase is skipped only when the accumulated phase contains operations. The guard checks operations, latency, energy, covered latency, and source. Empty and phase-only accumulated states retain the checked path. Differential coverage includes ordinary and diagnostic operations, source-only updates, phase-only totals, signed zero, and failure-state parity. No new end-to-end performance claim is made.

Verified at `7fddbd92`; the PR description lists the exact-head validation results.

crates/core/src/python.rs
jthomson04: Zero-work predictions now clone one immutable empty `Arc` stored in a `OnceLock`, with allocation only on first use. The existing zero-work test now checks Arc identity, unchanged accumulated evidence, and unchanged provider-call counts before and after nonzero work. This is in the separate zero-work commit; no benchmark claim is added.

Verified at `7fddbd92`; the PR description lists the exact-head validation results.

crates/core/src/python.rs
jthomson04: Added the same non-empty guard and exact error text to `NativeEntries`. The builder fixture now repeats an empty-row failure after seeding valid evidence and checks that the failed key stays absent, the prior cached Arc is retained, and accumulated evidence is unchanged. Direct builder coverage still includes an empty input, which is valid at that lower layer.

Verified at `7fddbd92`; the PR description lists the exact-head validation results.

crates/core/src/python.rs
jthomson04: Replaced the literal stride with `DEFAULT_STATIC_STRIDE` in both the production native call and the test-only Python bridge. The current value and prediction behavior are unchanged.

Verified at `7fddbd92`; the PR description lists the exact-head validation results.

crates/core/src/python.rs
jthomson04: Deferred intentionally. Reusing the existing mapper preserves both typed error formatting and its environment-dependent import fallback, as required by this optimization's compatibility contract. Python interaction remains confined to the failure path for operator evidence. A new shared Rust classifier would require a separate error-contract change and additional mirror changes, with no benefit to successful prediction calls.

crates/core/src/python.rs
jthomson04: Keeping the small branch-local selection and guard. Native results contain borrowed source tags; the test bridge contains owned strings, so returning both from one match would require additional type or conversion plumbing. The missing `NativeEntries` guard is now fixed directly and tested. This keeps the production path small and the bridge independent without adding a dispatch abstraction.

crates/core/src/python.rs
jthomson04: Keeping the checked conversion/builder as an independent reference. Sharing row validation or construction could make the same mistake affect both sides of the differential test. If only one side changes, the value/error comparison should fail and force an explicit compatibility decision. The small duplicated test-only conversion is deliberate; row-error precedence and failed cache publication remain covered.

crates/core/src/python.rs
jthomson04: Added focused native-path coverage rather than replacing every fixture. The helper is now named `python_timing_model` to make its provider explicit. The existing native-versus-bridge test exercises public prefill/decode methods, repeated cached queries, changing contexts/prefixes, cached Arc identity, and exact accumulated evidence. The tiny empty-engine fixture now exercises both the ordinary native provider and diagnostics. Python probes remain for provider-call counting, bridge parity, and FPM behavior.

Verified at `7fddbd92`; the PR description lists the exact-head validation results.

crates/core/src/python.rs
```diff
@@ -2120,18 +2213,208 @@ mod tests {
...
             use_fpm_decode_totals,
             fpm_decode_kv_ceiling: None,
-            evidence: Mutex::new(TimingEvidenceSummary::default()),
+            evidence: Mutex::new(TimingEvidenceAccumulator::default()),
             phase_cache: quick_cache::sync::Cache::new(128),
         }
     }
 
+    #[test]
+    fn native_phase_evidence_matches_python_bridge() {
```
jthomson04: **[CONFIRMED] This test is non-hermetic: it is the only test in the 1,517-test lib suite that needs the `aisimulate_core` Python package importable *and* the full gb300 / sglang-0.5.17 measured dataset on disk, so a plain checkout now fails `cargo test`.**

Reproduced on head `7fddbd92`:

```
cargo test --features embed-python --lib
test result: FAILED. 1516 passed; 1 failed
panicked at crates/core/src/python.rs:2247:
  called `Result::unwrap()` on an `Err` value: AIC timing provider could not construct the requested estimator
  Caused by: unsupported model for Rust core estimator: compile_engine:
    ModuleNotFoundError: No module named 'aisimulate_core'
```

CI passes only incidentally: `.github/workflows/ci.yml:358` pip-installs `./python/aisimulate[dev]` before the embedded-Rust job. This is a *new* hard dependency — on `main` no test reaches a successful `AicTimingModel::build`; the only build test (`main:2809`) asserts a failure.

Line 2242 also hard-codes `env!("CARGO_MANIFEST_DIR")/../../python/aisimulate/src/aisimulate_core/systems`, binding a Rust unit test to the Python package layout.

Suggest `#[ignore]`, or an env-var / data-presence gate that skips cleanly, so the crate's test suite stays runnable standalone.

crates/core/src/engine/timing.rs
```diff
@@ -225,6 +234,241 @@ pub struct TimingPhaseEvidence {
...
+impl ValidatedTimingPhase {
+    /// Native rows have unique names in first-encounter order. Each operation
+    /// is unchanged since `TimingOperationEvidence::new`, with no diagnostics.
+    /// Diagnostics must use `from_operations` after attaching details. Keep this private
+    /// contract separate from the checked constructor for public field values.
+    #[cfg(feature = "python")]
+    pub(crate) fn from_name_folded_operations(
+        operations: Vec<TimingOperationEvidence>,
+    ) -> Result<Self> {
+        debug_assert!(operations.iter().enumerate().all(|(index, operation)| {
```
jthomson04: **[PLAUSIBLE] Re-raising the `debug_assert!` name-uniqueness precondition — not to relitigate the deferral, but because the accumulator changes in this PR make the release-build consequence worse than when it was first deferred.**

_Prior thread on this line was answered with "Deferred for this PR. `run_static_per_op` already guarantees unique names in first-encounter order through `PerOpFold`." That guarantee is real today._

What changed: if a duplicate name ever does reach here in release, the two accumulate paths now disagree. `TimingEvidenceAccumulator::record`'s fast path pairs `phase.operations` and `incoming.operations` **positionally** (timing.rs:366), while the `try_accumulate` fallback folds **by name** via `merge_operation`'s `find`. So `[A,A] + [A,A]` yields 2 rows one way and 1 row (double-counted into index 0) the other. On `main` both paths folded.

The upstream guarantee lives in `PerOpFold::add`'s `find(|e| e.0 == name)` (runtime.rs:213) — one refactor away, with no compile-time tie to this function. An `ensure!` over ~30 names is free next to the model evaluation it guards.

crates/core/src/engine/timing.rs
```diff
@@ -225,6 +234,241 @@ pub struct TimingPhaseEvidence {
...
+        );
+        ensure!(
+            phase.covered_latency_ms.is_finite()
+                && phase.covered_latency_ms >= 0.0
+                && phase.covered_latency_ms <= phase.latency_ms,
+            "timing phase evidence returned invalid covered latency {}ms for {}ms total",
+            phase.covered_latency_ms,
+            phase.latency_ms
+        );
+        if phase.energy_wms.is_none() {
```
jthomson04: **[PLAUSIBLE] The inlined validation drops two steps that `TimingPhaseEvidence::canonicalized()` performs: the `energy_wms.filter(|energy| *energy > 0.0)` normalization and the `operation_totals_match()` invariant check.**

_Related to the answered thread at line 250; the debug-only round-trip assertions added there cover the per-operation inputs, not the assembled phase._

Concretely, the two constructors diverge on a phase energy of `Some(0.0)` / `Some(-0.0)`: `canonicalized()` maps it to `None` and then forces `covered_latency_ms = 0.0`, while this builder publishes `Some(0.0)` and keeps a nonzero `covered_latency_ms`.

Unreachable today only because `TimingOperationEvidence::new` has already filtered every op energy to `> 0.0`, so the `reduce` can't produce zero — which also means the differential tests in this PR can't reach it, and nothing enforces the precondition in release. Any future caller assembling operations through the public fields, or a change to `new()`'s filtering, silently produces phases the checked path would have normalized or rejected.

Calling `canonicalized()` on the assembled phase and keeping only the cheap per-op skip would close the gap without giving back the win.

crates/core/src/python.rs
```diff
@@ -483,11 +503,76 @@ impl AicTimingModel {
...
+        let phase = match &self.phase_provider {
+            #[cfg(test)]
+            AicPhaseProvider::NativeEntries(entries) => {
+                ensure!(
+                    !entries.is_empty(),
+                    "AIC {mode} returned empty operation evidence for nonzero work"
+                );
+                phase_evidence_from_native_entries(entries.clone())?
+            }
+            AicPhaseProvider::Native(engine) => {
```
jthomson04: **[PLAUSIBLE] This arm copies the body of the `AicEngine::run_static_per_op` pymethod into `python.rs`, which is why `parse_mode` and `aic_to_py` had to be widened from private to `pub(crate)` in `perfmodel/py.rs` (the +2/-2 in this PR).**

There are now two independent copies of the same call sequence — `perfmodel/py.rs:490-516` and `python.rs:519-547` — covering RuntimeConfig construction, `parse_mode`, `reset_provenance`, stride, and error mapping. Any change to the pymethod (extra validation, a different stride default, a provenance-scope change) silently leaves offline replay on the old behavior: no compile error, and no test links them except the non-hermetic one flagged at line 2230.

`python/aisimulate/.claude/rules/perfmodel-api.md` — which AGENTS.md makes a required read for Rust callers of the perf model — says: *"Existing low-level engine builders, compilation helpers, and compatibility adapters are implementation or migration machinery; they must not become a parallel public path."* This reaches around `ForwardPassPerfModel` straight into `perfmodel::engine::Engine`.

The structural fix is one helper on `Engine` / `ForwardPassPerfModel` returning `AicError`, called by both the pymethod and this adapter — which would also dissolve the `aic_to_py` round-trip flagged at line 538.

crates/core/src/python.rs
```diff
@@ -483,11 +503,76 @@ impl AicTimingModel {
...
+                let (context, generation) = crate::py::parse_mode(mode)
+                    .and_then(|mode| {
+                        engine.reset_provenance();
+                        engine
+                            .run_static_per_op(
+                                &runtime,
+                                mode,
+                                crate::perfmodel::engine::DEFAULT_STATIC_STRIDE,
+                            )
+                            .map_err(crate::py::aic_to_py)
```
jthomson04: **[PLAUSIBLE] Re-raising the `aic_to_py` round-trip with the specific cost, since the prior thread was deferred on the grounds of preserving typed error formatting.**

_Prior answer: "Reusing the existing mapper preserves both typed error formatting and its environment-dependent import fallback."_

The concern isn't the formatting, it's where the GIL gets taken. `aic_to_py` calls `Python::with_gil` and `py.import("aisimulate_core.sdk.errors")` (perfmodel/py.rs:107-133), and `PyErr`'s `Display` impl acquires the GIL *again* to normalize the exception — so line 540's `{error}` interpolation is a second acquisition. Previously all of this ran inside the pymethod with the GIL already held.

Now a replay worker thread that this PR deliberately made GIL-free re-acquires the GIL on every provider error (missing perf data, missing system flops, invalid mode), contending with the main Python thread, and the native evidence path stays coupled to an initialized interpreter purely to build a message string.

An `AicError -> anyhow` mapper would keep the typed formatting and leave the `PyErr` conversion at the pyo3 boundary where it belongs.

crates/core/src/engine/timing.rs
```diff
@@ -225,6 +234,241 @@ pub struct TimingPhaseEvidence {
...
+    pub(crate) fn record(&mut self, incoming: &ValidatedTimingPhase, prefill: bool) -> Result<()> {
+        let phase = if prefill {
+            &mut self.summary.prefill
+        } else {
+            &mut self.summary.decode
+        };
+        let incoming = incoming.as_phase();
+        // Only operation-backed totals are already reconciled. Keep empty and
+        // phase-only accumulated states on the checked path, including -0.0.
+        if !phase.operations.is_empty()
```
jthomson04: **[PLAUSIBLE] Follow-up on the narrowed shortcut added in `7fddbd92`: `!phase.operations.is_empty()` is not equivalent to "already reconciled", which is what the comment above it claims.**

`try_accumulate` skips `reconcile_operation_totals()` whenever *either* operand is phase-only (`has_phase_only_totals`). So a phase-only update carrying a tiny nonzero `latency_ms` — below `operation_totals_match`'s 1e-9 relative tolerance — commits a state where `phase.latency_ms != sum(op.latency_ms)` and still passes `canonicalized()`, with `operations` non-empty.

Before this commit, the next empty `record` ran the checked path, which reconciles and snaps the total back to the operation sum. With the skip, that drift is retained and compounds across every subsequent step of a long replay.

Unreachable today only because `predict_phase_evidence` returns either the all-zero `EMPTY` or an operation-backed phase (the comment at timing.rs:1478 says as much) — i.e. the shortcut's correctness rests on a provider-adapter property that nothing in this function enforces. Gating on the accumulated phase actually being operation-reconciled would make the comment true.

crates/core/src/python.rs
```diff
@@ -497,43 +582,48 @@ impl AicTimingModel {
...
-        } else {
-            evidence.decode.try_accumulate(phase)?;
-        }
-        Ok(())
+        evidence.record(phase, prefill)
     }
 }
 
-fn phase_evidence_from_python(
+fn phase_evidence_from_native_entries(
```
jthomson04: **[PLAUSIBLE] Re-raising the near-duplicate helpers, because the "independent reference" argument is what makes the duplication load-bearing rather than merely redundant.**

_Prior answer: "Keeping the checked conversion/builder as an independent reference. Sharing row validation or construction could make the same mistake twice."_

Agreed on the goal, but `phase_evidence_from_native_entries` and `phase_evidence_from_entries` (line 624) are byte-for-byte identical apart from `from_native` vs `from_provider` and which `ValidatedTimingPhase` constructor they call — including the `ensure!` text at 608/630. The independence that matters for the differential test is the *builder* (`from_name_folded_operations` vs `from_operations`), not the 12 lines of identical raw-energy validation and row construction wrapped around it.

Those 12 shared lines have to be kept in lockstep by hand. A fix applied to one and not the other (a new field, a changed error message) turns the reference into a silent false negative instead of a compile error — the exact failure mode the test exists to rule out. Parameterizing the source conversion and the builder keeps the independence where it's doing work and removes it where it's a liability.

crates/core/src/python.rs
```diff
@@ -483,11 +503,76 @@ impl AicTimingModel {
...
                 })
                 .collect::<Result<Vec<_>>>()?;
-            let phase = TimingPhaseEvidence::try_from_operations(operations)?;
-            self.phase_cache.insert(key, phase.clone());
+            let phase = Arc::new(ValidatedTimingPhase::from_operations(operations)?);
+            self.phase_cache.insert(key, Arc::clone(&phase));
             return Ok(phase);
         }
-        let (context, generation) = Python::with_gil(|py| {
+        let phase = match &self.phase_provider {
```
jthomson04: **[PLAUSIBLE] The phase selection and the non-empty guard are now written three times (lines 513, 542, 554), two of them in `#[cfg(test)]` arms.**

_The prior thread at line 541 was answered with "Keeping the small branch-local selection and guard. Native results contain borrowed source tags; the test bridge contains owned strings."_

That's true of `phase_evidence_from_*`, but not of the `let entries = if prefill { ... }` + `ensure!` pair, which is identical in all three arms and independent of the element type. The error text is triplicated; changing the wording or the prefill selection takes three edits, and the two `cfg(test)` arms can drift from the production arm without failing to compile — which is precisely the divergence the differential tests exist to detect.

Having each arm return only `(context, generation)` and hoisting the selection plus the `ensure!` below the match keeps the type distinction where it's real and removes it where it isn't.

crates/core/src/python.rs
```diff
@@ -388,14 +390,28 @@ impl AicTimingConfig {
 
 type PhaseEvidenceKey = (u32, u32, u32, u32, bool);
 
+#[cfg(test)]
+type PythonPerOpEntries = Vec<(String, f64, f64, String)>;
+
+enum AicPhaseProvider {
```
jthomson04: **[PLAUSIBLE] Two of this enum's three variants are `#[cfg(test)]`, so the shipping build carries a one-variant enum plus a `match` that exists purely as a test seam.**

_Related to the answered coverage thread at line 2221 (the helper renamed to `python_timing_model`)._

`AicTimingModel` now holds three handles to the same engine: `engine: Py<PyAny>`, `phase_provider: Native(Arc<PerfEngine>)`, and `diagnostic_model`'s `ForwardPassPerfModel`. And `phase_provider` is dead whenever `use_fpm_decode_totals` is true or `diagnostic_model` is `Some`, because the diagnostics branch at line 487 returns first.

A plain `phase_provider: Arc<PerfEngine>` field plus a separate `#[cfg(test)]` override hook expresses the same testability without putting test-only variants in a production type — and would remove the `unreachable!()` at line 2251 that exists only to discharge them.

crates/core/src/engine/timing.rs
```diff
@@ -225,6 +234,241 @@ pub struct TimingPhaseEvidence {
...
+    #[cfg(feature = "python")]
+    pub(crate) fn from_name_folded_operations(
+        operations: Vec<TimingOperationEvidence>,
+    ) -> Result<Self> {
+        debug_assert!(operations.iter().enumerate().all(|(index, operation)| {
+            !operations[..index]
+                .iter()
+                .any(|previous| previous.name == operation.name)
+        }));
+        debug_assert!(operations.iter().all(|operation| {
```
jthomson04: **[PLAUSIBLE] This assertion — added in `ae4b35f6` in response to the canonicalization thread — clones every operation and runs full `canonicalized()` validation on each, on every cache miss, on top of the O(n^2) name scan directly above it.**

`operation.clone().canonicalized()` is a deep clone including the `name` and `Other(String)` source allocations, plus an equality comparison, per operation per provider call. For this PR's own AgentX workload that's roughly 2,237 provider calls x ~24 operations.

That reintroduces exactly the validation cost the PR set out to remove, for every `cargo test` and every debug replay run — the configurations where the 34% CPU win is hardest to observe and easiest to mistake for noise.

The invariants are checkable in place without cloning: `details.is_none()`, `energy_wms.is_none_or(|e| e > 0.0)`, and the finite / non-negative field checks. Same coverage, no allocation.

Harrilee (approved):
Reviewed `4a95f4ea3f8208a96289698769890232dd55a328` against `e8828036e3d32d8361e2abb819e0ee60e569f518`. No blocking findings after tracing provider construction, cache ownership, checked/fast accumulation, energy/coverage/source handling, error precedence and atomicity, empty work, and diagnostics fallback.

Independent correctness evidence on this head:

- Fresh optimized native builds with identical Cargo/uv locks and the same 1,537 performance-data files.
- Embedded-Python Rust suite: 1,585 passed, one existing ignored test; native-versus-Python bridge parity explicitly executed. Power tests: 164 passed.
- Additional standalone differential harness: 100,000 randomized inputs; 94,756 successful updates matched the checked accumulator bitwise, 2,087 constructor rejections, and 3,157 matching atomic overflow rejections. All 28 included timing/support/differential tests passed.
- Public AgentX: all 4,198 per-request results and all 105 report keys, including timing evidence, matched exactly except the three host-speed metrics (`wall_time_ms`, `processed_tokens_per_s`, `processed_output_tokens_per_s`). The deterministic synthetic fixture likewise matched all 256 requests and 104 report keys. Every counted performance pair also had zero prediction/evidence differences.
- Diagnostics enabled (`--detail time,source`) also matched exactly on a separate 162-request trace. These benchmark profiles have zero positive power coverage; positive-energy arithmetic is covered separately by the power tests and randomized differential checks.
- Independently downloaded the exact-head CI old/new prediction artifacts: all 30 CSVs were byte-identical at their serialized precision (170,800 rows, including 83,379 successful predictions; remaining misses/invalid/error outcomes were also unchanged). No goldens changed.

Independent end-to-end measurement: optimized Linux x86_64 builds on an i9-14900K, CPU 4 pinned, warm filesystem caches, default diagnostics/capture off, fixed five alternating revision pairs per workload. CPU includes the CLI supervisor and waited descendants; subprocess accounting was separately verified. All 20 counted runs are retained.

| Workload | Base CPU median | PR CPU median | CPU reduction | Base wall median | PR wall median | Wall reduction |
|---|---:|---:|---:|---:|---:|---:|
| Public AgentX, 32 plays | 74.040 s | 41.743 s | **43.62% (1.774x)** | 74.077 s | 41.795 s | 43.58% |
| Synthetic long decode | 1.807 s | 1.596 s | **11.70% (1.133x)** | 1.821 s | 1.669 s | 8.33% |

AgentX uses public Weka dataset revision `8fecd2fc56694469f758f0afbbb6335ad3043740`, first 32 rows, 4,198 requests / 4,072,673 output tokens, Qwen3.8-2.4T-A95B-FP8 / GB300 / SGLang 0.5.17 / TP16 / MoE TP4 EP4, one client-play lane, explicit trace/KV block size 64. Input SHA-256: `ba7a19e045828db8d8299e71305d046b2ed3259e515907711a9a15e05d6c140c`. Synthetic uses 256 independent one-request plays, 32 lanes, 1,024 input / 4,096 output tokens, stable identities, and the same model/hardware topology.

The AgentX CPU reduction's nominal paired-log Student-t 95% interval is 38.49–48.39%; synthetic is 10.59–12.40%. The host is shared and brief unrelated compiler activity was observed; all samples and contemporaneous load records were retained. These intervals are small-sample estimates, not isolation guarantees. This is an independent, similarly sized public workload, not the author's original 4,265-request trace. The claimed speedup direction and magnitude are credible.

Setup receipts are retained: an initial resource precheck overestimated trace memory when source block size was omitted; explicitly matching the corpus's block size 64 admitted both revisions without changing token lengths. The initial non-agentic synthetic fixture generated random UUIDs and showed only tiny aggregation-order differences (about `6e-11` maximum) after requests were aligned; the stable-identity fixture above eliminates those differences without changing production code.

I also reviewed the five open, explicitly deferred author self-review threads. Current callers uphold the private builder's unique-name contract, and the independent checked reference and existing error mapper preserve the intended compatibility boundary. Those maintenance discussions do not establish a reachable blocking defect in this head. Their resolution and branch currency remain repository merge requirements.

Full CI run [35649460908](https://github.com/ai-dynamo/aisimulate/actions/runs/35649460908) passed on this exact head. Approval is scoped to this commit.

dreamtalen (approved):

